PR #5, opened 2019-07-24
Add tool for triggering screenshot

Jamedjo:
## What

Adds a tool that triggers a request for a screenshot and prints out the response.

## Why

Makes it easier to get started making changes to the existing code. In particular it helped me both verify that it was working, and to step through and understand the code using a debugger.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,19 @@ Make sure `XDG_CURRENT_DESKTOP=sway` is set.
 xdg-desktop-portal-wlr
 ```
 
+## Tooling
+
+Useful tools include `dbus-monitor` to watch requests being made,
+and `dbus-send` and the similar `busctl call` for manual dbus calls.
+
+Additionally you can test the integration with the provided helper scripts in the tools directory.
+
+To trigger a screenshot and open it you can run the following:
+
+```sh
+python3 ./tools/triger-screen-shot.py | xargs xdg-open
+```
+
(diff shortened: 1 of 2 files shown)


emersion:
Oh, thanks for the pull request! However there is already a full xdg-desktop-portal demo available here: https://github.com/matthiasclasen/portal-test

It supports screenshots and screen capture (the latter would be a little complicated to implement in Python I think).

Jamedjo:
> There is already a full xdg-desktop-portal demo available here: https://github.com/matthiasclasen/portal-test

Ah, perfect. I'll take a look at that and we can mention that in the README instead.

> It supports screenshots and screen capture (the latter would be a little complicated to implement in Python I think).

I actually based this on a python screen-capture snippet, https://gitlab.gnome.org/snippets/19, so definitely not too complicated.

The exercise was still useful as I now have a better understanding of DBus and the protocols used here.

emersion:
>Ah, perfect. I'll take a look at that and we can mention that in the README instead.

:+1:

>I actually based this on a python screen-capture snippet, https://gitlab.gnome.org/snippets/19, so definitely not too complicated.

Oh I see, and they use `pipewiresrc` to do the tricky part.

>The exercise was still useful as I now have a better understanding of DBus and the protocols used here.

Yeah definitely :)

Thanks for your work!

PR #6, opened 2019-07-24
Use sd_bus_message_append shorthand for screenshot

Jamedjo:
## What

Combine multiple `sd_bus_message_append` calls.

## Why

Reduces repetition and error handling overhead, makes it easier to see at a glance what is being done.

### DIFF
--- a/include/xdpw.h
+++ b/include/xdpw.h
@@ -8,6 +8,12 @@ struct xdpw_request {
 	sd_bus_slot *slot;
 };
 
+enum {
+	PORTAL_RESPONSE_SUCCESS = 0,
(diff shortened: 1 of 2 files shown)


include/xdpw.h
```diff
@@ -8,6 +8,12 @@ struct xdpw_request {
 	sd_bus_slot *slot;
 };
 
+enum {
+  PORTAL_RESPONSE_SUCCESS   = 0,
```
emersion: Style: tabs, don't align

Jamedjo: > Style: tabs, don't align

Updated :+1:

PR #7, opened 2019-07-24
Add tooling to README mentioning portal-test and python snippets

Jamedjo:
I haven't managed to get `portal-test` to work yet, so have also included links to the python snippets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,15 @@ Make sure `XDG_CURRENT_DESKTOP=sway` is set.
 xdg-desktop-portal-wlr
 ```
 
+## Tooling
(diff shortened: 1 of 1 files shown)


Jamedjo:
> FWIW, works for me. It was a long time ago though, so I don't remember the details. :P

I think I need to get flatpak set up first, the error I was getting was `Can't load uri https://sdk.gnome.org/gnome-nightly.flatpakrepo: TLS support is not available` attempting `flatpak remote-add ...`. Lots of search results / issues for the error but no obvious solution so I've fallen back to the python script for now.

PR #15, opened 2020-03-16
Add pipewire 0.3 support. Workaround pipewire bugs.

### LINKED ISSUE #10: screencast: pipewire 0.3 support
Screencast code will need rework for API changes in Pipewire 0.3.

[PR](https://gitlab.gnome.org/GNOME/mutter/merge_requests/1006#10194861d1fd363d5705e4a5788c5ea79a253490) for Mutter supporting Pipewire 0.3

danshick:
Fixes #10.

### DIFF
--- a/.build.yml
+++ b/.build.yml
@@ -5,7 +5,7 @@ packages:
   - meson
   - wayland
   - wayland-protocols
-  - libpipewire02
+  - pipewire
   - libdrm
 sources:
   - https://github.com/emersion/xdg-desktop-portal-wlr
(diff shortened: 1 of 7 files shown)


danshick:
For anyone wondering, if you're an arch user, this will work with the version of pipewire that is officially packaged. If you are using the AUR package "fedora-firefox-wayland-bin", you will still need to install the package called "libpipewire02" to get the proper libs. I've left a comment there to update that dependency.

This doesn't fix sessions, so it still needs to be restarted for each new connection (including the preview window in the firefox permissions dialog) and it still requires a flag "-p BGRx" to trick WebRTC into accepting our pixel format.

PR #17, opened 2020-03-22
add missing libdrm to meson.build

obayemi:
libdrm was missing in `meson.build` file, so configuration did not detect when it was absent from the system.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -57,7 +57,7 @@ executable(
 		logind,
 		pipewire,
 		rt,
-        drm
+		drm
 	],
 	include_directories: [inc],
 	install: true,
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -26,6 +26,7 @@ rt = cc.find_library('rt')
 pipewire = dependency('libpipewire-0.3', version: '>= 0.2.9')
 wayland_client = dependency('wayland-client')
 wayland_protos = dependency('wayland-protocols', version: '>=1.14')
+drm = dependency('libdrm')
```
emersion: Can we use a `partial_dependency`, since we only rely on the header constants?

obayemi: ok, looking at it (sorry, I don't know meson very well / at all)

obayemi: is `drm = dependency('libdrm').partial_dependency(includes: true)` ok ?
should I add an other `partial_drm` dependency ?

emersion: This should be fine, yes.

Issue #18, opened 2020-03-23
screencast: properly report cursor mode


danshick:
Should fix compatibility with obs-xdg-portal plugin per [this](https://gitlab.gnome.org/feaneron/obs-xdg-portal/-/issues/2).

Eventually, we should support all of these cursor modes properly, but for now it might be a good idea to allow this to be "fudged" with a CLI flag like we do with pixelformat, in case of compatibility issues.

emersion:
The wlroots protocol supports Hidden and Embedded. We can set AvailableCursorModes to `0x1 | 0x2`.

danshick:
We could probably even hack together mode 0x4 if we can retrieve the cursor position from wlroots. That's literally just "send me a location and a bitmap of a cursor" if I understand correctly.

emersion:
We can't get neither the cursor position nor the cursor image.

danshick:
Huh, okay. Guess i had assumed those would be easy.

Issue #9, opened 2020-01-15
screencast: dmabuf support


danshick:
Support passing a dmabuf fd to pipewire instead of using the screencopy protocol. There are three things pending before this can (or should) move forward.

1. ~~Waiting on Pipewire 0.3 to be released:
    https://gitlab.freedesktop.org/pipewire/pipewire~~

2. ~~Waiting on xdg-desktop-portal PR to support Pipewire 0.3:
    https://github.com/flatpak/xdg-desktop-portal/pull/436~~

3. ~~Waiting on downstream ecosystem to support drm_fourcc modifiers:
    [Discussion of downstream dependencies](https://github.com/emersion/xdg-desktop-portal-wlr/issues/1#issuecomment-571684093) that need to support modifiers.~~ resolved by newer screencopy protocol

[Here](https://github.com/danshick/wlr_zerocopy_pipewire) is a very early, broken demo

braderhart:
Pipewire 0.3 is out now. Hope that will help clear up the issues.

danshick:
That certainly means we can complete dmabuf style screensharing, but it'll be mostly useless unless and until downstream apps support various modifiers. Still worth building out, although it'll likely get hidden behind an argument.

danshick:
Maybe we could use glBlitFramebuffer to copy frames on GPU before sharing. This could avoid the concern about format modifiers. I need to research and experiment with this approach further.

emersion:
https://github.com/swaywm/wlroots/pull/2133 allows blitting DMA-BUFs on the compositor-side.

emersion:
>Do you think this should be implemented under a CLI flag?

Ideally it should be enabled by default. We should offer support for both DMA-BUF and good old shm, and clients should pick one of those.

danshick:
I totally forgot that Mutter already does this (dmabuf producer for xdg-desktop-portal-gtk, consumed by WebRTC). Basically, this should work fine in all cases.

We can't let clients pick one because there is no provision in the portal API for that negotiation to happen, but luckily, I don't think we have to.

The only fallback that would be necessary is if the screencopy protocol version is insufficient on the compositor side.

emersion:
FWIW, https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1442

I think GNOME also blits to LINEAR because all clients ignore the modifier hint and end up with garbage.

[...]

Issue #8, opened 2019-10-10
Possibility of crowd-funding to move things along


Tormen:
Dear @emersion ,

first I wanted to thank you for all your efforts in driving so many sway related bits and pieces!!!
I love sway and would not want to miss it.

Only one big pain point remains: Lack of screen sharing support (e.g. via zoom linux app, which works fine under GNOME).

Hence my interest in this project. 

Do you see any chance we or I could organise crowd-funding (I am willing to pay 200 EUR for this) in order to speed things up / drive that project forward into a usable state?

Thanks a lot in advance for your feedback!

Tormen

ddevault:
I think the issue is more related to lack of interest than lack of money. If you are interested in this project, perhaps you could step up and write some patches?

siwyd:
Lots of interest, zero C skills ;)

ddevault:
C is not especially difficult to learn.

siwyd:
Time is hard to come by right now unfortunately *sigh*.

emersion:
Thanks for your interest in the project!

Right now the main issue is that I have a finite amount of time and focus on other projects. I still want to work on this at some point, but I can't guarantee when.

ThenAndNow:
I recommend using Gitcoin to help fund open source work, as a result of this post I actually set up a $50 bounty on the screencast issue https://github.com/emersion/xdg-desktop-portal-wlr/issues/1

ddevault:
>which part is proprietary

Where is the source code for gitcoin.io or any of the tools on it?

>Why is this anything but a good thing? Developer gets paid, I'm happy because an issue I care about gets fixed that's a definition of win-win.

Because it removes the maintainers from the loop. Handling money in an open source project is a delicate issue that requires the maintainer's full support. Just arbitrarily assuming your crap is welcome in every open source project is an awful idea, it's been done before and it was terrible then too.

ddevault:
And furthermore: how much money does Gitcoin scrape off of the top? It's a business, they're not altruistic, they're a leech on open source software.

ThenAndNow:
> Where is the source code for gitcoin.io or any of the tools on it?
The whole thing happens on the blockchain so it's transparent, unless you're talking about the source code for the Gitcoin website but I don't understand why you'd want that.

[...]

PR #22, opened 2020-04-08
Screencast session support

### LINKED ISSUE #13: screencast: properly handle sessions
At the moment, if you stop one screencast and attempt to start another without restarting xdg-desktop-portal-wlr, it will segfault.

Ideally, we should separate out a data structure that represents each output and its pipewire stream, as well as another data structure representing each d-bus screencast session. This would allow multiple sessions both in parallel or serially and would support parallel sessions reusing a pipewire node. If no sessions were using an output, we could destroy the output and recreate it as needed.

danshick:
Fixes #13.

This is designed to take pairs of output id and cursor mode and create a single pipewire stream for each unique combo. Until output selection can be controlled at screencast initialization, this doesn't serve much of a purpose, but it is designed with that capability in mind.

~~At the moment, this gets all the way to a pipewire stream PAUSED state and continues processing wlroots frames like it is supposed to, but the stream is not able to activate. I need to investigate changes made to the pipewire portions of the code to see if I created a regression.~~

Edit: The pipewire issues are fixed.

@emersion, in order to give all of these event driven functions an understanding of which session they belong to, I've started to pass around "cast" session objects instead of the xdpw_state or xdpw_screencast_context, which are both more global in nature, but some functions still need access to data in the context or global state object (pw_loop comes to mind). To accomplish this, each session object has a pointer back to the screencast_context and the context holds a pointer back up to the state. Wanted to point this out in case you see any issue with this approach.

### DIFF
--- a/include/xdpw.h
+++ b/include/xdpw.h
@@ -7,15 +7,17 @@
 #elif HAVE_ELOGIND
 #include <elogind/sd-bus.h>
 #endif
-#include "logger.h"
-#include "screencast.h"
+
+#include "screencast_common.h"
 
 struct xdpw_state {
+	struct wl_list xdpw_sessions;
 	sd_bus *bus;
 	struct wl_display *wl_display;
 	struct pw_loop *pw_loop;
-
-	struct screencast_context screencast;
+	struct xdpw_screencast_context screencast;
+	enum source_types screencast_source_types;
+	enum cursor_modes screencast_cursor_modes;
 };
 
 struct xdpw_request {
@@ -23,7 +25,10 @@ struct xdpw_request {
 };
 
 struct xdpw_session {
+	struct wl_list link;
 	sd_bus_slot *slot;
+	const char *session_handle;
+	struct xdpw_screencast_instance *screencast_instance;
 };
 
 enum {
@@ -32,14 +37,14 @@ enum {
 	PORTAL_RESPONSE_ENDED = 2
 };
 
-int init_screenshot(struct xdpw_state *state);
-int init_screencast(struct xdpw_state *state, const char *output_name,
+int xdpw_screenshot_init(struct xdpw_state *state);
+int xdpw_screencast_init(struct xdpw_state *state, const char *output_name,
 	const char *forced_pixelformat);
 
-struct xdpw_request *request_create(sd_bus *bus, const char *object_path);
-void request_destroy(struct xdpw_request *req);
+struct xdpw_request *xdpw_request_create(sd_bus *bus, const char *object_path);
+void xdpw_request_destroy(struct xdpw_request *req);
 
-struct xdpw_session *session_create(sd_bus *bus, const char *object_path);
-void session_destroy(struct xdpw_session *req);
+struct xdpw_session *xdpw_session_create(struct xdpw_state *state, sd_bus *bus, const char *object_path);
+void xdpw_session_destroy(struct xdpw_session *req);
 
 #endif
--- a/include/xdpw.h
+++ b/include/xdpw.h
@@ -16,8 +16,9 @@ struct xdpw_state {
 	struct wl_display *wl_display;
 	struct pw_loop *pw_loop;
 	struct xdpw_screencast_context screencast;
-	enum source_types screencast_source_types;
-	enum cursor_modes screencast_cursor_modes;
+	uint32_t screencast_source_types; // bitfield of enum source_types
+	uint32_t screencast_cursor_modes; // bitfield of enum cursor_modes
+	uint32_t screencast_version;
 };
 
 struct xdpw_request {
@@ -27,7 +28,7 @@ struct xdpw_request {
 struct xdpw_session {
 	struct wl_list link;
 	sd_bus_slot *slot;
-	const char *session_handle;
+	char *session_handle;
 	struct xdpw_screencast_instance *screencast_instance;
 };
 
@@ -44,7 +45,7 @@ int xdpw_screencast_init(struct xdpw_state *state, const char *output_name,
 struct xdpw_request *xdpw_request_create(sd_bus *bus, const char *object_path);
 void xdpw_request_destroy(struct xdpw_request *req);
 
-struct xdpw_session *xdpw_session_create(struct xdpw_state *state, sd_bus *bus, const char *object_path);
+struct xdpw_session *xdpw_session_create(struct xdpw_state *state, sd_bus *bus, char *object_path);
 void xdpw_session_destroy(struct xdpw_session *req);
 
 #endif
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -6,6 +6,8 @@
 #include <pipewire/pipewire.h>
 #include <spa/utils/result.h>
 #include "xdpw.h"
+#include "screencast_common.h"
+#include "logger.h"
 
 enum event_loop_fd {
 	EVENT_LOOP_DBUS,
@@ -83,6 +85,7 @@ int main(int argc, char *argv[]) {
 		logprint(ERROR, "wayland: failed to connect to display");
 		goto error;
 	}
+	logprint(INFO, "wlroots: wl_display fd: %d", wl_display_get_fd(wl_display));
 
 	pw_init(NULL, NULL);
 	struct pw_loop *pw_loop = pw_loop_new(NULL);
@@ -92,13 +95,18 @@ int main(int argc, char *argv[]) {
 	}
 
 	struct xdpw_state state = {
+		.xdpw_sessions = (struct wl_list) { 0 },
 		.bus = bus,
 		.wl_display = wl_display,
 		.pw_loop = pw_loop,
+		.screencast_source_types = MONITOR,
+		.screencast_cursor_modes = HIDDEN | EMBEDDED,
 	};
 
-	init_screenshot(&state);
-	init_screencast(&state, output_name, forced_pixelformat);
+	wl_list_init(&state.xdpw_sessions);
+
+	xdpw_screenshot_init(&state);
+	xdpw_screencast_init(&state, output_name, forced_pixelformat);
 
 	ret = sd_bus_request_name(bus, service_name, 0);
 	if (ret < 0) {
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -6,7 +6,6 @@
 #include <pipewire/pipewire.h>
 #include <spa/utils/result.h>
 #include "xdpw.h"
-#include "screencast_common.h"
 #include "logger.h"
 
 enum event_loop_fd {
@@ -79,13 +78,14 @@ int main(int argc, char *argv[]) {
 		logprint(ERROR, "dbus: failed to connect to user bus: %s", strerror(-ret));
 		goto error;
 	}
+	logprint(TRACE, "dbus: connected");
 
 	struct wl_display *wl_display = wl_display_connect(NULL);
 	if (!wl_display) {
 		logprint(ERROR, "wayland: failed to connect to display");
 		goto error;
 	}
-	logprint(INFO, "wlroots: wl_display fd: %d", wl_display_get_fd(wl_display));
+	logprint(TRACE, "wlroots: wl_display connected");
 
 	pw_init(NULL, NULL);
 	struct pw_loop *pw_loop = pw_loop_new(NULL);
@@ -93,20 +93,25 @@ int main(int argc, char *argv[]) {
 		logprint(ERROR, "pipewire: failed to create loop");
 		goto error;
 	}
+	logprint(TRACE, "pipewire: pw_loop created");
 
 	struct xdpw_state state = {
-		.xdpw_sessions = (struct wl_list) { 0 },
 		.bus = bus,
 		.wl_display = wl_display,
 		.pw_loop = pw_loop,
 		.screencast_source_types = MONITOR,
 		.screencast_cursor_modes = HIDDEN | EMBEDDED,
+		.screencast_version = XDP_CAST_PROTO_VER,
 	};
 
 	wl_list_init(&state.xdpw_sessions);
 
 	xdpw_screenshot_init(&state);
-	xdpw_screencast_init(&state, output_name, forced_pixelformat);
+	ret = xdpw_screencast_init(&state, output_name, forced_pixelformat);
+	if (ret < 0) {
+		logprint(ERROR, "xdpw: failed to initialize screencast");
+		goto error;
+	}
 
 	ret = sd_bus_request_name(bus, service_name, 0);
 	if (ret < 0) {
@@ -181,5 +186,6 @@ error:
 	sd_bus_unref(bus);
 	pw_loop_leave(state.pw_loop);
 	pw_loop_destroy(state.pw_loop);
+	wl_display_disconnect(state.wl_display);
 	return EXIT_FAILURE;
 }
--- a/src/core/request.c
+++ b/src/core/request.c
@@ -11,8 +11,6 @@ static int method_close(sd_bus_message *msg, void *data,
 		sd_bus_error *ret_error) {
 	struct xdpw_request *req = data;
 	int ret = 0;
-	// struct xdpw_request *req = data;
-	// TODO
 	logprint(INFO, "dbus: request closed");
 
 	sd_bus_message *reply = NULL;
--- a/src/core/request.c
+++ b/src/core/request.c
@@ -3,13 +3,14 @@
 #include <stdlib.h>
 #include <string.h>
 #include "xdpw.h"
+#include "logger.h"
 
 static const char interface_name[] = "org.freedesktop.impl.portal.Request";
 
 static int method_close(sd_bus_message *msg, void *data,
 		sd_bus_error *ret_error) {
-	
-	int ret = 0; 
+	struct xdpw_request *req = data;
+	int ret = 0;
 	// struct xdpw_request *req = data;
 	// TODO
 	logprint(INFO, "dbus: request closed");
@@ -25,7 +26,9 @@ static int method_close(sd_bus_message *msg, void *data,
 		return ret;
 	}
 
-	sd_bus_message_unref(reply);	
+	sd_bus_message_unref(reply);
+
+	xdpw_request_destroy(req);
 
 	return 0;
 }
@@ -36,7 +39,7 @@ static const sd_bus_vtable request_vtable[] = {
 	SD_BUS_VTABLE_END
 };
 
-struct xdpw_request *request_create(sd_bus *bus, const char *object_path) {
+struct xdpw_request *xdpw_request_create(sd_bus *bus, const char *object_path) {
 	struct xdpw_request *req = calloc(1, sizeof(struct xdpw_request));
 
 	if (sd_bus_add_object_vtable(bus, &req->slot, object_path, interface_name,
@@ -50,7 +53,7 @@ struct xdpw_request *request_create(sd_bus *bus, const char *object_path) {
 	return req;
 }
 
-void request_destroy(struct xdpw_request *req) {
+void xdpw_request_destroy(struct xdpw_request *req) {
 	if (req == NULL) {
 		return;
 	}
--- a/src/core/session.c
+++ b/src/core/session.c
@@ -2,6 +2,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <assert.h>
 #include "xdpw.h"
 #include "screencast.h"
 #include "logger.h"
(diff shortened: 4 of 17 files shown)


danshick:
This will likely fix #18 as well before it's completed.

emersion:
>I've started to pass around "cast" session objects instead of the xdpw_state or xdpw_screencast_context

Yeah, this sounds perfectly fine

danshick:
Okay, this is working-ish. It's still a little finicky. If anyone wants to try it out (especially stress test it, open lots of streams, close them, in parallel or in series), that'd be great.

There's something weird going on right now with chromium still. Has to do with it requesting individual window shares from us, even though we advertise that we don't support that. I'll work on cleaning that up.

alebastr:
Thanks!
Haven't done any meaningful testing, however ASan already reported some issues:
<details>
<summary>ERROR: AddressSanitizer: heap-use-after-free on address 0x612000001c2c at pc 0x00000049b312 bp 0x7ffe5a4f7230 sp 0x7ffe5a4f69e0</summary>
<pre>
2020/04/08 14:50:08 [INFO] - wlroots: wl_display fd: 4
2020/04/08 14:50:13 [INFO] - dbus: create session method invoked
2020/04/08 14:50:13 [INFO] - dbus: request_handle: /org/freedesktop/portal/desktop/request/1_217/u1
2020/04/08 14:50:13 [INFO] - dbus: session_handle: /org/freedesktop/portal/desktop/session/1_217/u1
2020/04/08 14:50:13 [INFO] - dbus: app_id:
2020/04/08 14:50:13 [INFO] - dbus: select sources method invoked
2020/04/08 14:50:13 [INFO] - dbus: request_handle: /org/freedesktop/portal/desktop/request/1_217/u2
2020/04/08 14:50:13 [INFO] - dbus: session_handle: /org/freedesktop/portal/desktop/session/1_217/u1
2020/04/08 14:50:13 [INFO] - dbus: app_id:
2020/04/08 14:50:13 [INFO] - dbus: option types:3
2020/04/08 14:50:13 [INFO] - dbus: option multiple: 0
=================================================================
==48441==ERROR: AddressSanitizer: heap-use-after-free on address 0x612000001c2c at pc 0x00000049b312 bp 0x7ffe5a4f7230 sp 0x7ffe5a4f69e0
READ of size 49 at 0x612000001c2c thread T0
    #0 0x49b311 in __interceptor_strcmp.part.0 (/home/alebastr/sources/xdg-desktop-portal-wlr/build-asan/xdg-desktop-portal-wlr+0x49b311)
    #1 0x4fe493 in method_screencast_select_sources /home/alebastr/sources/xdg-desktop-portal-wlr/build-asan/../src/screencast/screencast.c:259:6
    #2 0x7f5903d4ffb5  ../src/libsystemd/sd-bus/bus-objects.c:409:21
    #3 0x7f5903d4ffb5  ../src/libsystemd/sd-bus/bus-objects.c:1332:21
    #4 0x7f5903da3edc  ../src/libsystemd/sd-bus/bus-objects.c:1452:21
    #5 0x7f5903da3edc  ../src/libsystemd/sd-bus/sd-bus.c:2763:13
    #6 0x7f5903da3edc  ../src/libsystemd/sd-bus/sd-bus.c:2805:13
    #7 0x7f5903da3edc  ../src/libsystemd/sd-bus/sd-bus.c:3025:21
    #8 0x4f5eb6 in main /home/alebastr/sources/xdg-desktop-portal-wlr/build-asan/../src/core/main.c:142:11
    #9 0x7f590393e041 in __libc_start_main (/lib64/libc.so.6+0x27041)
    #10 0x41d7fd in _start (/home/alebastr/sources/xdg-desktop-portal-wlr/build-asan/xdg-desktop-portal-wlr+0x41d7fd)

0x612000001c2c is located 236 bytes inside of 304-byte region [0x612000001b40,0x612000001c70)
freed by thread T0 here:
    #0 0x4bf317 in free (/home/alebastr/sources/xdg-desktop-portal-wlr/build-asan/xdg-desktop-portal-wlr+0x4bf317)
    #1 0x7f5903d36d0b  ../src/libsystemd/sd-bus/bus-message.c:131:17

previously allocated by thread T0 here:
    #0 0x4bf998 in realloc (/home/alebastr/sources/xdg-desktop-portal-wlr/build-asan/xdg-desktop-portal-wlr+0x4bf998)
    #1 0x7f5903d5b704  ../src/libsystemd/sd-bus/bus-socket.c:1179:13

SUMMARY: AddressSanitizer: heap-use-after-free (/home/alebastr/sources/xdg-desktop-portal-wlr/build-asan/xdg-desktop-portal-wlr+0x49b311) in __interceptor_strcmp.part.0
</pre>
</details>
Build command, for reference:

<code>env CC=clang CXX=clang++  meson setup build-asan --buildtype debug  -Db_lundef=false -Db_sanitize=address,undefined</code>

Will try normal build once my working day ends.

danshick:
Hey, thanks for testing this PR. It definitely still has some issues where it can get into strange states.

Your first ASan run indicates line number 259 in screencast.c, which in the latest commit is a closing brace. Not sure what to make of that.

I love ASan but sadly xdpw has often crashed when using ASan for an unknown reason. If that's no longer the case, id be very happy. Confused, but happy.

danshick:
For your second issue, maybe I need to initialize the cast local to NULL so the NULL pointer check right above it triggers. Didn't think i needed to do that.

danshick:
This PR does, in fact, fix #18, which in turn means that this now correctly supports obs-xdg-portal. Not sure this is in any way better than wlrobs, but it's another option.

alebastr:
:+1:
No more crashes and multiple simultaneous sessions are working. Got the same issue with pipewire though.
I believe the restart crash I reported was another case of use-after-free that you've just fixed.


> maybe I need to initialize the cast local to NULL so the NULL pointer check right above it triggers. Didn't think i needed to do that.

I personally prefer to initialize everything, just to be safe. And then remove redundant initializations where it's clear that all branches below are setting value to the variable before using it. Saved me a lot of time in the past.

danshick:
I haven't been able to figure out what exactly is wrong with pipewire. If you run it manually with the environment variable PIPEWIRE_DEBUG=3 (or even 4) and you manage to get it into a bad state, that log might help @wtay and others to improve it's stability. Our plugin shouldn't put it into an unstable state no matter what we're doing. I honestly can't even figure out when or how it gets that way.

I'd also love if someone could help me track down the issues with Chromium now. It's a strange failure case that I can't make sense of.

nostdm:
I tested both `chromium-dev-ozone` and regular `chromium` and it seems that the `xdg-desktop-portal-wlr` no longer crashes and stops/restarts properly. That said I can see output only in the thumbnail when picking the screen to share, but no output after that. 

The popup correctly shows a real-time thumbnail for the whole screen:

popup

After sharing the screen there is no output (I tried both https://meet.google.com and https://www.webrtc-experiment.com/Pluginfree-Screen-Sharing). Below screenshot shows no output with a loading wheel:

webrtc

danshick:
Thanks, that's helpful, and actually a totally different failure mode than I was experiencing. Pipewire has a permissions model that, until this point, I've been happy to ignore. I think the default is permissive. This looks like it might be related.

My experience with Chromium started similarly with the preview functioning, but my full screen share appeared garbled instead of not at all.

danshick:
@emersion I'm debating ignoring this Chromium issue for the purposes of this l
PR. It definitely deserves it's own issue and fix, but this is behaving correctly for any number of Firefox, xdp-screen-cast.py, and obs-xdg-portal sessions that I can throw at it.

src/screencast/screencast_common.c
```diff
@@ -1,18 +1,28 @@
 #include "screencast_common.h"
 
-enum spa_video_format pipewire_from_wl_shm(void *data) {
-	struct screencast_context *ctx = data;
+void randname(char *buf) {
```
emersion, line 3: Hmm, this is a little bit dangerous, since this relies on `buf` having at least 6 bytes. Maybe at least add `assert(buf[i] == 'X')` in the loop?

include/xdpw.h
```diff
@@ -7,23 +7,29 @@
...
 
 struct xdpw_state {
+	struct wl_list xdpw_sessions;
 	sd_bus *bus;
 	struct wl_display *wl_display;
 	struct pw_loop *pw_loop;
-
-	struct screencast_context screencast;
+	struct xdpw_screencast_context screencast;
+	enum source_types screencast_source_types;
```
emersion, line 19: enums should only take values enumerated in the definition. For bitfields, this isn't the case. Prefer something like:

```c
uint32_t screencast_source_types; // bitfield of enum source_types
```

src/core/main.c
```diff
@@ -83,6 +84,7 @@ int main(int argc, char *argv[]) {
 		logprint(ERROR, "wayland: failed to connect to display");
 		goto error;
 	}
+	logprint(INFO, "wlroots: wl_display fd: %d", wl_display_get_fd(wl_display));
```
emersion, line 87: Hmm, is the FD number useful?

src/core/main.c
```diff
@@ -92,13 +94,19 @@ int main(int argc, char *argv[]) {
 	}
 
 	struct xdpw_state state = {
+		.xdpw_sessions = (struct wl_list) { 0 },
```
emersion, line 97: No need to zero-initialize this field. Omitted fields are already zero-initialized.

src/core/request.c
```diff
@@ -3,13 +3,14 @@
...
 
 static const char interface_name[] = "org.freedesktop.impl.portal.Request";
 
 static int method_close(sd_bus_message *msg, void *data,
 		sd_bus_error *ret_error) {
-	
-	int ret = 0; 
+	struct xdpw_request *req = data;
+	int ret = 0;
 	// struct xdpw_request *req = data;
```
emersion, line 14: This comment can be removed.

src/core/session.c
```diff
@@ -36,24 +38,39 @@ static const sd_bus_vtable session_vtable[] = {
...
-	if (req == NULL) {
+void xdpw_session_destroy(struct xdpw_session *sess) {
+	logprint(TRACE, "dbus: destroying session");
+	if (!sess) {
 		return;
 	}
-	sd_bus_slot_unref(req->slot);
-	free(req);
+	struct xdpw_screencast_instance *cast = sess->screencast_instance;
+	if (cast){
```
emersion, line 64: Style nit: missing space between `)` and `{`

src/core/session.c
```diff
@@ -36,24 +38,39 @@ static const sd_bus_vtable session_vtable[] = {
...
+		--cast->refcount;
```
emersion, line 65: Add `assert(cast->refcount > 0)` right before as a safety net.

src/screencast/pipewire_screencast.c
```diff
@@ -128,66 +135,87 @@ static const struct pw_stream_events pwr_stream_events = {
...
 
-	return NULL;
+}
+
+void xdpw_pwr_core_connect(struct xdpw_state *state){
+	struct xdpw_screencast_context *ctx = &state->screencast;
+
+	logprint(TRACE, "pipewire: establishing connection to core");
+
+	if(!ctx->pwr_context){
```
emersion, line 199: Style nit: missing spaces between `if` and `(`, and between `)` and `{`

src/screencast/pipewire_screencast.c
```diff
@@ -128,66 +135,87 @@ static const struct pw_stream_events pwr_stream_events = {
...
+		ctx->pwr_context = pw_context_new(state->pw_loop, NULL, 0);
+		if (!ctx->pwr_context) {
+			logprint(ERROR, "pipewire: failed to create context");
+			abort();
```
emersion, line 203: Can we use a return value instead of aborting? Aborting is not great if it's caused by runtime errors.

src/screencast/pipewire_screencast.c
```diff
@@ -128,66 +135,87 @@ static const struct pw_stream_events pwr_stream_events = {
...
+		}
+	}
+}
+
+void xdpw_pwr_stream_destroy(struct xdpw_screencast_instance *cast){
+	logprint(TRACE, "pipewire: destroying stream");
+	pw_stream_flush(cast->stream, false);
+	pw_stream_disconnect(cast->stream);
+	pw_stream_destroy(cast->stream);
```
emersion, line 220: Maybe set `cast->stream = NULL` too, to make sure it doesn't get used-after-free by someone else?

src/screencast/screencast.c
```diff
@@ -1,61 +1,113 @@
...
+#include "wlr_screencast.h"
+#include "xdpw.h"
+#include "logger.h"
+
 static const char object_path[] = "/org/freedesktop/portal/desktop";
 static const char interface_name[] = "org.freedesktop.impl.portal.ScreenCast";
 
-int setup_outputs(struct xdpw_state *state) {
+void xdpw_screencast_instance_init(struct xdpw_screencast_context *ctx,
+		struct xdpw_screencast_instance *cast, struct xdpw_wlr_output *out, bool with_cursor){
```
emersion, line 23: Style nit: missing space between `)` and `{`

src/screencast/screencast.c
```diff
@@ -1,61 +1,113 @@
...
+	cast->framerate = out->framerate;
+	cast->with_cursor = with_cursor;
+	cast->refcount = 1;
+	logprint(INFO, "xdpw: screencast instance %p has %d references", cast, cast->refcount);
+	wl_list_insert(&ctx->screencast_instances, &cast->link);
+	logprint(INFO, "xdpw: %d active screencast instances",
+		wl_list_length(&ctx->screencast_instances));
+}
+
+void xdpw_screencast_instance_destroy(struct xdpw_screencast_instance *cast){
```
emersion, line 35: Style nit: missing space between `)` and `{`

src/screencast/screencast.c
```diff
@@ -1,61 +1,113 @@
...
+	logprint(TRACE, "xdpw: destroying cast instance");
```
emersion, line 36: Should this be called only when `cast->refcount == 0`? If so, would be nice to add an assertion.

src/screencast/screencast.c
```diff
@@ -1,61 +1,113 @@
...
-	ctx->with_cursor = true;
+	struct xdpw_screencast_instance *cast, *tmp_c;
+	wl_list_for_each_reverse_safe(cast, tmp_c, &ctx->screencast_instances, link) {
+		logprint(INFO, "xdpw: existing screencast instance: %d %s cursor",
+			cast->target_output->id,
+			cast->with_cursor ? "with" : "without");
 
-	logprint(INFO, "wlroots: output: %s", ctx->target_output->name);
-	logprint(INFO, "wlroots: wl_display fd: %d", wl_display_get_fd(state->wl_display));
+		if(cast->target_output->id == out->id && cast->with_cursor == with_cursor){
```
emersion, line 71: Style: missing spaces here as well

src/screencast/screencast.c
```diff
@@ -178,7 +229,22 @@ static int method_screencast_select_sources(sd_bus_message *msg, void *data,
 		} else if (strcmp(key, "types") == 0) {
 			uint32_t mask;
 			sd_bus_message_read(msg, "v", "u", &mask);
+			if(mask & (1<<WINDOW)){
```
emersion, line 232: Style: spaces here too

src/screencast/screencast.c
```diff
@@ -178,7 +229,22 @@ static int method_screencast_select_sources(sd_bus_message *msg, void *data,
...
 			logprint(INFO, "dbus: option types:%x", mask);
+		} else if (strcmp(key, "cursor_mode") == 0) {
+			uint32_t cursor_mode;
+			sd_bus_message_read(msg, "v", "u", &cursor_mode);
+			if (cursor_mode & (1<<HIDDEN)){
+				cursor_embedded = false;
+			}
+			if (cursor_mode & (1<<METADATA)){
+				logprint(ERROR, "dbus: unsupported cursor mode requested, exiting");
+				abort();
```
emersion, line 245: Send an error instead?

src/screencast/screencast.c
```diff
@@ -266,24 +338,40 @@ static int method_screencast_start(sd_bus_message *msg, void *data,
...
 	}
 
-	sd_bus_message *reply = NULL;
-	ret = sd_bus_message_new_method_return(msg, &reply);
-	if (ret < 0) {
-		return ret;
+	struct xdpw_screencast_instance *cast = NULL;
+	struct xdpw_session *sess, *tmp_s;
+	wl_list_for_each_reverse_safe(sess, tmp_s, &state->xdpw_sessions, link){
+		if(strcmp(sess->session_handle, session_handle) == 0){
```
emersion, line 344: Style: missing spaces

src/screencast/wlr_screencast.c
```diff
@@ -1,31 +1,72 @@
...
 #include "xdpw.h"
+#include "logger.h"
 
-void wlr_frame_free(struct xdpw_state *state) {
-	zwlr_screencopy_frame_v1_destroy(state->screencast.wlr_frame);
-	munmap(state->screencast.simple_frame.data, state->screencast.simple_frame.size);
-	wl_buffer_destroy(state->screencast.simple_frame.buffer);
+void xdpw_wlr_frame_free(struct xdpw_screencast_instance *cast) {
+	zwlr_screencopy_frame_v1_destroy(cast->wlr_frame);
+	munmap(cast->simple_frame.data, cast->simple_frame.size);
```
emersion, line 26: We should reset all these free'ed fields to NULL/zero, to make sure we don't use-after-free and make debugging easier.

src/screencast/wlr_screencast.c
```diff
@@ -1,31 +1,72 @@
...
 
-	if (!state->screencast.quit && !state->screencast.err) {
-		wlr_register_cb(state);
+	if (cast->quit || cast->err) {
+		xdpw_screencast_instance_destroy(cast);
+		return ;
 	}
+
+	xdpw_wlr_register_cb(cast);
+
```
emersion, line 37: Style: stray newline

danshick:
Thanks for taking the time to review. I agree across the board.

One of these days, I'm going to develop the muscle memory to start including spaces around my conditionals (or strong arm my editor into doing it on save).

lsjostro:
> I tested both `chromium-dev-ozone` and regular `chromium` and it seems that the `xdg-desktop-portal-wlr` no longer crashes and stops/restarts properly. That said I can see output only in the thumbnail when picking the screen to share, but no output after that.
> 
> The popup correctly shows a real-time thumbnail for the whole screen:
> 
> popup
> 
> After sharing the screen there is no output (I tried both https://meet.google.com and https://www.webrtc-experiment.com/Pluginfree-Screen-Sharing). Below screenshot shows no output with a loading wheel:
> 
> webrtc
> 
> I also occasionally get this error from `xdg-desktop-portal`:
> 
> ```
> [E][000054978.592812][core.c:71 core_event_error()] core 0x7f20e403ff00: proxy 0x7f20e403ff00 id:0: seq:5 res:-13 (Permission denied) msg:"no permission to call method 2 on 0 (requires 00000040, have 00000000)"
> ```
I get the same above with chrome/chromium.
here some more logs from pipewire

`systemctl --user status pipewire`

```
[W][000001265.154665][impl-node.c:327 suspend_node()] node 0x559a50dd4c10: error unset format input: Input/output error
[W][000001265.162238][impl-node.c:327 suspend_node()] node 0x559a50dbd580: error unset format input: Input/output error
[W][000001265.190417][module-protocol-native.c:370 client_new()] server 0x559a50cd9a90: no peersec: Protocol not available
[W][000001265.190832][impl-client.c:608 pw_impl_client_update_permissions()] client 0x559a50db24c0: invalid global 104495344
[W][000001265.192569][connection.c:311 prepare_packet()] old version detected
[E][000001265.346362][impl-link.c:118 pw_impl_link_update_state()] link 0x559a50dd8010: update state negotiating -> error (no more input formats)

```

danshick:
Hmmm, most of that is safely ignorable, but that last line of pipewire output may be a hint. I'm not sure why, but it looks like we can't find an agreeable format in common? Weird because the WebRTC code is common to Firefox and Chromium. I'll open a new issue once this PR lands and debug it further.

lsjostro:
created a gist with full pipewire debug log.
look at the end for relavent error

https://gist.github.com/lsjostro/f08398b120a2309af36d000a0aa3c2c7

src/core/main.c
danshick: No, I was probably debugging something long before I added decent logging, and mindlessly s/printf(/logprint(INFO,/

abmantis:
@danshick Just as a note (couldn't find the issue you mentioned you would open): I had the same problem of black screen in chrome, and was able to make it work by specifying the -p option (either RGBx or BGRx work, but the first makes the colors funky).

danshick:
@abmantis Thanks, this is a known issue. A permanent fix is forthcoming both here and WebRTC (we will both start being more flexible with supported formats). I'm also adding significantly more documentation in the mean time that will mention this. Check out the current open PR to see those new docs and tell me what you think.

Glad you got it working. Hit me up on IRC if you have any issues. #sway on freenode.

Issue #28, opened 2020-04-22
Compile errors on Alpine Linux musl


codebam:
```
komodo:~/Packages/xdg-desktop-portal-wlr$ ninja -C build
[1/9] Compiling C object 'xdg-desktop-portal-wlr@exe/src_screencast_pipewire_screencast.c.o'
[2/9] Compiling C object 'xdg-desktop-portal-wlr@exe/src_screencast_screencast_common.c.o'
[3/9] Compiling C object 'xdg-desktop-portal-wlr@exe/src_screenshot_screenshot.c.o'
[4/9] Compiling C object 'xdg-desktop-portal-wlr@exe/src_core_session.c.o'
[5/9] Compiling C object 'xdg-desktop-portal-wlr@exe/src_core_request.c.o'
[6/9] Compiling C object 'xdg-desktop-portal-wlr@exe/src_core_main.c.o'
ninja: job failed: cc -Ixdg-desktop-portal-wlr@exe -I. -I.. -I../include -Iprotocols -I/usr/include/elogind -I/usr/include/pipewire-0.3 -I/usr/include/spa-0.2 -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Werror -std=c11 -g -Wno-missing-braces -Wno-missing-field-initializers -Wno-unused-parameter -DHAVE_SYSTEMD=1 -D_REENTRANT -MD -MQ 'xdg-desktop-portal-wlr@exe/src_screencast_screencast_common.c.o' -MF 'xdg-desktop-portal-wlr@exe/src_screencast_screencast_common.c.o.d' -o 'xdg-desktop-portal-wlr@exe/src_screencast_screencast_common.c.o' -c ../src/screencast/screencast_common.c
In file included from /usr/include/spa-0.2/spa/support/loop.h:34,
                 from /usr/include/pipewire-0.3/pipewire/loop.h:32,
                 from /usr/include/pipewire-0.3/pipewire/context.h:65,
                 from /usr/include/pipewire-0.3/pipewire/protocol.h:36,
                 from /usr/include/pipewire-0.3/pipewire/proxy.h:103,
                 from /usr/include/pipewire-0.3/pipewire/client.h:35,
                 from /usr/include/pipewire-0.3/pipewire/pipewire.h:35,
                 from ../include/screencast_common.h:4,
                 from ../src/screencast/screencast_common.c:1:
[...]

PR #29, opened 2020-04-22
Fix musl issue with time.h

danshick:
This should fix our issues with musl (#28), but pipewire spa code is still misbehaving.


emersion:
@codebam: does it fix your issue?

codebam:
Answered on IRC, but yes this fixes the compile issue as long as -Werror is turned off in the meson.build.

danshick:
Updated to add the _POSIX_C_SOURCE to the meson.build and remove it from all files. Ready for merge.

Issue #23, opened 2020-04-14
screencast: remove force pixel format option


emersion:
This is a hack used to make some broken clients work. Clients should be fixed instead.

emersion:
Or maybe we could have a less dangerous flag like `--strip-alpha`.

danshick:
I've been chatting with @jgrulich on #pipewire. He's the original implementer in WebRTC. He was under the impression that BGRx or RGBx would accept alpha channel metadata by default, but we confirmed it does not.

Anyway, he's working on some refactoring (and eventually pipewire 0.3 support) in the WebRTC codebase and will include a fix for these pixel formats being rejected as well as correcting support for framerates other than a variable framerate.

In the meantime though, I agree with you, we can at least modify the flag to be less of a hack.

emersion:
Maybe we can advertise support for both the format with alpha and the format without alpha? This wouldn't require any flag.

danshick:
I thought about doing this in the beginning but I didn't want to introduce any format conversion here, and I wanted to accurately report what wlroots provided as the default. I suppose it's not too far off to offer either A or x if the x literally means "these bits are present, but we ignore them".

emersion:
I agree we don't want to do any format conversion, but yeah stripping alpha doesn't require any format conversion. We already need to maintain a format table so I don't think it'd be a big deal.

danshick:
I don't even think we need to change any bytes, because that last byte of color information is just ignored. Is there any spec that says those bytes must be zeroed for proper format compliance?

emersion:
No, the X means "it can be anything". We already do something similar in wlroots' DRM backend: https://github.com/swaywm/wlroots/blob/904c37845dec60fc5ee91d11b5e6c75656651b33/backend/drm/drm.c#L337

PR #33, opened 2020-05-03
Advertise format with alpha channel stripped

emersion:
Tested with Chromium.

Closes: https://github.com/emersion/xdg-desktop-portal-wlr/issues/23

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -23,7 +23,6 @@ int xdpw_usage(FILE* stream, int rc) {
 		"    -l, --loglevel=<loglevel>        Select log level (default is ERROR).\n"
 		"                                     QUIET, ERROR, WARN, INFO, DEBUG, TRACE\n"
 		"    -o, --output=<name>              Select output to capture.\n"
-		"    -p,--pixelformat=BGRx|RGBx       Force a pixelformat in pipewire\n"
 		"                                     metadata (performs no conversion).\n"
 		"    -h,--help                        Get help (this text).\n"
 		"\n";
@@ -34,14 +33,12 @@ int xdpw_usage(FILE* stream, int rc) {
 
 int main(int argc, char *argv[]) {
 	const char* output_name = NULL;
-	const char* forced_pixelformat = NULL;
 	enum LOGLEVEL loglevel = ERROR;
 
 	static const char* shortopts = "l:o:p:h";
 	static const struct option longopts[] = {
 		{ "loglevel", required_argument, NULL, 'l' },
 		{ "output", required_argument, NULL, 'o' },
-		{ "pixelformat", required_argument, NULL, 'p' },
 		{ "help", no_argument, NULL, 'h' },
 		{ NULL, 0, NULL, 0 }
 	};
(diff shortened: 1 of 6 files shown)


danshick:
I'll give this a once over first thing tomorrow.

soyuka:
Works on chromium without `-p BGRx` on my end.

danshick:
Unfortunately, this seems to break clients asking for BGRA. Not sure why yet.

src/screencast/pipewire_screencast.c
```diff
@@ -163,11 +162,20 @@ void xdpw_pwr_stream_init(struct xdpw_screencast_instance *cast) {
...
+		n_formats++;
+	}
+
+	const struct spa_pod *param = spa_pod_builder_add_object(&b,
 		SPA_TYPE_OBJECT_Format, SPA_PARAM_EnumFormat,
 		SPA_FORMAT_mediaType,       SPA_POD_Id(SPA_MEDIA_TYPE_video),
 		SPA_FORMAT_mediaSubtype,    SPA_POD_Id(SPA_MEDIA_SUBTYPE_raw),
-		SPA_FORMAT_VIDEO_format,    SPA_POD_Id(xdpw_format_pw_from_wl_shm(cast)),
+		SPA_FORMAT_VIDEO_format,    SPA_POD_CHOICE_ENUM_Id(n_formats,
+			format, format_without_alpha),
```
danshick, line 178: Okay, this is why it isn't working any more for BGRA clients. This needs to be:
```c
SPA_FORMAT_VIDEO_format,    SPA_POD_CHOICE_ENUM_Id(
		++n_formats, // where this is incremented because we now have three arguments
		format, // where this indicates a default choice from a list to follow
		format, // where this begins the list of choices
		format_without_alpha), // where this is the last element of the list
```

danshick: Basically, we were saying that the unstripped format was the default, but then not including it in the list of allowed values, which only contained the stripped version. That allowed it to work in Chromium.

I'm a little surprised that the builder doesn't complain about this.

danshick:
By the way, if we ever need to test a specific pixel format, you can use a capsfilter with the gstreamer plugin. I've been testing with xdg-screen-cast.py and changing the pipeline definition like this...

```py
pipeline = Gst.parse_launch('pipewiresrc fd=%d path=%u ! video/x-raw,format=BGRA ! videoconvert ! xvimagesink'%(fd, node_id))
```
...in order to explicitly test BGRA.

src/core/main.c
```diff
@@ -58,9 +55,6 @@ int main(int argc, char *argv[]) {
 		case 'o':
 			output_name = optarg;
 			break;
-		case 'p':
-			forced_pixelformat = optarg;
```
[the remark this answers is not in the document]
emersion: Nah, I prefer things to break hard when they don't do what you expect.

src/screencast/pipewire_screencast.c
emersion: Ah, good catch, thanks!

src/core/main.c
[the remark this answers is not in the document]
danshick: Hah, okay. Not like we have a huge user base yet.

danshick:
LGTM

PR #35, opened 2020-05-21
core/main.c: Fix error handling

maribu:
The error handling at the `error:` label tears down the whole state. Thus, the state needs to be fully initialized in order for the tear down to succeed. Currently, if e.g. `sd_bus_open_user()` fails, a `segfault` is triggered by the tear down. This commit adds individual tear down code that only touches
stuff that until that point was successfully initialized.

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -70,7 +70,7 @@ int main(int argc, char *argv[]) {
 	ret = sd_bus_open_user(&bus);
 	if (ret < 0) {
 		logprint(ERROR, "dbus: failed to connect to user bus: %s", strerror(-ret));
-		exit(EXIT_FAILURE);
+		return EXIT_FAILURE;
 	}
 	logprint(DEBUG, "dbus: connected");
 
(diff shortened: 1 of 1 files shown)


src/core/main.c
```diff
@@ -70,22 +70,25 @@ int main(int argc, char *argv[]) {
 	ret = sd_bus_open_user(&bus);
 	if (ret < 0) {
 		logprint(ERROR, "dbus: failed to connect to user bus: %s", strerror(-ret));
-		goto error;
+		exit(EXIT_FAILURE);
```
emersion, line 73: We can just `return EXIT_FAILURE` instead of calling `exit`

maribu: Done

danshick:
There's probably other places where we bail out entirely when we could recover more gracefully. This PR looks good to me, but I should spend some time cleaning up the error handling throughout.

maribu:
You're welcome. Thanks for developing xdpw

PR #37, opened 2020-05-26
core/main: Clean up option handling

maribu:
- Add a missing space in the usage string
- Drop short option "p", which is unused


danshick:
LGTM, thanks.

PR #43, opened 2020-08-04
screencast/wlr_screencast Reuse wl_buffer if possible

columbarius:
This draft tries to minimize creation and destruction of the cast->simple_frame.buffer and cast->simple_frame.data.

### DIFF
--- a/src/screencast/wlr_screencast.c
+++ b/src/screencast/wlr_screencast.c
@@ -105,10 +105,12 @@ static void wlr_frame_buffer(void *data, struct zwlr_screencopy_frame_v1 *frame,
 
 	logprint(TRACE, "wlroots: buffer event handler");
 	cast->wlr_frame = frame;
-	if (cast->simple_frame.buffer == NULL || cast->simple_frame.width != width || cast->simple_frame.height != height || cast->simple_frame.stride != stride || cast->simple_frame.format != format) {
+	if (cast->simple_frame.buffer == NULL || cast->simple_frame.data == NULL || cast->simple_frame.width != width || cast->simple_frame.height != height || cast->simple_frame.stride != stride || cast->simple_frame.format != format) {
 		logprint(TRACE, "wlroots: buffer properties changed");
-		munmap(cast->simple_frame.data, cast->simple_frame.size);
-		cast->simple_frame.data = NULL;
+		if (cast->simple_frame.data != NULL) {
+			munmap(cast->simple_frame.data, cast->simple_frame.size);
+			cast->simple_frame.data = NULL;
+		}
 		cast->simple_frame.width = width;
 		cast->simple_frame.height = height;
 		cast->simple_frame.stride = stride;
--- a/src/screencast/wlr_screencast.c
+++ b/src/screencast/wlr_screencast.c
@@ -28,8 +28,9 @@ void xdpw_wlr_frame_free(struct xdpw_screencast_instance *cast) {
 		cast->simple_frame.data = NULL;
 		wl_buffer_destroy(cast->simple_frame.buffer);
 		cast->simple_frame.buffer = NULL;
-		logprint(TRACE, "wlroots: frame destroyed");
+		logprint(TRACE, "xdpw: simple_frame buffer destroyed");
 	}
+	logprint(TRACE, "wlroots: frame destroyed");
 
 	if (cast->quit || cast->err) {
 		// TODO: revisit the exit condition (remove quit?)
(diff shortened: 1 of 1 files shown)


danshick:
Thanks for the PR. I'll take a look over the next couple of days. @emersion may also have some comments or reviews.

danshick:
Haven't forgotten about you. It's been a crazy week/weekend.

Also, @wtay posted this example regarding renegotiation...https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/examples/video-src-reneg.c

src/screencast/wlr_screencast.c
```diff
@@ -97,19 +108,36 @@ static struct wl_buffer *create_shm_buffer(struct xdpw_screencast_instance *cast
...
+		logprint(TRACE, "wlroots: buffer properties changed");
+		wlr_frame_buffer_chparam(cast, format, width, height, stride);
+	}
+
+	if (cast->simple_frame.buffer == NULL) {
+		logprint(DEBUG, "wlroots: create shm buffer");
+		cast->simple_frame.buffer = create_shm_buffer(cast, format, width, height,
+			stride, &cast->simple_frame.data);
+	} else {
+		logprint(DEBUG,"wlroots: shm buffer exists");
```
danshick, line 139: Can you make this log a TRACE level log? This will print every single frame where the params haven't changed. I'm trying to keep "per frame" logging events in TRACE.

danshick:
Other than that, I think this looks good. I'll run this fork for the next week and if I don't seen any stability issues, I'll approve. Thanks for your patience.

src/screencast/wlr_screencast.c
columbarius: Shure. This log was only for me for testing purposes, since I wasn't shure, if something should be happening in the other case. Since in the else part nothing is happening and the if part has a log, we also could just remove the whole else part.

columbarius:
No worries. Thanks for your answer an the link from @wtay.
Thanks for testing, I'm excited about the results. No problem.

danshick:
We follow the wlroots [CONTRIBUTING](https://www.github.com/swaywm/wlroots/tree/master/CONTRIBUTING.md) guidelines. The terms finish and destroy have specific meaning that might be useful here.

columbarius:
> We follow the wlroots [CONTRIBUTING](https://www.github.com/swaywm/wlroots/tree/master/CONTRIBUTING.md) guidelines. The terms finish and destroy have specific meaning that might be useful here.

Thanks, I changed the name to destroy.

src/screencast/wlr_screencast.c
```diff
@@ -19,14 +19,27 @@
...
+		cast->simple_frame.buffer = NULL;
+	}
+}
+
 void xdpw_wlr_frame_free(struct xdpw_screencast_instance *cast) {
 	zwlr_screencopy_frame_v1_destroy(cast->wlr_frame);
 	cast->wlr_frame = NULL;
-	munmap(cast->simple_frame.data, cast->simple_frame.size);
-	cast->simple_frame.data = NULL;
 	// TODO: reuse this buffer unless we quit or error out
```
emersion, line 38: This TODO can be removed, correct?

columbarius: Sure. I would remove it, when this draft changed to a PR.

danshick:
> From a quick glance, looks good, although I haven't tested.

I've been running this branch for the past two days and so far so good. I'll continue doing so in the background to ensure no strange crashes.

src/screencast/wlr_screencast.c
danshick: Makes sense. If it should be unreachable, it's probably a good idea to put an abort in there, although it doesn't seem like it should be a failure case. I'll have to give it another look.

src/screencast/wlr_screencast.c
danshick: Go ahead and pull this todo. I think we're good to merge.

src/screencast/wlr_screencast.c
```diff
@@ -19,14 +19,26 @@
 #include "xdpw.h"
 #include "logger.h"
 
+static void wlr_frame_buffer_destroy(struct xdpw_screencast_instance *cast) {
+	// This check isn't needed
```
danshick, line 23: If this check isn't necessary, you can just remove it. You can leave a comment below explaining that wl_buffer_destroy will not work on NULL if you like.

columbarius: Ok, I've changed it. Thanks.

src/screencast/wlr_screencast.c
columbarius: Todo removed.

danshick:
@emersion I'm good if you are.

emersion:
Thanks!

PR #44, opened 2020-08-06
systemd: ensure service is stopped on exit of graphical session

maximbaz:
I'm learning more about systemd 🙂 

`PartOf=` will ensure that when graphical session ends, this service will be gracefully shutdown. Otherwise display just disappears, and I think the process will crash.

`After=` is only needed to give systemd a hint about order of starting services, I don't think it's strictly necessary for a DBUS-activated service like this one or mako, I guess people add it mostly for semantic reasons.

`[Install]` is not really needed for a DBUS-activated service, because it will start even if you don't `systemctl enable` it.

However if you want I can add it (like in mako), and the effect will be that if a user explicitly `systemctl enable`s this service, it will be launched not on DBUS, but immediately after graphical session starts.

### DIFF
--- a/contrib/systemd/xdg-desktop-portal-wlr.service.in
+++ b/contrib/systemd/xdg-desktop-portal-wlr.service.in
@@ -1,5 +1,7 @@
 [Unit]
 Description=Portal service (wlroots implementation)
+PartOf=graphical-session.target
+After=graphical-session.target
 
 [Service]
 Type=dbus


danshick:
I'm not sure many people start sway with systemd, so I'd be concerned that graphical-session.target isn't triggered properly for many who launch from a TTY.

Would this change affect users who don't trigger that target?

maximbaz:
Just tested, DBUS still starts this service even though my graphical-session.target is not active.

By the way, I also don't start sway with systemd, but as I'm learning now if you want to use systemd services together with sway, it is really beneficial to add `sway-session.target` and start it from your sway config. If you are interested to learn more, I invite you to check out [sway wiki](https://github.com/swaywm/sway/wiki/Systemd-integration) and [some of my notes](https://github.com/maximbaz/dotfiles/issues/23).

WhyNotHugo:
> I'm not sure many people start sway with systemd, so I'd be concerned that graphical-session.target isn't triggered properly for many who launch from a TTY.

`graphical-session.target` should be started regardless of whether `sway` is run as a systemd service or not.

Personally, I run it via `.profile`, but the target is still activated by `sway`. This is the [recommended approach](https://github.com/swaywm/sway/wiki/Systemd-integration#managing-user-applications-with-systemd), and also aligns with how systemd services are intended to be managed.

danshick:
Sure, that may be true, but while recommended, it isn't common to all sway users. Personally, I don't much care to have systemd manage processes specific to my graphical session, and I know I'm not unique in that regard, especially amongst sway users.

As long as d-bus still activates xdpw correctly, regardless of the target reached, I don't have any objection to this change.

It may also be useful to update our documentation to point to the part of the sway wiki that details graphical-session.target activation, for users who do want to take this approach.

WhyNotHugo:
> Sure, that may be true, but while recommended, it isn't common to all sway users. Personally, I don't much care to have systemd manage processes specific to my graphical session, and I know I'm not unique in that regard, especially amongst sway users.

Oh, these changes should not break anything for this scenario. It only forces an already running portal portal to exit when `graphical-session.target` explicitly transitions from running to stopped, which only happens if you're actually using it.

PR #46, opened 2020-08-14
Pipewire: Renegotiate stream params on changed buffer size

columbarius:


columbarius:
It's possible with the current changes to change the screen resolution and the pipewire stream is adapting to that. Sadly on my part I was only able to capture the stream with the [pipewire video-play example](https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/examples/video-play.c). Beware the blue colouring, probably something about the colourspaces. Since the gstreamer plugin currenty doesn't support changing the format, the [xdg-screencast.py](https://gitlab.gnome.org/snippets/19) script freezes and doesn't continues with screencasting once the resolution is set to the state before, which is a regression. Currently I have no working webrtc setup, so i couldn't test that.

columbarius:
I've checked it with the pipewire enabled firefox from fedora and the webrtc stream freezed at change of resolution. Resetting the resolution didn't resume the stream neither on this branch nor master. So I will halt working on this branch until there is a more realworld consumer than the [video-play.c](https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/examples/video-play.c) example to test this on.

columbarius:
Checked today again with pipewire 0.3.10 and the fedora build of Firefox 80.0 and changing the screenresolution worked. 
Using the gstreamer plugin with: ``% PIPEWIRE_NODE=xx gst-launch-1.0 pipewiresrc ! videoconvert ! autovideosink`` still does some weird things, depending on when it's added to the stream, like showing a green image, freezing, or throwing an error:
> WARNING: Element /GstPipeline:pipeline0/GstVideoConvert:videoconvert0: Internal GStreamer error: code not implemented.  Please file a bug at http://bugzilla.gnome.org/enter_bug.cgi?product=GStreamer.
> Additional Debuginformation: ../gst-plugins-base/gst-libs/gst/video/gstvideofilter.c(292): gst_video_filter_transform (): /GstPipeline:pipeline0/GstVideoConvert:videoconvert0: invalid video buffer received

columbarius:
I'm using this for a while now in my testing build and haven't had any issues in normal usage. On the client side, some clients support renegotiation of buffers (eg. firefox), others don't, but break anyway. 
Should I remove the WIP tag, since it doesn't really need more work from our side, aside from rebasing?

emersion:
If you think it's ready for review, feel free to remove the WIP flag, yes.

danshick:
> Edit: If you switch the resolution of your montitor too fast, something jams up the transport in pipewire

What do you mean by too fast? Like more than once, very close together?

columbarius:
@danshick sry i missed your answer.

> What do you mean by too fast? Like more than once, very close together?

I mean that pipewire can segfault in pw_stream_queue_buffer ([pipewire issue produced with this patch](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/740)) if you try to queue an old frame with the wrong dimensions after the frame size was renegotiated. I would guess gnome didn't found this issue since they use full desktop frames with cropping for windows.

emersion (approved):
LGTM as well, thanks!

PR #48, opened 2020-08-23
[WIP]: dmabuf support

columbarius:
This Draft is going to implement screen sharing via wlr-export-dmabuf-unstable-v1    #9.
The first step will be to move from the current shm implementation to a protocol agnostic one with different choosable protocols.
After that the above mentioned protocol will be added.

This Draft is currently to display and discuss the current progress and will not be compilable at every time. 

It contains 3 or 4 steps marked by dependent branches:
* Update screencopy to version 3 #53 https://github.com/columbarius/xdg-desktop-portal-wlr/tree/screencopy-interface-v3
* Separate screencopy from screencast code https://github.com/columbarius/xdg-desktop-portal-wlr/tree/seperate-screencopy
* Adding dmabuf support for the screencopy protocol https://github.com/columbarius/xdg-desktop-portal-wlr/tree/dmabuf
* Test the viability of screencast via the export dmabuf protocol


columbarius:
Almost finished seperating the shm instance code from the generic one.
Could you please take a look at the creation of the spa_pod param? I first tried to but the initialisation into a seperate function, but couldn't get it to work, so I changed it to static inline functions. I'm not shure, if this is good style, or if I should create an backend independent xdpw_frame_geometry struct, which each backend can export.
That's the original one: https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/src/screencast/pipewire_screencast.c#L116
That's my try with "normal" functions: https://github.com/columbarius/xdg-desktop-portal-wlr/blob/857ff4dd1a5d27eb5cd2173e22f383f0a934f685/src/screencast/pipewire_screencast_scp_shm.c#L52
That's how I got it working: https://github.com/columbarius/xdg-desktop-portal-wlr/blob/dmabuf/include/pipewire_screencast_scp_shm.h#L13

columbarius:
Currently I use the strlcpy function from bsd as it is used in wayvnc. I'm not a huge fan, of using a string function not provided by the stdlibs of the os. If you have a preferred function to copy strings, I'm happy to use that.

columbarius:
@emersion This build probably needs at least libdrm to build on alpine. Would it be possible to add that?

danshick:
You can use strdup, which is nice and safe, and is available everywhere.

As for libdrm, we had that dependency once upon a time and it should be fine to add. Just update the meson config and the .build.yml file.

columbarius:
@danshick All right, thanks.

src/screencast/screencast.c
```diff
@@ -25,6 +25,7 @@ void xdpw_screencast_instance_init(struct xdpw_screencast_context *ctx,
 	cast->framerate = out->framerate;
 	cast->with_cursor = with_cursor;
 	cast->refcount = 1;
+	cast->type = XDPW_INSTANCE_SCP_SHM;
```
columbarius, line 28: Currently this declaration is used to change between shm screencopy and dmabuf screencopy.

columbarius:
I found from the wlroots [screencop-dmabuf.c](https://github.com/swaywm/wlroots/blob/master/examples/screencopy-dmabuf.c) example the option to create a linear dmabuf... which didn't work with pipewire (on intel).
So I added inspired from the same example an instance type, which prepares the data for pipewire in ram ... I guess that's not the type of dmabuf copy we want to achieve, but now we have a way to test the exported dmabuf.

columbarius:
I realised, we messed up the difference between the modifier and the format.
The format describes the colour layout and the modifier describes the specific hardware modifications for this layout.

https://github.com/wayland-project/wayland-protocols/blob/master/unstable/linux-dmabuf/linux-dmabuf-unstable-v1.xml#L144
https://github.com/wayland-project/wayland-protocols/blob/master/unstable/linux-dmabuf/linux-dmabuf-unstable-v1.xml#L202

danshick:
> So I added inspired from the same example an instance type, which prepares the data for pipewire in ram ... I guess that's not the type of dmabuf copy we want to achieve, but now we have a way to test the exported dmabuf.

Does this offer any advantage over screencopy alone? If this is just an experiment, that's fine, but I don't think it'll be merged.

> I realised, we messed up the difference between the modifier and the format.
> The format describes the colour layout and the modifier describes the specific hardware modifications for this layout.

Where did we mess this up? At the moment, the color format is being handled properly AFAIK.

columbarius:
> Does this offer any advantage over screencopy alone? If this is just an experiment, that's fine, but I don't think it'll be merged.

I guess not really. I don't know how to benchmark it to be sure, but it's more like a workaround for testing. I probably move it to another branch I guess?

> Where did we mess this up? At the moment, the color format is being handled properly AFAIK.

The color format yes, but the modifier is not handled properly.

danshick:
The modifier is meaningless for traditional screencopy. It is always a linear frame.

columbarius:
All this commit was missing was the Flag
PW_STREAM_FLAG_ALLOC_BUFFERS

Will rebase this pr and add that flag ...

columbarius:
One possible option to solve temporarily afaik would be to use the modifier format parameter to determine if a consumer uses mmap for dmabufs. If a client or pipewire (PW_FLAG_MAP_BUFFER) uses mmap, they should set the modifier to 0 (This has to be proposed to pipewire). In that case, we could just deliver MemPtr (if supported, but that should be the case every time),  mapped by us.

danshick:
> In that case, we could just deliver MemPtr (if supported, but that should be the case every time),  mapped by us. 

I think I need to recap. On the wlroots side, we have screencopy, screencopy-dmabuf, and export-dmabuf. In the memptr case (which I believe is what we do right now), we would be forced to continue using screencopy? Or am I misunderstanding memptr?

Of course, if we see support for modifiers, we can supply a dmabuf fd from export-dmabuf (or screencopy-dmabuf, if we're concerned about a race condition).

I guess what I'd like to understand is how the [buffer types](https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/spa/include/spa/buffer/type-info.h#L47) correspond with the various flags that can be set, and what the corresponding mapping behavior should be.

columbarius:
Sure.
Wlroots:
On the wlroots side we can "export" the screenbuffer in 3 different ways:

* screencopy: Create an annonymous file descriptor in ram and let the compositor copy the buffer to it.
* screencopy-dmabuf: Create an dmabuf linked via an file descriptor on the gpu and let the compositor copy the buffer to it.
* export-dmabuf: Let the compositor just export the dmabuf file descriptor of the output buffer.

Pipewire:
Pipewire supports for us 3 relevant buffer types: MemPtr, MemFd and DmaBuf (all declared by SPA_DATA_*).

* MemPtr: If you use the PW_STREAM_FLAG_MAP_BUFFERS Flag as a producer, pipewire will allocate internally memory via an file descriptor and maps it to pw_buffer buffer->buffer->datas[].data. You just get a buffer with mapped memory, ready to copy the content. 
If you want another buffer type PW_STREAM_FLAG_MAP_BUFFERS is not an option, you have to use PW_STREAM_FLAG_ALLOC_BUFFERS and do the buffer allocation on your own, that's where we need:

* MemFd: Create a filedescriptor to shared memory via create_memfd, or the trick from emersion with the wl_shm. We can map this buffer and copy our buffer information into it as usual, or just send the file descriptor where screencopy put it's content in (just thought about it, should try it later, could be racy). On the client side, you receive either a MemFd or a MemPtr, second if you let pipewire do the mapping for you (PW_STREAM_FLAG_MAP_BUFFERS).

* DmaBuf: This just let's us send the file descriptor to the underlying dmabuf. Client will receive the same file descriptor as with MemFd, but knows, it should treat it as an dmabuf (should ... sigh). If the client uses PW_STREAM_FLAG_MAP_BUFFERS, pipewire will mmap the file descriptor to the data pointer and the client will receive an MemPtr buffer.

So long story short: 
MemPtr and MemFd are interchangeable, since both are mmaped/mmapable areas of memory in ram to transport buffers.
DmaBuf: Tells the client, the file descriptor they receive should not be treated as an shared memory file descriptor, but as memory on the graphics card.

Even if we use screencopy dmabuf, we can either create a linear buffer or one with a modifier. For now i only have seen clients either support full modifiers (obs, either wlrobs or obs-xdg-portal (currently merging a patch)), or none (which means mmaping dmabuf). In the none case i would guess, we would get better results, if we do the transition from dmabuf to shm and map the dmabuf with gbm and copy the content to shared memory. We can't just switch to the screencopy backend in that case, since the screencopy buffer might have another format and we would need to do the whole negotiation again from start, hoping it might work.

danshick:
>     * DmaBuf: This just let's us send the file descriptor to the underlying dmabuf. Client will receive the same file descriptor as with MemFd, but knows, it should treat it as an dmabuf (should ... sigh). If the client uses PW_STREAM_FLAG_MAP_BUFFERS, pipewire will mmap the file descriptor to the data pointer and the client will receive an MemPtr buffer.

Do we know why pipewire is still offering to mmap dmabufs? Is there any case in which this is desirable? Also, doesn't the client have to *opt-in* to dmabuf buffers for negotiation purposes? Shouldn't the MAP_BUFFERS flag preclude the client from negotiating a dmabuf format successfully? (I'm only asking here because I know you've had more discussions with Wim on this than me)

> MemPtr and MemFd are interchangeable, since both are mmaped/mmapable areas of memory in ram to transport buffers.

Okay, this is the part I was most confused about. So a producer can claim MemFd with the ALLOC_BUFFERS flag, but there's no reason that a consumer couldn't then request MemPtr on its end and let pipewire do the mapping.

> Even if we use screencopy dmabuf, we can either create a linear buffer or one with a modifier. For now i only have seen clients either support full modifiers (obs, either wlrobs or obs-xdg-portal (currently merging a patch)), or none (which means mmaping dmabuf). In the none case i would guess, we would get better results, if we do the transition from dmabuf to shm and map the dmabuf with gbm and copy the content to shared memory.

So if we were to use screencopy-dmabuf exclusively and used gbm to map it linearly, we would still have to call this a dmabuf buffer on the pipewire side, because even in linear, it still isn't kosher to mmap a dma-buf ([without fencing?](https://01.org/linuxgraphics/gfx-docs/drm/driver-api/dma-buf.html#fence-poll-support), I remember daniels mentioning this sowewhere).

> Tangent about changing wroots protocol on the fly...

I really think we need to discuss with Wim about some way to allow negotiation to involve a round trip. Right now, a producer cannot change much about what it offers if the client doesn't support renegotiation. It'd be great to be able to signal to pipewire (maybe inside of the param_changed callback) that we would like it to hint to the consumer to try a different pipewire node. We could set up a node for each protocol, start with best possible, and continually fall back until we can negotiate properly. This should be easy for consumers to implement too, it's just a retry, and consumers that don't listen to this hint could still fail like they do today, so it should also be backwards compatible.

columbarius:
> > ```
> > * DmaBuf: This just let's us send the file descriptor to the underlying dmabuf. Client will receive the same file descriptor as with MemFd, but knows, it should treat it as an dmabuf (should ... sigh). If the client uses PW_STREAM_FLAG_MAP_BUFFERS, pipewire will mmap the file descriptor to the data pointer and the client will receive an MemPtr buffer.
> > ```
> 
> Do we know why pipewire is still offering to mmap dmabufs? Is there any case in which this is desirable? Also, doesn't the client have to _opt-in_ to dmabuf buffers for negotiation purposes? Shouldn't the MAP_BUFFERS flag preclude the client from negotiating a dmabuf format successfully? (I'm only asking here because I know you've had more discussions with Wim on this than me)

It's afaik a better than nothing approach and i wouldn't have a problem with it, if it would declare that it only can handle Modifier 0/linear buffer. So please correct me, but afaik there are only two options to map a dmabuf to ram: gbm and mmap. Gbm is the only method, which supports all modifiers, but it is not available on nvidia, that's why most projects try to avoid it for now. So the only way left to map a dmabuf is mmap. There are some projects, like obs, which are able to receive a dmabuf and deal with it internally, so the pipewire plugin (e.g. obs-xdg-portal) only has to handover the buffer.
The only case, where i find it desirable is for testing, to implement a real dmabuf backend later, or if you want to handle dmabufs independing from a specific vendor as fallback. 
There is no opt-in, and im not sure, if we need it as opt in. It would be sufficient, if the consumer (and/or pipewire) would declare their capabilities correctly. This would mean setting the buffer type to allow MemPtr, MemFd and DmaBuf and also set the modifier in the format negotiation to 0. I currently don't know any other usecase of this parameter, so it would not interfere with the other two buffer types. (Example from my obs-xdg-portal pr: [dmabuf via mmap](https://gitlab.gnome.org/feaneron/obs-xdg-portal/-/merge_requests/9/diffs?commit_id=5865b9b79c67f8c21fbae82b044aa0c6de3d64d0), [dmabuf with all modifiers](https://gitlab.gnome.org/feaneron/obs-xdg-portal/-/merge_requests/9/diffs#19512ca962bb78b5cb7e329c0c6a6580c517d80d_507_552))
Most importantly, if a consumer delegates buffer mapping with MAP_BUFFERS, pipewire should set those parameters for the client, since the client doesn't know about pipewires capabilities. This is the one point most important, because it would make it a kind of standard, which could be communicated to the consumers as a requirement.
That's my current thinking. If i have gotten something wrong, please correct me.

> 
> > MemPtr and MemFd are interchangeable, since both are mmaped/mmapable areas of memory in ram to transport buffers.
> 
> Okay, this is the part I was most confused about. So a producer can claim MemFd with the ALLOC_BUFFERS flag, but there's no reason that a consumer couldn't then request MemPtr on its end and let pipewire do the mapping.
> 
If a consumer uses the MAP_BUFFER flag, pipewire will only provide it with already mapped buffers with buffer type MemPtr, regardless of what the producer created.
If a producer uses the MAP_BUFFER flag, pipewire will create/alloc an internal MemFd, map it to the data pointer in the buffer struct and hand that to the producer as a MemPtr. I would guess, that pointer to shared memory are only valid on the same page(??), or the same process and can't be shared between different processes, but MemFd can be shared, since they are somewhat global.
If a producer uses the ALLOC_BUFFERS flag, pipewire won't create/alloc any memory and the producer is responsible to create the correct type of buffer. The current most common ones are MemFd and DmaBuf (don't now what's MemId is for), which are both represented by a shared fd. I can map those to the data pointer at creation time (add_buffer) and they stay valid on the producer side, but i guess thats only because of the same process reason. 

> > Even if we use screencopy dmabuf, we can either create a linear buffer or one with a modifier. For now i only have seen clients either support full modifiers (obs, either wlrobs or obs-xdg-portal (currently merging a patch)), or none (which means mmaping dmabuf). In the none case i would guess, we would get better results, if we do the transition from dmabuf to shm and map the dmabuf with gbm and copy the content to shared memory.
> 
> So if we were to use screencopy-dmabuf exclusively and used gbm to map it linearly, we would still have to call this a dmabuf buffer on the pipewire side, because even in linear, it still isn't kosher to mmap a dma-buf ([without fencing?](https://01.org/linuxgraphics/gfx-docs/drm/driver-api/dma-buf.html#fence-poll-support), I remember daniels mentioning this sowewhere).
> 
No, we would have two choices:
Create a linear dmabuf and hand it to wlroots to copy the screen and after that hand the fd over to pipewire with buffer type DmaBuf, or
Get an arbitrary dmabuf and let wlroots copy the screen into it, allocate for pipewire an MemFd in Ram and use gbm to map it linear in ram to copy it from the dmabuf into the MemFd and hand this MemFd to pipewire. There is no mapping from gpu to cpu we can do for a consumer, since the buffers are transported via file descriptors, since memory pointer are not global. 

> > Tangent about changing wroots protocol on the fly...
> 
> I really think we need to discuss with Wim about some way to allow negotiation to involve a round trip. Right now, a producer cannot change much about what it offers if the client doesn't support renegotiation. It'd be great to be able to signal to pipewire (maybe inside of the param_changed callback) that we would like it to hint to the consumer to try a different pipewire node. We could set up a node for each protocol, start with best possible, and continually fall back until we can negotiate properly. This should be easy for consumers to implement too, it's just a retry, and consumers that don't listen to this hint could still fail like they do today, so it should also be backwards compatible.

Yes, that's  definitely something Wim has the best overview. 
One idea discussed shortly would be to create for every screen one node with one port for MemFd and one for DmaBuf and we would hand over the node id and pipewire would connect the consumer to the correct port of that node, but that's something not implemented yet in the session manager. Another thing to consider is, if it's better to run 2 screencopies (one shm and one dmabuf) at the same time, or just run one screencopy (dmabuf) and copy, if needed, the exported data to the cpu.
I really like the idea of multiple ports, but i haven't found a way yet to notice, if a new client connected to an existing node, since that doesn't trigger a renegotiation for all participators.

danshick:
Thanks. Makes sense.

> One idea discussed shortly would be to create for every screen one node with one port for MemFd and one for DmaBuf and we would hand over the node id and pipewire would connect the consumer to the correct port of that node, but that's something not implemented yet in the session manager. Another thing to consider is, if it's better to run 2 screencopies (one shm and one dmabuf) at the same time, or just run one screencopy (dmabuf) and copy, if needed, the exported data to the cpu.
> I really like the idea of multiple ports, but i haven't found a way yet to notice, if a new client connected to an existing node, since that doesn't trigger a renegotiation for all participators.

Would this mean running multiple screencopies for each output all of the time? Not sure I'm crazy about that for performance reasons and also because it doesn't scale to window captures, thus the hinting suggestion.

emersion:
Let's not mmap dmabufs. This is going to at best be slow, at worse be broken.

danshick:
> Let's not mmap dmabufs. This is going to at best be slow, at worse be broken.

I don't think it is really up to us whether or not we use mmap on a dmabuf, it's a pipewire issue.

Anyway, I found a mention from daniels who said:

> directly mmaping a dmabuf is totally fine, _as long as_ you issue the sync ioctls on it before you begin and end access

Does this resolve the performance issue you mention or does this just make it *not broken*? Other than using mmap or gbm, is there an alternative for dealing with dmabufs?

We are well past the point of being able to fix this here. It sounds to me like contributions need to be made to pipewire so that consumers asking pipewire to map buffers for them will fail negotiation for dmabufs, or pipewire needs to do something to properly handle dmabufs beyond just mmapping them.

emersion:
>Does this resolve the performance issue you mention or does this just make it not broken? 

No, this is just about being able to read a complete frame (as opposed to a half-rendered one).

>Other than using mmap or gbm, is there an alternative for dealing with dmabufs?

If you mean "download the DMA-BUF to the main memory", the last option would be to import it into OpenGL and then call `glReadPixels` (which is pretty much what wlroots does). But it would be a bad idea to do this in xdpw.

I'd much rather spend the time to send patches to Pipewire rather than implement hacks that won't properly work.

danshick:
> I'd much rather spend the time to send patches to Pipewire rather than implement hacks that won't properly work.

Agreed, if we're supposed to be able to supply pipewire with dmabuf FDs, I'd rather fix it there than here.

columbarius:
> I'd much rather spend the time to send patches to Pipewire rather than implement hacks that won't properly work.

Agreed. My intend was to discuss, what we want to patch in pipewire. I see currently two problems:
1. Buffertypehandling/declaration (especially dmabuf)
1.1 How to handle dmabuf
1.2 How to detect the correct buffertype for the client
2. The ability to offer multiple streams with different formats, or find a suitable format for all consumers
2.1 The ability to find a working combination from the set of tuples containing the buffertype with its videoformat.
2.2 Connecting a new client to an existing stream, with matching format and buffertype, or changing the stream to a combination, both clients understand

About those points:
1. Wim said that the MAP_BUFFERS flag is just a fallback for simple cases and in general all clients (producer and consumer) should do all mapping themself, so it's outside the pipewire deamon ([firefox](https://src.fedoraproject.org/rpms/firefox/blob/master/f/firefox-pipewire-0-3.patch#_459), [gstreamersink](https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/gst/gstpipewiresink.c#L525), [gstreamersrc](https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/gst/gstpipewiresrc.c#L690) does the mapping on their own, [obs-xdg-portal](https://gitlab.gnome.org/feaneron/obs-xdg-portal/-/merge_requests/9/diffs#19512ca962bb78b5cb7e329c0c6a6580c517d80d_625_674) will be patched to do the same, I don't know what chromium does).
1.1 About dmabuf: We could easily create a patch to map dmabufs inside pipewire using gbm, but Wim said, that he would like to avoid it as a dependency. Since the mapping should in general be outside of pipewire, I don't see that as the real answer to our problem. What I don't know is, how easy/difficult it is to compile a program, which works on mesa and nvidia at the same time, choosing the respective codepath, or using dmabufs only in one case. If it's easy i can see clients using dmabuf.
1.2 My main problem at the moment is, that no client specifies, what SPA_PARAM_BUFFERS_dataType it can use. I think this needs to be patched. We can detect, if a client does not set this parameter and we could just end the pipewire stream, but this would break compatibility with (almost) all consumers. And there is the point, where we have to patch pipewire imho. If a consumer sets the MAP_BUFFERS flag, pipewire has to overwrite the BUFFERS_dataType parameter with a bitmap suggesting MemPtr, MemFd (and DmaBuf if also the FORMAT_modifier is set to 0). That would let the producer select a valid buffer type. I don't have a problem, if a client wants to map a dmabuf with mmap on his side, as long, as it announces DmaBuf only with modifier 0. We could later decide, that if a consumer only supports modifier 0, that we don't use dmabuf an another announced buffer type instead, but that's debatable later. With this suggestion, we would have to message the maintainers of (, or patch) firefox/chromium/webrtc and the gstreamerplugins (pipewire) and hope, that future clients will adapt this, if they see it documented. I looked into pipewire how this could be done, but i had not the time to really understand it.
2. This probably requires more work in pipewire and I don't have a solution at hand, that I can fully propose, since I'm lacking experience. 
2.1 This probably requires some round trip, but I'm not convinced, if it's worth all the complexity.
2.2 My preferred option for this would be, to create one node with two ports, one for dmabuf and one for memfd. Then either we or pipewire could look at the client and either try to connect it to dmabuf or shm depending on it's capabilities. If a client connects to the dmabuf, we could just hand over the buffer, and one connects to the shm port, we can map the dmabuf and copy it for the client to ram. Since we only can offer the video format, which we receive from wlroots and we don't want to transcode the buffer data, it's pointless to offer ports with another difference then the buffer type used for transport. I don't know if ports on the same node are synced or not, or how to handle them.

So in short:
My suggestion would be to first, fix the BUFFERS_dataType announcement in pipewire and the main consumers (webrtc and gstreamer) and then discuss on how to solve the negotiation issue.

columbarius:
> Would this mean running multiple screencopies for each output all of the time? Not sure I'm crazy about that for performance reasons and also because it doesn't scale to window captures, thus the hinting suggestion.

Your hinting suggestion? I would try to run as few screencopies as possible and would probably try to only run the dmabuf screencopy and if a client only supports MemFd, I would rather just serve the dmabuf to the previous clients and mount and download the same buffer for this client instead of changing starting a new screencopy.

columbarius:
> But it would be a bad idea to do this in xdpw.

Agreed, the thing in question is, if there is a suitable way to implement download of dmabuf in pipewire in the MAP_BUFFERS case, which means that it is no huge dependency, is available at every target platform and doesn't create a huge amount of code just to implement it. And i think there is currently no sollution, or is it?

columbarius:
Because of #65 I looked into chromium screensharing, which didn't work for me. Still doesn't work with this patch. Reason seams to be https://github.com/emersion/xdg-desktop-portal-wlr/pull/48/files#diff-25e29cec459c41df3c682e4e3ed522d43aeda1c932c4b8e96d3910925e50fbf4R361 which stops the negotiation.

columbarius:
Some discussion i had with danshick

```
19:36:40 columbarius | I probably have to redo my whole dmabuf work since we have to use shm and dmabuf copy in tandem enabling and disabling them if needed.
19:37:57    danshick | There's got to be something better we can do in pipewire (or even xdg-desktop-portal) to improve that wha
19:38:02    danshick | *whole situation
19:38:53 columbarius | I thought about creating a inbetween struct screencast_session or something, which gets created for each combination of output and cursor_mode and inherits the pipewire node.
19:40:02    danshick | What would that struct hold?
19:40:41 columbarius | Then it has two screencast_intances, one for shm and one for dmabuf and two corresponding pipewire ports. We can declare format and buffer type at the start and listen if a client connects to one
                     | port and then start the corresponding screencast instance
19:42:22 columbarius | Via dbus we could return the id of the node and pipewire has to connect the consumer to the correct port. I don't know, if it's able to do so at the moment, but i think that would be the best
                     | solution
19:45:40 columbarius | the screencast_session would hold the montor information, the pipewire information and all the two screencast_instances can share. It's more like moving the wlr_screencopy part into it's own struct
                     | and adding the port information
19:47:13    danshick | Ah, gotcha.
19:47:28 columbarius | I havn't tried to do it yet, but if it works, it would avoid the whole negotiation issue. We would offer all formats, we support and if a client is not compatible with any port, there is not much we
                     | can do   
19:47:41    danshick | I definitely don't know enough about pipewire nodes vs. ports to say what can be done with that.
19:47:55    danshick | Makes sense.
19:55:56 columbarius | I also don't know enough, but i looked into the api for nodes and you can create arbitrary nodes, so i think it's worth to look into it.
19:56:25 columbarius | I think, it will also help to formulate what kind of patches we need from / to add to pipewire

```

columbarius:
This is obsolete by #152

PR #53, opened 2020-09-11
Update to wlr-screencopy-unstable-v1 version 3

columbarius:
This PR updates the wlr-screencopy-unstable-v1 to version 3 with only implementing an shm buffer. It also provides an fallback to version 2, if the compositor only supports that. It also lists all available protocols with the used version at startup, when in loglevel debug.

### DIFF
--- a/src/screencast/wlr_screencast.c
+++ b/src/screencast/wlr_screencast.c
@@ -325,10 +325,12 @@ static void wlr_registry_handle_add(void *data, struct wl_registry *reg,
 		uint32_t id, const char *interface, uint32_t ver) {
 	struct xdpw_screencast_context *ctx = data;
 
+	logprint(DEBUG, "wlroots: interface to register %s  (Version: %u)",interface, ver);
 	if (!strcmp(interface, wl_output_interface.name)) {
 		struct xdpw_wlr_output *output = malloc(sizeof(*output));
 
 		output->id = id;
+		logprint(DEBUG, "wlroots: |-- registered to interface %s (Version %u)", interface, WL_OUTPUT_VERSION);
 		output->output = wl_registry_bind(reg, id, &wl_output_interface, WL_OUTPUT_VERSION);
 
 		wl_output_add_listener(output->output, &wlr_output_listener, output);
@@ -336,15 +338,19 @@ static void wlr_registry_handle_add(void *data, struct wl_registry *reg,
 	}
 
 	if (!strcmp(interface, zwlr_screencopy_manager_v1_interface.name)) {
+		uint32_t version = SC_MANAGER_VERSION;
+		logprint(DEBUG, "wlroots: |-- registered to interface %s (Version %u)", interface, version);
 		ctx->screencopy_manager = wl_registry_bind(
-			reg, id, &zwlr_screencopy_manager_v1_interface, SC_MANAGER_VERSION);
+			reg, id, &zwlr_screencopy_manager_v1_interface, version);
 	}
 
 	if (strcmp(interface, wl_shm_interface.name) == 0) {
+		logprint(DEBUG, "wlroots: |-- registered to interface %s (Version %u)", interface, WL_SHM_VERSION);
 		ctx->shm = wl_registry_bind(reg, id, &wl_shm_interface, WL_SHM_VERSION);
 	}
 
 	if (strcmp(interface, zxdg_output_manager_v1_interface.name) == 0) {
+		logprint(DEBUG, "wlroots: |-- registered to interface %s (Version %u)", interface, XDG_OUTPUT_MANAGER_VERSION);
 		ctx->xdg_output_manager =
 			wl_registry_bind(reg, id, &zxdg_output_manager_v1_interface, XDG_OUTPUT_MANAGER_VERSION);
 	}
--- a/src/screencast/wlr_screencast.c
+++ b/src/screencast/wlr_screencast.c
@@ -329,7 +329,7 @@ static void wlr_registry_handle_add(void *data, struct wl_registry *reg,
 		struct xdpw_wlr_output *output = malloc(sizeof(*output));
 
 		output->id = id;
-		output->output = wl_registry_bind(reg, id, &wl_output_interface, 1);
+		output->output = wl_registry_bind(reg, id, &wl_output_interface, WL_OUTPUT_VERSION);
 
 		wl_output_add_listener(output->output, &wlr_output_listener, output);
 		wl_list_insert(&ctx->output_list, &output->link);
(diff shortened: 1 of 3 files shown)


src/screencast/wlr_screencast.c
```diff
@@ -118,6 +118,23 @@ static void wlr_frame_buffer_chparam(struct xdpw_screencast_instance *cast,
 	wlr_frame_buffer_destroy(cast);
 }
 
+static void wlr_frame_linux_dmabuf(void *data,
+		struct zwlr_screencopy_frame_v1 *frame,
+		uint32_t format, uint32_t width, uint32_t height) {
+	//struct xdpw_screencast_instance *cast = data;
```
danshick, line 124: No need to leave this comment.

src/screencast/wlr_screencast.c
```diff
@@ -320,28 +340,39 @@ static void wlr_registry_handle_add(void *data, struct wl_registry *reg,
 		uint32_t id, const char *interface, uint32_t ver) {
 	struct xdpw_screencast_context *ctx = data;
 
+	uint32_t version = 0;
```
danshick, line 343: I like the idea of getting rid of these magic numbers, but I think I'd rather declare additional constants (like SC_MANAGER_VERSION) for each of these interfaces, and if you need to do some comparing/checking like you're doing on L#358, just keep the variables scoped to the if block.

danshick (changes requested):

src/screencast/wlr_screencast.c
columbarius: Ok. Good idea to change those to macros. Changed them.

src/screencast/wlr_screencast.c
columbarius: Ok, removed that. Will add that line again, when i actual use this function. Had commented that out, because of that unused variable compiler warning.

danshick: Thanks, that's better. I think at one point I had even written a noop function just for an unused event to avoid the warning too.

src/screencast/wlr_screencast.c
```diff
@@ -325,28 +344,35 @@ static void wlr_registry_handle_add(void *data, struct wl_registry *reg,
...
-		output->output = wl_registry_bind(reg, id, &wl_output_interface, 1);
+		logprint(DEBUG, "wlroots: |-- registered to interface %s (Version %u)", interface, WL_OUTPUT_VERSION);
+		output->output = wl_registry_bind(reg, id, &wl_output_interface, WL_OUTPUT_VERSION);
 
 		wl_output_add_listener(output->output, &wlr_output_listener, output);
 		wl_list_insert(&ctx->output_list, &output->link);
 	}
 
 	if (!strcmp(interface, zwlr_screencopy_manager_v1_interface.name)) {
+		uint32_t version = SC_MANAGER_VERSION < ver ? SC_MANAGER_VERSION : ver;
```
emersion, line 360: We should check `version >=2`, because we don't support v1.

columbarius: ok, added a macro for it

src/screencast/wlr_screencast.c
```diff
@@ -325,28 +344,36 @@ static void wlr_registry_handle_add(void *data, struct wl_registry *reg,
...
+		logprint(DEBUG, "wlroots: |-- registered to interface %s (Version %u)", interface, WL_OUTPUT_VERSION);
+		output->output = wl_registry_bind(reg, id, &wl_output_interface, WL_OUTPUT_VERSION);
 
 		wl_output_add_listener(output->output, &wlr_output_listener, output);
 		wl_list_insert(&ctx->output_list, &output->link);
 	}
 
 	if (!strcmp(interface, zwlr_screencopy_manager_v1_interface.name)) {
+		uint32_t version = SC_MANAGER_VERSION < ver ? SC_MANAGER_VERSION :
+			ver < SC_MANAGER_VERSION_MIN ? SC_MANAGER_VERSION_MIN : ver;
```
danshick, line 361: I am guilty of trying to get away with one-liners like this too, but this is way more complicated to read than:
```c
uint32_t version = ver;
if (SC_MANAGER_VERSION < ver) {
  version = SC_MANAGER_VERSION;
}
if (SC_MANAGER_VERSION_MIN > ver) {
  version = SC_MANAGER_VERSION_MIN;
}
```
Also, I think we need to bail out if the version offered is less than the minimum supported. Similarly, we haven't been checking for the other interfaces if the version offered is compatible with the version(s) we support. @emersion, are these interfaces necessarily backwards compatible?

danshick (changes requested):

src/screencast/wlr_screencast.c
emersion: >Similarly, we haven't been checking for the other interfaces if the version offered is compatible with the version(s) we support.

If we request v3 and the compositor only supports v2, I think libwayland will error out. Would need to double-check.

> @emersion, are these interfaces necessarily backwards compatible?

Yes, but not in a way you'd necessarily expect.

- If the compositor supports v3, it *must* support v1 and v2 as well
- If the client requests v3, the compositor may behave differently than if the client requests v1 or v2 (not necessarily in a backwards-compatible way)

columbarius: > I am guilty of trying to get away with one-liners like this too, but this is way more complicated to read than:

Sure, changed it.

danshick: Looks good, but given emersion's comment, might be a good idea to look into handling errors arising from `bind`ing a version that is too high (or I suppose we could just call exit() or similar if the compositor doesn't offer the minimum required version).

src/screencast/wlr_screencast.c
```diff
@@ -325,28 +344,40 @@ static void wlr_registry_handle_add(void *data, struct wl_registry *reg,
...
 	}
 
 	if (!strcmp(interface, zwlr_screencopy_manager_v1_interface.name)) {
+		uint32_t version = ver;
+		if (SC_MANAGER_VERSION < ver) {
+			version = SC_MANAGER_VERSION;
+		} else if (ver < SC_MANAGER_VERSION_MIN) {
+			version = SC_MANAGER_VERSION_MIN;
+		}
+		ctx->state->screencast_version = version;
```
emersion, line 366: No need to store this information. We can get the version back via `zwlr_screencopy_manager_v1_get_version`.

columbarius: We need this information later for every frame (https://github.com/columbarius/xdg-desktop-portal-wlr/blob/screencopy-interface-v3/src/screencast/wlr_screencast.c#L169), so would prefer to safe this and not request it every time. It also doesn't change between different screencasts.

emersion: `zwlr_screencopy_manager_v1_get_version` is not a request, it's just a getter.

columbarius: Ahh thanks. Changed it.

emersion (approved):
Thanks!

columbarius:
Thanks for merging!

PR #55, opened 2020-09-18
Add pipewire buffers_dataType parameter in pipwire_screencast.c

columbarius:
Pipewire has a bitmap (dataType buffer parameter) to compare the available buffer types since 0.3.2.


emersion:
Can you bump the minimum required Pipewire version in `meson.build` as well?

emersion (approved):
LGTM, thanks!

PR #61, opened 2020-10-23
Improve pipewire buffertype handling

columbarius:
Updated and rebased and probably rewritten large parts of this MR to work with #141. So there are some old and probably some wrong things in the discussions.

-----------------------------------------------------------------------------------------------------------

I looked into buffertype handling in pipewire and talked to @wtay at #pipewire about it.
This draft is to show the current state (of at least my understanding).

-  SPA_PARAM_BUFFERS_dataType
SPA_PARAM_BUFFERS_dataType should be set by a consumer not by a producer as a bitmap to indicate the supported buffer types.
- {add,remove}_buffer callbacks
those callbacks are called, if pipewire adds or removes a new pw_buffer. 
If no alloc flag is used to connect to pipewire the pw_buffer is read-only and buffer->datas[].type contains the supported buffer type bitmap from the client or SPA_DATA_Invalid if the client has not specified anything.
If PW_STREAM_FLAG_ALLOC_BUFFERS is used it's to producers duty to read the bitmap and set at least it's used buffer type and maybe other buffer parameters like size, stride, etc (not sure about that).
- PW_STREAM_FLAG_MAP_BUFFERS
This flag tells pipewire to mmap in case of dmabuf or memfd the fd to the data pointer. Since we are doing our own fd handling i guess, we don't need that (Will clarify if it does anything in case of SPA_DATA_MemPtr).

### DIFF
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -5,6 +5,8 @@
 #include <spa/param/props.h>
 #include <spa/param/format-utils.h>
 #include <spa/param/video/format-utils.h>
+#include <sys/mman.h>
+#include <unistd.h>
 
 #include "wlr_screencast.h"
 #include "xdpw.h"
@@ -145,7 +147,7 @@ static void pwr_handle_stream_param_changed(void *data, uint32_t id,
 		SPA_PARAM_BUFFERS_size,    SPA_POD_Int(cast->simple_frame.size),
 		SPA_PARAM_BUFFERS_stride,  SPA_POD_Int(cast->simple_frame.stride),
 		SPA_PARAM_BUFFERS_align,   SPA_POD_Int(XDPW_PWR_ALIGN),
-		SPA_PARAM_BUFFERS_dataType,SPA_POD_CHOICE_FLAGS_Int(1<<SPA_DATA_MemPtr));
+		SPA_PARAM_BUFFERS_dataType,SPA_POD_CHOICE_FLAGS_Int(1<<SPA_DATA_MemFd));
 
 	params[1] = spa_pod_builder_add_object(&b,
 		SPA_TYPE_OBJECT_ParamMeta, SPA_PARAM_Meta,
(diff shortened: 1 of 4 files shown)


danshick:
Is there any proper documentation of those flag behaviors for pipewire? Other than reading from the examples/tutorials, I've never found a solid explanation. Anyway, I'd like to comment why we're using certain flags/metadata ao we can keep track of what we're intending. Something like [this](https://webrtc.googlesource.com/src/+/refs/heads/master/modules/desktop_capture/linux/base_capturer_pipewire.cc#343).

columbarius:
I can look up the discussions at #pipewire, but afaik nothing official.
That's a good idea with those comments. 

I just opened this Draft to discuss this with the ability to point at explicit code parts.
Reference add_buffer
https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/examples/video-src-alloc.c#L224
Reference Datatype bitmask
https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/47ab07589593a3b13b4d2d79aaa4eedb242f1b86

danshick:
> I can look up the discussions at #pipewire, but afaik nothing official.

My bouncer follows #pipewire, I'll go back and take a look.

src/screencast/pipewire_screencast.c
```diff
@@ -214,8 +214,7 @@ void xdpw_pwr_stream_init(struct xdpw_screencast_instance *cast) {
 	pw_stream_connect(cast->stream,
 		PW_DIRECTION_OUTPUT,
 		PW_ID_ANY,
-		(PW_STREAM_FLAG_DRIVER |
-			PW_STREAM_FLAG_MAP_BUFFERS),
+		(PW_STREAM_FLAG_DRIVER),
```
emersion, line 217: Extra parens here

src/screencast/pipewire_screencast.c
```diff
@@ -129,10 +129,29 @@ static void pwr_handle_stream_param_changed(void *data, uint32_t id,
 	pw_stream_update_params(stream, params, 2);
 }
 
+static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
+	struct spa_data *d;
+
+	logprint(TRACE, "pipewire: add buffer event handle");
+
+	d = buffer->buffer->datas;
+	logprint(TRACE, "pipewire: buffer type %08x",d[0].type);
```
emersion, line 139: Style: missing space after comma

src/screencast/pipewire_screencast.c
```diff
@@ -129,10 +129,29 @@ static void pwr_handle_stream_param_changed(void *data, uint32_t id,
...
+	if (d[0].type == SPA_DATA_Invalid || ((d[0].type & (1<<SPA_DATA_MemPtr)) == 0)) {
+		logprint(ERROR, "pipewire: wrong buffer type %08x",d[0].type);
```
emersion, line 141: Style: missing space after comma

src/screencast/pipewire_screencast.c
```diff
@@ -129,10 +129,29 @@ static void pwr_handle_stream_param_changed(void *data, uint32_t id,
...
+	}
+}
+
+static void pwr_handle_stream_remove_buffer(void *data, struct pw_buffer *buffer) {
+
```
emersion, line 147: Style: extra newline here

emersion:
I think "Add {add,remove}_buffer handler to pipewire_screencast.c" and "Fix typecheck" can be squashed together?

It would be nice to add short explanations of the reason of the changes in each commit.

columbarius:
@emersion thanks for the review. 
This draft is very early and i would rebase #48 first onto it, to test, how the add_buffer functions could be used best to check the negotiated parameters. Since we need to wait for pipewire to add the discussed defaults for dmabufs, for this to have a real advantage to the current state. 
So I would keep small commits for now to have discrete markers for different features, which i would squash if it's ready.

emersion:
No problem, let me know when you need a new review.

columbarius:
@emersion @danshick 
I currently can't figure out a good way to create some hierarchy on how to select which buffer-type and respectively backend (screencopy dmabuf/export dmabuf/screencopy shm) should be used. The current way is:
* Choose a backend and request a frame for dimention and format information for this backend.
* Offer this information to pipewire
* pipewire negotiates the format
* pipewire calls the param_changed handle, we can compare the result and set the buffer information
* pipewire negotiates the buffer
* pipewire calls add_buffer, we can inspect the negotiated buffer type and can either use one available buffer type of our backend, or we have to change our backend and do the whole negotiation again.

What about opening one port with its format and buffer type for every backend and let pipewire to the negotiation and selection of the correct port and we will just start copying frames if the pipewire stream is used.
Would that be possible without more overhead?

src/screencast/pipewire_screencast.c
```diff
@@ -163,10 +157,75 @@ static void pwr_handle_stream_param_changed(void *data, uint32_t id,
...
+		d[0].mapoffset = 0;
+		d[0].chunk->size = cast->simple_frame.size;
+		d[0].chunk->stride = cast->simple_frame.stride;
+		d[0].chunk->offset = 0;
+		d[0].flags = 0;
+		d[0].fd = anonymous_shm_open();
+
+		if (d[0].fd == -1) {
+			logprint(ERROR, "pipewire: unable to create anonymous filedescriptor");
+			return;
```
emersion, line 190: Should we set `cast->err` for this error condition? (And the next 2 as well)

src/screencast/pipewire_screencast.c
```diff
@@ -163,10 +157,75 @@ static void pwr_handle_stream_param_changed(void *data, uint32_t id,
...
+		}
+
+		if (ftruncate(d[0].fd, d[0].maxsize) < 0) {
+			logprint(ERROR, "pipewire: unable to truncate filedescriptor");
+			close(d[0].fd);
+			d[0].fd = -1;
+			return;
+		}
+
+		// mmap buffer, so we can use the data_ptr int on_process
```
emersion, line 200: Typo: "in"

emersion:
LGTM apart from the error handling.

src/screencast/pipewire_screencast.c
columbarius: yeah sounds good. Don't know if we should check for that before registering the wlroots callback or somewhere else.

src/screencast/pipewire_screencast.c
columbarius: thx

src/screencast/pipewire_screencast.c
emersion: Hm, yeah we should probably check this as soon as we get the broken buffer from somewhere else, which is probably when sending the screencopy request to the compositor. It's unfortunate that `add_buffer` can't return an error.

emersion (approved):
LGTM, thanks!

columbarius:
Thanks for merging!

PR #62, opened 2020-10-27
Add Wayfire to portal UseIn list, as it is also supported

kode54:
Quick and dirty change. I intend to add some basic setup to the readme as well, if nobody else is doing that. Mainly just a hint to use the systemctl or dbus to systemd environment setup in autostart, or else the auto loaded portal won't actually work.

### DIFF
--- a/wlr.portal
+++ b/wlr.portal
@@ -1,4 +1,4 @@
 [portal]
 DBusName=org.freedesktop.impl.portal.desktop.wlr
 Interfaces=org.freedesktop.impl.portal.Screenshot;org.freedesktop.impl.portal.ScreenCast;
-UseIn=sway
+UseIn=sway;Wayfire


danshick:
No problem extending the UseIn list. Without looking, I think the semicolon syntax is correct.

As for the readme changes, I'm planning on overhauling the docs a bit shortly. Namely adding man pages. I've also been working on a wiki page explaining the quirks of environment variables and systemd/dbus activated services. I'd appreciate your feedback as it is brand new. Let me know if it covers the setup instructions you were planning to add. Once more people have reviewed it, I plan on linking to it from the readme prominently.

Thanks for the PR!

kode54:
Well, my usage involves a full environment pass, but that's a bit drastic, as it probably just needs one more variable than I was passing before. Possibly the one which identifies the DE in use. These steps are employed by the autostart for Wayfire, using a 0 prefix on the entry name, so they run first.

Arch's sway and AUR sway-git packages also include a /etc/sway/config.d insert to do this automatically, but only for a limited number of variables, and I think I need to file a bug report to get the same variable added to the upstream package's config snippet, so it will work under sway by default.

I'm also going to file feature request for something I may end up implementing myself.

emersion (approved):
LGTM as well, thanks!

PR #64, opened 2020-11-06
Fix cursor_mode checks in screencast.c

columbarius:
see https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/src/core/main.c#L100, or https://github.com/flatpak/xdg-desktop-portal-gtk/blob/master/src/screencast.c#L542


emersion (approved):
Nice catch, thanks!

Issue #66, opened 2020-11-11
Laggy mouse when screen sharing on 4K resolution


zsolt-donca:
When starting screen sharing from a browser with WebRTC, the mouse cursor gains a significant lag on the display that is being shared. My other screen doesn't seem to be affected, the mouse cursor works just fine over there, even while the screen sharing is being active. Stopping the screen sharing makes the lag disappear. 

I am using two 4K displays. This is probably related to the 4K resolution, as if I switch the resolution of one of the displays to 1920x1080 and share that display, the mouse cursor is moving perfectly smooth.

This lag also feels when typing on the keyboard, so I think it's a general input lag problem, and not just specific to the mouse.

I made a screenshot of htop with the screen sharing being active on the other screen. It It shows `sway`, `firefox`, `xdg-desktop-portal-wlr` being the biggest CPU users:

![screenshot_2020-11-09_19-30-31_735861790](https://user-images.githubusercontent.com/6824167/98577577-d5277780-22c4-11eb-81c8-ee4f00207dd8.png)

For comparison, the mouse is not laggy when on GNOME in Wayland mode (tested with a Fedora 33 live image), nor in Xorg mode. I wouldn't mind the relatively high CPU usage if there wasn't this input lag.

Another related question: why can I share only one of my displays? When the browser prompts, there is only a single display being available. It happens to be my "main" display, so actually it's fine for me, I'm just curious.

I also mentioned this in #65 .

emersion:
Screen sharing disables hw cursors.

https://github.com/swaywm/wlroots/issues/2091 may help wrt. Sway CPU usage.

zsolt-donca:
I see, I guess making the operations asynchronous will help also with the input lag issue. I wonder in what stage that ticket and the associated MR is, but I will ask in there.

Just for the record, GNOME's Wayland compositor (Mutter, I think) doesn't have this issue. I guess somebody could take a look at how this problem is solved over there, maybe the same solution applies here as well.

What about the fact that there is only a single monitor available for screen share? Is that a known limitation? If not, I would happily open another ticket with that issue also.

emersion:
Maybe GNOME never disables hw cursors when screensharing, but I don't think it should make that big of a difference...

solarkraft:
I have the same issue and while I don't understand much about the technology behind it I've made some observations that may be note worthy:

- The cursor on the other screen is only smooth until something on the shared screen changes (for the amount of time there is change). 

[...]

PR #74, opened 2020-12-25
Control how many frames are captured per second

zsolt-donca:
Addresses #66 by introducing an optional command-line argument to limit the number of frames captured per second, as the mouse lag issue becomes less noticeable if the frame rate is smaller. The argument represents the targeted frames-per-second value (e.g. pass `-f 10` for 10 FPS).

The implementation took inspiration from https://github.com/any1/wayvnc, but no code was copied other than the file `include/time-util.h`.

The actual FPS value can be lower than the targeted one because xdg-desktop-portal-wlr uses the `zwlr_screencopy_frame_v1_copy_with_damage` function (emphasis on `with_damage`), which waits until the screen changes.

The implementation differs from that of wayvnc that it does not try to "smooth" the capture rate over time, and it always uses the capture time of the last frame as its best estimate for how much time to sleep for current frame, and in order to reach the targeted FPS value. Let me know if there is interest in implementing this - for one, I am not sure if its worth it, and especially on possible implications with regards to xdp only capturing the frame on damage, which can make the measured "time to capture" value vary significantly, and could impact the smoothing negatively (which expects the values to be predictable).

Another difference when compared to wayvnc's implementation is that this code simply suspends the current thread with `usleep`, while wayvnc uses a more sophisticated timer-based solution based on the author's own library https://github.com/any1/aml.

With regards to testing, it seems to (approximately) produce the given targeted FPS value for me, at least as long as I keep changing the screen contents (by moving the mouse, for example). For example, when running with `xdg-desktop-portal-wlr -l DEBUG -f 10`, it produced these logs for me:
```
2020/12/25 18:55:24 [DEBUG] - fps_limit: average FPS in the last 5 seconds: 10.04
2020/12/25 18:55:29 [DEBUG] - fps_limit: average FPS in the last 5 seconds: 9.97
2020/12/25 18:55:34 [DEBUG] - fps_limit: average FPS in the last 5 seconds: 10.07
2020/12/25 18:55:39 [DEBUG] - fps_limit: average FPS in the last 5 seconds: 9.96
2020/12/25 18:55:45 [DEBUG] - fps_limit: average FPS in the last 5 seconds: 10.02
```

Another example with `-f 20`:
```
2020/12/25 19:01:30 [DEBUG] - fps_limit: average FPS in the last 5 seconds: 19.57
2020/12/25 19:01:35 [DEBUG] - fps_limit: average FPS in the last 5 seconds: 19.94
2020/12/25 19:01:40 [DEBUG] - fps_limit: average FPS in the last 5 seconds: 19.93
2020/12/25 19:01:45 [DEBUG] - fps_limit: average FPS in the last 5 seconds: 19.92
```

### DIFF
--- a/include/fps_limit.h
+++ b/include/fps_limit.h
@@ -1,0 +1,18 @@
+#ifndef FPS_LIMIT_H
+#define FPS_LIMIT_H
+
+#include <stdint.h>
+#include <time.h>
+
+struct fps_limit_state {
+	struct timespec frame_last_time;
+	
+	struct timespec fps_last_time;
+	uint64_t fps_frame_count;
+};
+
+void fps_limit_measure_start(struct fps_limit_state *state, double max_fps);
+
+uint64_t fps_limit_measure_end(struct fps_limit_state *state, double max_fps);
+
+#endif
--- a/include/screencast_common.h
+++ b/include/screencast_common.h
@@ -5,6 +5,8 @@
 #include <spa/param/video/format-utils.h>
 #include <wayland-client-protocol.h>
 
+#include "fps_limit.h"
+
 // this seems to be right based on
 // https://github.com/flatpak/xdg-desktop-portal/blob/309a1fc0cf2fb32cceb91dbc666d20cf0a3202c2/src/screen-cast.c#L955
 #define XDP_CAST_PROTO_VER 2
@@ -54,7 +56,7 @@ struct xdpw_screencast_context {
 	struct wl_list output_list;
 	struct wl_registry *registry;
 	struct zwlr_screencopy_manager_v1 *screencopy_manager;
-	struct zxdg_output_manager_v1* xdg_output_manager;
+	struct zxdg_output_manager_v1 *xdg_output_manager;
 	struct wl_shm *shm;
 
 	// sessions
@@ -88,6 +90,9 @@ struct xdpw_screencast_instance {
 	bool with_cursor;
 	int err;
 	bool quit;
+
+	// fps limit
+	struct fps_limit_state fps_limit;
 };
 
 struct xdpw_wlr_output {
--- a/include/timespec_util.h
+++ b/include/timespec_util.h
@@ -1,0 +1,18 @@
+#ifndef TIMESPEC_UTIL_H
+#define TIMESPEC_UTIL_H
+
+#include <time.h>
+#include <stdbool.h>
+#include <stdint.h>
+
+#define TIMESPEC_NSEC_PER_SEC 1000000000L
+
+void timespec_add(struct timespec *t, int64_t delta_ns);
+
+bool timespec_less(struct timespec *t1, struct timespec *t2);
+
+bool timespec_is_zero(struct timespec *t);
+
+int64_t timespec_diff_ns(struct timespec *t1, struct timespec *t2);
+
+#endif
--- a/meson.build
+++ b/meson.build
@@ -28,6 +28,12 @@ wayland_client = dependency('wayland-client')
 wayland_protos = dependency('wayland-protocols', version: '>=1.14')
 iniparser = cc.find_library('iniparser', dirs: [join_paths(get_option('prefix'),get_option('libdir'))])
 
+epoll = dependency('', required: false)
+if (not cc.has_function('timerfd_create', prefix: '#include <sys/timerfd.h>') or
+    not cc.has_function('signalfd', prefix: '#include <sys/signalfd.h>'))
+	epoll = dependency('epoll-shim')
+endif
+
 if get_option('sd-bus-provider') == 'auto'
 	assert(get_option('auto_features').auto(), 'sd-bus-provider must not be set to auto since auto_features != auto')
 	sdbus = dependency('libsystemd',
@@ -63,11 +69,14 @@ executable(
 		'src/core/config.c',
 		'src/core/request.c',
 		'src/core/session.c',
+		'src/core/timer.c',
+		'src/core/timespec_util.c',
 		'src/screenshot/screenshot.c',
 		'src/screencast/screencast.c',
 		'src/screencast/screencast_common.c',
 		'src/screencast/wlr_screencast.c',
 		'src/screencast/pipewire_screencast.c',
+		'src/screencast/fps_limit.c'
 	]),
 	dependencies: [
 		wayland_client,
@@ -76,6 +85,7 @@ executable(
 		pipewire,
 		rt,
 		iniparser,
+		epoll,
 	],
 	include_directories: [inc],
 	install: true,
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -1,10 +1,13 @@
 #include <errno.h>
 #include <stdlib.h>
 #include <stdio.h>
+#include <sys/timerfd.h>
 #include <getopt.h>
 #include <poll.h>
 #include <pipewire/pipewire.h>
 #include <spa/utils/result.h>
+#include <unistd.h>
+
 #include "xdpw.h"
 #include "logger.h"
 
@@ -12,12 +15,13 @@ enum event_loop_fd {
 	EVENT_LOOP_DBUS,
 	EVENT_LOOP_WAYLAND,
 	EVENT_LOOP_PIPEWIRE,
+	EVENT_LOOP_TIMER,
 };
 
 static const char service_name[] = "org.freedesktop.impl.portal.desktop.wlr";
 
-static int xdpw_usage(FILE* stream, int rc) {
-	static const char* usage =
+static int xdpw_usage(FILE *stream, int rc) {
+	static const char *usage =
 		"Usage: xdg-desktop-portal-wlr [options]\n"
 		"\n"
 		"    -l, --loglevel=<loglevel>        Select log level (default is ERROR).\n"
@@ -27,6 +31,7 @@ static int xdpw_usage(FILE* stream, int rc) {
 		"    -c, --config=<config file>	      Select config file.\n"
 		"                                     (default is $XDG_CONFIG_HOME/xdg-desktop-portal-wlr/config)\n"
 		"    -r, --replace                    Replace a running instance.\n"
+		"    -f, --max-fps=<fps>              Set the FPS limit (default 0, no limit).\n"
 		"    -h, --help                       Get help (this text).\n"
 		"\n";
 
@@ -46,11 +51,12 @@ int main(int argc, char *argv[]) {
 	enum LOGLEVEL loglevel = DEFAULT_LOGLEVEL;
 	bool replace = false;
 
-	static const char* shortopts = "l:o:c:rh";
+	static const char *shortopts = "l:o:c:f:rh";
 	static const struct option longopts[] = {
 		{ "loglevel", required_argument, NULL, 'l' },
 		{ "output", required_argument, NULL, 'o' },
 		{ "config", required_argument, NULL, 'c' },
+		{ "max-fps", required_argument, NULL, 'f' },
 		{ "replace", no_argument, NULL, 'r' },
 		{ "help", no_argument, NULL, 'h' },
 		{ NULL, 0, NULL, 0 }
@@ -58,8 +64,9 @@ int main(int argc, char *argv[]) {
 
 	while (1) {
 		int c = getopt_long(argc, argv, shortopts, longopts, NULL);
-		if (c < 0)
+		if (c < 0) {
 			break;
+		}
 
 		switch (c) {
 		case 'l':
@@ -73,6 +80,9 @@ int main(int argc, char *argv[]) {
 			break;
 		case 'r':
 			replace = true;
+			break;
+		case 'f':
+			config.screencast_conf.max_fps = atof(optarg);
 			break;
 		case 'h':
 			return xdpw_usage(stdout, EXIT_SUCCESS);
@@ -166,6 +176,8 @@ int main(int argc, char *argv[]) {
 		goto error;
 	}
 
+	wl_list_init(&state.timers);
+
 	struct pollfd pollfds[] = {
 		[EVENT_LOOP_DBUS] = {
 			.fd = sd_bus_get_fd(state.bus),
@@ -179,7 +191,13 @@ int main(int argc, char *argv[]) {
 			.fd = pw_loop_get_fd(state.pw_loop),
 			.events = POLLIN,
 		},
+		[EVENT_LOOP_TIMER] = {
+			.fd = timerfd_create(CLOCK_MONOTONIC, TFD_CLOEXEC),
+			.events = POLLIN,
+		}
 	};
+
+	state.timer_poll_fd = pollfds[EVENT_LOOP_TIMER].fd;
 
 	while (1) {
 		ret = poll(pollfds, sizeof(pollfds) / sizeof(pollfds[0]), -1);
@@ -214,6 +232,27 @@ int main(int argc, char *argv[]) {
 			if (ret < 0) {
 				logprint(ERROR, "pw_loop_iterate failed: %s", spa_strerror(ret));
 				goto error;
+			}
+		}
+
+		if (pollfds[EVENT_LOOP_TIMER].revents & POLLIN) {
+			logprint(TRACE, "event-loop: got a timer event");
+			
+			int timer_fd = pollfds[EVENT_LOOP_TIMER].fd;
+			uint64_t expirations;
+			ssize_t n = read(timer_fd, &expirations, sizeof(expirations));
+			if (n < 0) {
+				logprint(ERROR, "failed to read from timer FD\n");
+				goto error;
+			}
+
+			struct xdpw_timer *timer = state.next_timer;
+			if (timer != NULL) {
+				xdpw_event_loop_timer_func_t func = timer->func;
+				void *user_data = timer->user_data;
+				xdpw_destroy_timer(timer);
+
+				func(user_data);
 			}
 		}
 
--- a/src/core/timer.c
+++ b/src/core/timer.c
@@ -1,0 +1,69 @@
+#include <poll.h>
+#include <wayland-util.h>
+#include <sys/timerfd.h>
+
+#include "xdpw.h"
+#include "logger.h"
+#include "timespec_util.h"
+
+static void update_timer(struct xdpw_state *state) {
+	int timer_fd = state->timer_poll_fd;
+	if (timer_fd < 0) {
+		return;
+	}
+
+	bool updated = false;
+	struct xdpw_timer *timer;
+	wl_list_for_each(timer, &state->timers, link) {
+		if (state->next_timer == NULL ||
+				timespec_less(&timer->at, &state->next_timer->at)) {
+			state->next_timer = timer;
+			updated = true;
+		}
+	}
+
+	if (updated) {
+		struct itimerspec delay = { .it_value = state->next_timer->at };
+		errno = 0;
+		int ret = timerfd_settime(timer_fd, TFD_TIMER_ABSTIME, &delay, NULL);
+		if (ret < 0) {
+			fprintf(stderr, "failed to timerfd_settime(): %s\n",
+				strerror(errno));
+		}
+	}
+}
+
+struct xdpw_timer *xdpw_add_timer(struct xdpw_state *state,
+		uint64_t delay_ns, xdpw_event_loop_timer_func_t func, void *data) {
+	struct xdpw_timer *timer = calloc(1, sizeof(struct xdpw_timer));
+	if (timer == NULL) {
+		logprint(ERROR, "Timer allocation failed");
+		return NULL;
+	}
+	timer->state = state;
+	timer->func = func;
+	timer->user_data = data;
+	wl_list_insert(&state->timers, &timer->link);
+
+	clock_gettime(CLOCK_MONOTONIC, &timer->at);
+	timespec_add(&timer->at, delay_ns);
+
+	update_timer(state);
+	return timer;
+}
+
+void xdpw_destroy_timer(struct xdpw_timer *timer) {
+	if (timer == NULL) {
+		return;
+	}
+	struct xdpw_state *state = timer->state;
+
+	if (state->next_timer == timer) {
+		state->next_timer = NULL;
+	}
+
+	wl_list_remove(&timer->link);
+	free(timer);
+
+	update_timer(state);
+}
--- a/src/core/timespec_util.c
+++ b/src/core/timespec_util.c
@@ -1,0 +1,33 @@
+#include "timespec_util.h"
+#include <time.h>
+
+void timespec_add(struct timespec *t, int64_t delta_ns) {
+	int delta_ns_low = delta_ns % TIMESPEC_NSEC_PER_SEC;
+	int delta_s_high = delta_ns / TIMESPEC_NSEC_PER_SEC;
+
+	t->tv_sec += delta_s_high;
+
+	t->tv_nsec += (long)delta_ns_low;
+	if (t->tv_nsec >= TIMESPEC_NSEC_PER_SEC) {
+		t->tv_nsec -= TIMESPEC_NSEC_PER_SEC;
+		++t->tv_sec;
+	}
+}
+
+bool timespec_less(struct timespec *t1, struct timespec *t2) {
+	if (t1->tv_sec != t2->tv_sec) {
+		return t1->tv_sec < t2->tv_sec;
+	}
+	return t1->tv_nsec < t2->tv_nsec;
+}
+
+bool timespec_is_zero(struct timespec *t) {
+	return t->tv_sec == 0 && t->tv_nsec == 0;
+}
+
+int64_t timespec_diff_ns(struct timespec *t1, struct timespec *t2) {
+	int64_t s = t1->tv_sec - t2->tv_sec;
+	int64_t ns = t1->tv_nsec - t2->tv_nsec;
+
+	return s * TIMESPEC_NSEC_PER_SEC + ns;
+}
--- a/src/screencast/fps_limit.c
+++ b/src/screencast/fps_limit.c
@@ -1,0 +1,70 @@
+#include "fps_limit.h"
+#include "logger.h"
+#include "timespec_util.h"
+#include <stdint.h>
+#include <time.h>
+#include <unistd.h>
+#include <assert.h>
+
+#define FPS_MEASURE_PERIOD_SEC 5.0
+
+void measure_fps(struct fps_limit_state *state, struct timespec *now);
+
+void fps_limit_measure_start(struct fps_limit_state *state, double max_fps) {
+	if (max_fps <= 0.0) {
+		return;
+	}
+
+	clock_gettime(CLOCK_MONOTONIC, &state->frame_last_time);
+}
+
+uint64_t fps_limit_measure_end(struct fps_limit_state *state, double max_fps) {
+	if (max_fps <= 0.0) {
+		return 0;
+	}
+
+	// `fps_limit_measure_start` was not called?
+	assert(!timespec_is_zero(&state->frame_last_time));
+
+	struct timespec now;
+	clock_gettime(CLOCK_MONOTONIC, &now);
+	int64_t elapsed_ns = timespec_diff_ns(&now, &state->frame_last_time);
+
+	measure_fps(state, &now);
+
+	int64_t target_ns = (1.0 / max_fps) * TIMESPEC_NSEC_PER_SEC;
+	int64_t delay_ns = target_ns - elapsed_ns;
+	if (delay_ns > 0) {
+		logprint(TRACE, "fps_limit: elapsed time since the last measurement: %u, "
+			"target %u, should delay for %u (ns)", elapsed_ns, target_ns, delay_ns);
+		return delay_ns;
+	} else {
+		logprint(TRACE, "fps_limit: elapsed time since the last measurement: %u, "
+			"target %u, target not met (ns)", elapsed_ns, target_ns);
+		return 0;
+	}
+}
+
+void measure_fps(struct fps_limit_state *state, struct timespec *now) {
+	if (timespec_is_zero(&state->fps_last_time)) {
+		state->fps_last_time = *now;
+		return;
+	}
+
+	state->fps_frame_count++;
+
+	int64_t elapsed_ns = timespec_diff_ns(now, &state->fps_last_time);
+
+	double elapsed_sec = (double) elapsed_ns / (double) TIMESPEC_NSEC_PER_SEC;
+	if (elapsed_sec < FPS_MEASURE_PERIOD_SEC) {
+		return;
+	}
+
+	double avg_frames_per_sec = state->fps_frame_count / elapsed_sec;
+
+	logprint(DEBUG, "fps_limit: average FPS in the last %0.2f seconds: %0.2f",
+		elapsed_sec, avg_frames_per_sec);
+
+	state->fps_last_time = *now;
+	state->fps_frame_count = 0;
+}
--- a/src/screencast/wlr_screencast.c
+++ b/src/screencast/wlr_screencast.c
@@ -18,6 +18,7 @@
 #include "pipewire_screencast.h"
 #include "xdpw.h"
 #include "logger.h"
+#include "fps_limit.h"
 
 static void wlr_frame_buffer_destroy(struct xdpw_screencast_instance *cast) {
 	// Even though this check may be deemed unnecessary,
@@ -51,7 +52,13 @@ void xdpw_wlr_frame_free(struct xdpw_screencast_instance *cast) {
 		return ;
 	}
 
-	xdpw_wlr_register_cb(cast);
+	uint64_t delay_ns = fps_limit_measure_end(&cast->fps_limit, cast->ctx->state->config->screencast_conf.max_fps);
+	if (delay_ns > 0) {
+		xdpw_add_timer(cast->ctx->state, delay_ns,
+			(xdpw_event_loop_timer_func_t) xdpw_wlr_register_cb, cast);
+	} else {
+		xdpw_wlr_register_cb(cast);
+	}
 }
 
 static int anonymous_shm_open(void) {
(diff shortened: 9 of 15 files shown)


meson.build
```diff
@@ -18,6 +18,7 @@ add_project_arguments(cc.get_supported_arguments([
 	'-Wno-missing-field-initializers',
 	'-Wno-unused-parameter',
 	'-D_POSIX_C_SOURCE=200809L',
+	'-D_GNU_SOURCE',
```
jbeich, line 21: Fails on FreeBSD as `usleep` [was removed](https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xsh_chap01.html#tag_22_01_01_05) in POSIX 2008 (aka X/Open Issue 7). Maybe downgrade `-D_POSIX_C_SOURCE=200809L` to `-D_POSIX_C_SOURCE=200112L` instead of cheating on Linux via `-D_GNU_SOURCE`.
```c
$ meson setup _build && meson compile -C _build
../src/screencast/fps_limit.c:29:4: error: implicit declaration of function 'usleep' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
                        usleep(delay_us);
                        ^
../src/screencast/fps_limit.c:29:4: note: did you mean 'sleep'?
/usr/include/unistd.h:365:15: note: 'sleep' declared here
unsigned int     sleep(unsigned int);
                 ^
1 error generated.
```

zsolt-donca: The suggested flag makes the compilation fail for me on Arch Linux with the exact above error message. I don't have access to FreeBSD, but I could spin one up in a VM if we cannot resolve this otherwise.

I must admit that I have no idea how to fix this. I was taking the above flag from wayvnc's build: https://github.com/any1/wayvnc/blob/13323a742f1449355852b2501b25d8df804cefee/meson.build#L17

Can anybody please help?

emersion: > 4.3BSD, POSIX.1-2001.  POSIX.1-2001 declares this function obsolete; use nanosleep(2) instead.  POSIX.1-2008 removes the specification of usleep()

zsolt-donca: @emersion Thank you! I just pushed another version that uses `nanosleep`, and it actually works without changes to `build.meson`. I hope it works on FreeBSD also.

jbeich: I confirm, builds fine on FreeBSD. Rebase after cf3e02695e3d if you want to see build log.

zsolt-donca: Glad to hear that! I am marking this comment as resolved, then.

I also just fixup'd the fix on the build and `nanosleep` into the first commit, and rebased the whole branch onto the latest master.

maximbaz:
I can confirm that this PR helps a lot on 4k screen, nice job 👍 

For reference, if I run with `-l DEBUG -f 60` or `-l DEBUG -f 10000`, I can see that my frame rate does not exceed `30`, so I assume this is the natural rate, the maximum that I can achieve.

I can only feel that the mouse lag is almost gone at the value `-f 15`, anything higher already feels bad.

Just curious, @zsolt-donca what value did you end up using?

zsolt-donca:
@maximbaz I'm glad it works for you as well!

I ended up using `-f 8`, as that's about the same frame rate that my (company's) favorite video-conferencing application limits to as well (http://8x8.vc/). With this frame rate, I don't feel any mouse lag at all. This is definitely not the highest value I can go without a noticable mouse lag, it's just that anything higher than this is a waste of resources in my case.

include/screencast_common.h
```diff
@@ -91,6 +94,10 @@ struct xdpw_screencast_instance {
 	bool with_cursor;
 	int err;
 	bool quit;
+
+	// fps limit
+	double max_fps; // 0 means no limit
```
danshick, line 99: We don't need to store this in two places. If it is currently implemented for all screencasts, it isn't needed here.

src/screencast/wlr_screencast.c
```diff
@@ -133,6 +134,9 @@ static void wlr_frame_buffer_done(void *data,
 	struct xdpw_screencast_instance *cast = data;
 
 	logprint(TRACE, "wlroots: buffer_done event handler");
+	
+	fps_limit_apply(&cast->fps_limit, cast->max_fps);
```
danshick, line 138: Because xdpw is single-threaded and `nanosleep` ties up our only thread, calling this function here will delay all of xdpw. This probably doesn't create any issues for us, but it would be better if we could conditionally wait to request another frame in the event loop.

This could be addressed by queueing the call to reregister the frame listener callback ([currently invoked here](https://github.com/emersion/xdg-desktop-portal-wlr/blob/e103e120e20e83acec5b66ce4cd33a0eb86df0b0/src/screencast/wlr_screencast.c#L54)) and invoking the next calls in that queue [somewhere in the event loop](https://github.com/emersion/xdg-desktop-portal-wlr/blob/e103e120e20e83acec5b66ce4cd33a0eb86df0b0/src/core/main.c#L175-L217) after checking that the required amount of time has passed.

I'm curious to hear @emersion's thoughts as to whether what I'm suggesting is overkill or not.

include/screencast_common.h
zsolt-donca: The reason I added `max_fps` here is so that I could use it in `wlr_frame_buffer_done`:

```C
static void wlr_frame_buffer_done(void *data,
		struct zwlr_screencopy_frame_v1 *frame) {
	struct xdpw_screencast_instance *cast = data;

	logprint(TRACE, "wlroots: buffer_done event handler");

	fps_limit_apply(&cast->fps_limit, cast->max_fps);
...
```

Since this is a callback, I presume the only way to pass information to it is to use include it in the `void *data` argument. How am I supposed to pass this information here otherwise? The only other way I can think of is to introduce a global variable, but I am not sure if that's a good idea.

danshick: The screencast context is [available from any screencast instance](https://github.com/emersion/xdg-desktop-portal-wlr/blob/e103e120e20e83acec5b66ce4cd33a0eb86df0b0/include/screencast_common.h#L73) via
```c
cast->ctx;
```

Doing it this way keeps us from maintaining this information in more than one place.

zsolt-donca: You are right, I completely missed the context being already there.

src/screencast/wlr_screencast.c
emersion: Yes, I agree we should do our best not to block anywhere.

For reference, mako's event loop allows registering timers: https://github.com/emersion/mako/blob/dbd9c2bc98ef0cec7801a049f4bf8d61f7be36d0/event-loop.c#L128

zsolt-donca: How do you want to proceed? Introducing support for this in the event loop and registering callbacks looks like a major change. Do you, maintainers, want to work on this, or should I try to implement it? My C skills are rusty but I think I might be able to figure it out, though I cannot make any promises.

danshick: Your C looks at least as good as mine. I'd be fine with you taking a crack at it if you are willing. I'll do my best to review as you go and @emersion has always been super helpful with my crappy C.

You may also want to look at https://github.com/any1/aml, which is the eventloop wayvnc/neatvnc uses. I pointed you to that code base once before, so maybe you're already more familiar with it.

I'm not sure we want to bother using aml as a dependency, but maybe? Should be easy enough to register timers without it, but it may still be a good reference for how to do it.

zsolt-donca: When taking a look at wayvnc, I've noticed that it uses https://github.com/any1/aml as its basic event loop mechanism, including using it for timing the next frame capture. I was basically hoping that this sleep-based solution in this PR would be good enough for `xdg-desktop-portal-wlr` :P

With regards to the technical implementation, which option do you prefer, @danshick @emersion ? Should I go with integrating `aml` into this codebase, which would replace its exisitng event loop, or should I extend the existing event loop with support for timers, based on something like `mako`? The latter makes more sense to me, since its simpler and both `mako` and `xdg-desktop-portal-wlr` are @emersion 's projects. I would just like to confirm before spending too much time with this.

danshick: Yeah, I think following mako's example makes the most sense. I only shared the link to aml as an alternative example. I don't think we need that level of capability here.

emersion: Yeah, I've had a look at aml a while ago. My main gripes with it are:

- It involves threading, which always makes things more complicated. The workers could just have been left out, it's easy for the caller to use a pipe instead.
- It involves reference counting for all of its data structures, which is pretty unnecessary for an event loop. Also has a global registry of object IDs for some reason.
- Lack of type safety, many functions take a `void *`.

I still haven't found any go-to event loop library. The Wayland server event loop is pretty good, but I'm not too comfortable depending on wayland-server in a client.

tl;dr is that I've been including a mini event loop in each client for now, with just enough features for the client's purposes.

zsolt-donca: Added a new commit that introduces timers based on [emersion/mako](https://github.com/emersion/mako)'s timer implementation, mostly copy-pasting the changes required for the definition of timers and the event loop, but changing the `mako` prefix to `xdpw`. Also changed the resolution, from milliseconds to nanoseconds.

Also adapted the rate control code now to the new method, which involved some refactoring as well. Retested the actual rate limit, and I think it is more precise now. 

I did not manage to execute the suggested test of running xdpw with a capture configured to a reallt low frame-rate (e.g. `0.1` FPS) and doing a screenshot in the meantime, because the link to `trigger-screen-shot.py` in the [Contributing](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/CONTRIBUTING.md) guide is now dead. Got no luck with [portal-test](https://github.com/matthiasclasen/portal-test) either, even though I managed to build it locally, it was running buggy and, for example, could not do two subsequent screenshots with it (triggered errors in xdpw). Does anybody still have `trigger-screen-shot.py` somewhere?

danshick: I may have it lying around on a local dev machine. I'll see if I can dig it up.

You might also try [ksnip](https://github.com/ksnip/ksnip). Not sure how easy it is to explicitly specify portal screenshot as the technique, I've never used it.

src/screencast/wlr_screencast.c
```diff
@@ -34,6 +35,11 @@ static void wlr_frame_buffer_destroy(struct xdpw_screencast_instance *cast) {
 	}
 }
 
+static void register_callback(void *data) {
```
danshick, line 38: Is this being added only to change the argument to a void pointer?

danshick (commented):
This looks good to me for the most part. I'll test this PR soon.

src/screencast/wlr_screencast.c
zsolt-donca: Yeah, it is. I have no idea how to cast function pointers ☺️ I am open to suggestions.

danshick: I'd say just rewrite the original function signature and add the type cast there if it is going to be regularly called by a callback. It won't affect existing functionality.

We'll lose a small amount of type safety but I don't know of a better way either.

zsolt-donca: I just pushed a commit that removes this function and replaces its use with an actual cast to `xdpw_wlr_register_cb`. There was no need after all to change the function `xdpw_wlr_register_cb`. The cast turned out to be simple, I don't know why it didn't compile for me when I tried it before. Anyway, I am marking this as resolved - @danshick please reopen if you don't like the way it came out.

zsolt-donca:
I just realized that in certain places I am not respecting the [contribution guide on line length](https://github.com/swaywm/wlroots/blob/master/CONTRIBUTING.md#line-length):
> Try to keep your lines under 80 columns, but you can go up to 100 if it improves readability. Don't break lines indiscriminately, try to find nice breaking points so your code is easy to read.

I will soon update my PR to fix this.

src/screencast/wlr_screencast.c
zsolt-donca: I managed to test this with ksnip. It worked well. Based on the trace logs, I could see that knsip was able to take the screen shot using my `xdgw` instance while otherwise it was "waiting" between two consecutive frames for screen sharing (10 seconds with FPS = 0.1).

src/screencast/wlr_screencast.c
danshick: Yeah, this looks fine! I didn't know that cast would work TBH but this is much better.

danshick:
BTW, your FreeBSD builds are failing, which we need to resolve before we can merge this.

Mako fixes your issues with libepoll-shim on FreeBSD.

You need to do this in the meson.build:

https://github.com/emersion/mako/blob/dbd9c2bc98ef0cec7801a049f4bf8d61f7be36d0/meson.build#L32-L36

And you need to edit the .builds/freebsd.yml to include this:

https://github.com/emersion/mako/blob/dbd9c2bc98ef0cec7801a049f4bf8d61f7be36d0/.builds/freebsd.yml#L5

zsolt-donca:
@danshick sorry about that, I wasn't paying attention to the automated builds linked in here. Thanks for info on how to fix it as well.

include/fps_limit.h
```diff
@@ -0,0 +1,18 @@
+#ifndef FPS_LIMIT_H
+#define FPS_LIMIT_H
+
+#include <stdint.h>
+#include <time.h>
+
+struct fps_limit_state {
+	struct timespec last_time;
```
emersion, line 8: Maybe a better name would be `frame_start_time`?

src/screencast/fps_limit.c
```diff
@@ -0,0 +1,67 @@
...
+void measure_fps(struct fps_limit_state* state, struct timespec *now) {
+	if (timespec_is_zero(&state->fps_last_time)) {
+		state->fps_last_time = *now;
+	} else {
+		state->fps_frame_count++;
+
+		int64_t elapsed_ns = timespec_diff_ns(now, &state->fps_last_time);
+
+		double elapsed_sec = (double) elapsed_ns / (double) TIMESPEC_NS_IN_S;
+		if (elapsed_sec >= FPS_MEASURE_PERIOD_SEC) {
```
emersion, line 56: Here again, I think an early return can improve readability.

src/screencast/fps_limit.c
```diff
@@ -0,0 +1,67 @@
...
+			double avg_delay_sec = elapsed_sec / state->fps_frame_count;
+			double avg_frames_per_sec = 1.0 / avg_delay_sec;
```
emersion, line 58: Can't we simplify these two lines?

```c
double avg_frames_per_sec = state->fps_frame_count / elapsed_sec;
```

src/screencast/fps_limit.c
```diff
@@ -0,0 +1,67 @@
...
+
+			logprint(DEBUG, "fps_limit: average FPS in the last %0.0f seconds: %0.2f", 
+				FPS_MEASURE_PERIOD_SEC, avg_frames_per_sec);
+
+			state->fps_last_time = *now;
+			state->fps_frame_count = 0;
```
emersion, line 64: Style: trailing whitespace

emersion (changes requested):
Overall the logic looks good to me. Here are some comments to improve readability and style.

emersion:
Also needs a rebase on top of master.

columbarius:
When you rebase it, could you please add fps_limit to screencast_conf and the config file? I think that could be useful. 
@emersion @danshick Any objections?

emersion:
Yup looks like a good idea to me

include/timespec_util.h
```diff
@@ -0,0 +1,18 @@
+#ifndef TIMESPEC_UTIL_H
+#define TIMESPEC_UTIL_H
+
+#include <time.h>
+#include <stdbool.h>
+#include <stdint.h>
+
+#define TIMESPEC_NS_IN_S 1000000000L
```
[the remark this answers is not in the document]
danshick: I read this as X many nanoseconds in a second, so it makes sense to me. I could see it either way.

meson.build
```diff
@@ -23,6 +23,12 @@ pipewire = dependency('libpipewire-0.3', version: '>= 0.3.2')
 wayland_client = dependency('wayland-client')
 wayland_protos = dependency('wayland-protocols', version: '>=1.14')
 
+epoll = dependency('', required: false)
+if (not cc.has_function('timerfd_create', prefix: '#include <sys/timerfd.h>') or
```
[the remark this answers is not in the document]
zsolt-donca: I might be misunderstanding what you mean here, but if I change this code from

```build
if (not cc.has_function('timerfd_create', prefix: '#include <sys/timerfd.h>') or
    not cc.has_function('signalfd', prefix: '#include <sys/signalfd.h>'))
	epoll = dependency('epoll-shim')
endif
```

to this code

```build
if not cc.has_function('timerfd_create', prefix: '#include <sys/timerfd.h>') or
    not cc.has_function('signalfd', prefix: '#include <sys/signalfd.h>')
	epoll = dependency('epoll-shim')
endif
```

then `ninja -C build` it complaints with: `../meson.build:27:7: ERROR: Unknown statement.`, where line 27 is the line with the `if` statement.

src/core/main.c
```diff
@@ -42,12 +47,14 @@ int main(int argc, char *argv[]) {
...
 	bool replace = false;
+	double max_fps = 0;
 
-	static const char* shortopts = "l:o:rh";
+	static const char* shortopts = "l:o:f:rh";
 	static const struct option longopts[] = {
 		{ "loglevel", required_argument, NULL, 'l' },
 		{ "output", required_argument, NULL, 'o' },
 		{ "replace", no_argument, NULL, 'r' },
+		{ "max-fps", required_argument, NULL, 'f' },
```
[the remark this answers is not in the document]
zsolt-donca: I was thinking the same thing, didn't want to start working on it before the configs were merged. Will update the PR soon with this.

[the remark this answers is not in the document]
emersion: Oh, because it's multi-line. Makes sense.

src/core/main.c
```diff
@@ -170,8 +182,14 @@ int main(int argc, char *argv[]) {
 			.fd = pw_loop_get_fd(state.pw_loop),
 			.events = POLLIN,
 		},
+		[EVENT_LOOP_TIMER] = {
+			.fd = timerfd_create(CLOCK_MONOTONIC, TFD_CLOEXEC),
+			.events = POLLIN,
+		}
 	};
 
+	state.timer_pollfd = &pollfds[EVENT_LOOP_TIMER];
```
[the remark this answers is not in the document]
zsolt-donca: Yes, good idea. I updated the field and also renamed it to `timer_poll_fd`, as this is now just an `fd`, not a `pollfd` struct.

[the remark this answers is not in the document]
zsolt-donca: I am marking this comment as resolved as I think I addressed the comment entirely; please let me know if in the future you would like me to leave these conversations open if you would prefer to confirm and close them.

src/core/timer.c
```diff
@@ -0,0 +1,69 @@
...
+		int ret = timerfd_settime(timer_fd, TFD_TIMER_ABSTIME, &delay, NULL);
+		if (ret < 0) {
+			fprintf(stderr, "failed to timerfd_settime(): %s\n",
+				strerror(errno));
+		}
+	}
+}
+
+struct xdpw_timer *xdpw_add_timer(struct xdpw_state *state,
+	uint64_t delay_ns, xdpw_event_loop_timer_func_t func, void *data) {
```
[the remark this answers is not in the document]
zsolt-donca: Yes, good catch. Updated this.

src/core/timespec_util.c
```diff
@@ -0,0 +1,34 @@
+
```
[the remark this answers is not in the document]
zsolt-donca: Updated, thanks.

src/screencast/wlr_screencast.c
```diff
@@ -51,7 +52,13 @@ void xdpw_wlr_frame_free(struct xdpw_screencast_instance *cast) {
 		return ;
 	}
 
-	xdpw_wlr_register_cb(cast);
+	uint64_t delay_ns = fps_limit_measure_end(&cast->fps_limit, cast->ctx->max_fps);
+	if (delay_ns > 0) {
+		xdpw_add_timer(cast->ctx->state, delay_ns, 
+		(xdpw_event_loop_timer_func_t) xdpw_wlr_register_cb, cast);
```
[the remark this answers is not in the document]
zsolt-donca: Good catch, I did not intend for this to look like it. Updated.

src/screencast/fps_limit.c
```diff
@@ -0,0 +1,67 @@
...
+#include <stdint.h>
+#include <time.h>
+#include <unistd.h>
+
+#define FPS_MEASURE_PERIOD_SEC 5.0
+
+void measure_fps(struct fps_limit_state* state, struct timespec *now);
+
+void fps_limit_measure_start(struct fps_limit_state* state, double max_fps) {
+	if (max_fps <= 0.0)
```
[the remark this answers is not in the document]
zsolt-donca: Sorry about this; I actually read the style guide a while ago when I first started working on this, this just slipped through. Updated.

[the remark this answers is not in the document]
zsolt-donca: I used some regexp to look for other cases of missing brances, and also took the libery of updating another missing case in the codebase. I hope you don't mind. For details, check https://github.com/emersion/xdg-desktop-portal-wlr/pull/74/commits/06d36a084e8cf2b83196a58d76844154201ba9a3

src/screencast/fps_limit.c
```diff
@@ -0,0 +1,67 @@
...
+		logprint(TRACE, "fps_limit: elapsed time since the last measurement: %u, "
+			"target %u, target not met (ns)", elapsed_ns, target_ns);
+		return 0;
+	}
+}
+
+void measure_fps(struct fps_limit_state* state, struct timespec *now) {
+	if (timespec_is_zero(&state->fps_last_time)) {
+		state->fps_last_time = *now;
+	} else {
```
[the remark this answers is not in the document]
zsolt-donca: Updated.

src/screencast/fps_limit.c
```diff
@@ -0,0 +1,67 @@
...
+void measure_fps(struct fps_limit_state* state, struct timespec *now);
+
+void fps_limit_measure_start(struct fps_limit_state* state, double max_fps) {
+	if (max_fps <= 0.0) 
+		return;
+
+	clock_gettime(CLOCK_MONOTONIC, &state->last_time);
+}
+
+uint64_t fps_limit_measure_end(struct fps_limit_state* state, double max_fps) {
```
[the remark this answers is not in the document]
zsolt-donca: Fixed this, and other cases that I found. I used some regexp to search, and updated all places that I found, including lines that were otherwise not affected by my code. For details see https://github.com/emersion/xdg-desktop-portal-wlr/pull/74/commits/6037bc3967e9ffdf47d47473e9a505db951cd80f . I hope you don't mind; if you'd like me to revert the changes that on lines that I otherwise did not change, let me now.

src/screencast/fps_limit.c
```diff
@@ -0,0 +1,67 @@
+#include "fps_limit.h"
+#include "logger.h"
+#include "timespec_util.h"
+#include <stdint.h>
+#include <time.h>
+#include <unistd.h>
+
+#define FPS_MEASURE_PERIOD_SEC 5.0
+
+void measure_fps(struct fps_limit_state* state, struct timespec *now);
```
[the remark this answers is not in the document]
zsolt-donca: Updated this in https://github.com/emersion/xdg-desktop-portal-wlr/pull/74/commits/6037bc3967e9ffdf47d47473e9a505db951cd80f

src/screencast/fps_limit.c
```diff
@@ -0,0 +1,67 @@
...
+
+	clock_gettime(CLOCK_MONOTONIC, &state->last_time);
+}
+
+uint64_t fps_limit_measure_end(struct fps_limit_state* state, double max_fps) {
+	if (max_fps <= 0.0) 
+		return 0;
+
+	if (timespec_is_zero(&state->last_time)) {
+		// `fps_limit_measure_start` was not called?
```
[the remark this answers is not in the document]
zsolt-donca: I am not sure. Do we want to abort the program in case the assertion fails? I mean, limiting the number of frames captured by the application is not a critical feature; in case something goes wrong here, I could imagine that we'd want to continue rather than to crash.

Let me know what you think; if you want me, I can certainly add an assertion here.

include/timespec_util.h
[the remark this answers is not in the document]
zsolt-donca: Well, the fact that we are talking about this is proof enough that the original naming was ambigous. I like the suggested naming, and updated the name of the constant to `TIMESPEC_NSEC_PER_SEC`, with using the `TIMESPEC` prefix since this is part of `timespec_util.h`'s interface.

Let me know if you think this is okay, or if I should just rename it to `NSEC_PER_SEC`.

include/fps_limit.h
zsolt-donca: Good idea, updated.

src/screencast/fps_limit.c
zsolt-donca: Updated.

src/screencast/fps_limit.c
zsolt-donca: I agree, the suggested formula is simple enough. Updated.

zsolt-donca:
@emersion I think I addressed all your comments, thank you for the review. Please note that some questions are still open.

@emersion @danshick  Will come back soon with a rebase on top of the current master, and with support for the fps limit in the newly introduced config file.

danshick:
Thanks for all of your work on this and your patience with us through these reviews and redesigns.

src/screencast/fps_limit.c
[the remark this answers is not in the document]
danshick: No, good catch! Feel free to fix anything you find.

src/screencast/fps_limit.c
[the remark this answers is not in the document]
danshick: Nope, this is great. Thanks for the fixes.

src/core/main.c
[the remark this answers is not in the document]
zsolt-donca: FYI this is the fix: https://github.com/emersion/xdg-desktop-portal-wlr/pull/74/commits/e992ef464d452cec675c99a2661424836acb8dd8

src/core/timer.c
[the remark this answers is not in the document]
zsolt-donca: FYI this is the fix: https://github.com/emersion/xdg-desktop-portal-wlr/pull/74/commits/d275db835fd18f83c6377bc833f5625b8e1a02a9

src/screencast/fps_limit.c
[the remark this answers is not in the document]
zsolt-donca: Thanks, @danshick ! Then I'll mark this as Resolved.

src/screencast/fps_limit.c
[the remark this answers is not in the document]
zsolt-donca: Thanks! I'll mark this as resolved, then.

include/fps_limit.h
zsolt-donca: FYI https://github.com/emersion/xdg-desktop-portal-wlr/pull/74/commits/af7f5b29befafa0b6664d6814b8f8a4445019139

include/timespec_util.h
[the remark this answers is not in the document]
emersion: Yup, WFM!

src/screencast/fps_limit.c
zsolt-donca: FYI https://github.com/emersion/xdg-desktop-portal-wlr/pull/74/commits/3b9d96b3d923dad710df3c92789b455add3c906a

meson.build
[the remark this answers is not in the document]
emersion: I see, the parentheses are necessary because of the continuation line.

src/screencast/fps_limit.c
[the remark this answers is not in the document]
emersion: In general I prefer to fail hard when I end up in an unexpected state, this makes the users report this as an issue and allows us to fix potential bugs instead of ending up with more subtle bugs that are harder to debug.

I'm not feeling strongly about this particular case though, feel free to ignore my comment.

src/screencast/fps_limit.c
zsolt-donca: Addressed. Also cleaned up trailing whitespace in other places: https://github.com/emersion/xdg-desktop-portal-wlr/pull/74/commits/9709807ebc25445648c5f3ef2a6012e33f423c82

zsolt-donca:
I rebased the entire branch on the current master. Introduced the key `screencast:max_fps` in the config file; for example:

```
[screencast]
output_name=DP-6
max_fps=8
```

Note that the sample config had an error and was using `ouput` instead of `output_name`. Fixed it in this commit: https://github.com/emersion/xdg-desktop-portal-wlr/pull/74/commits/9acf4eb12ed6582759063d65a6b97d878c0ec2ca 

Also note that I could not test the log statements and, in particular, the fps measurement after doing this rebase because of https://github.com/emersion/xdg-desktop-portal-wlr/issues/91 . Will retest everything once that's fixed.

src/screencast/fps_limit.c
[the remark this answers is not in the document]
zsolt-donca: I'll do some more thorough testing with the assertion in place after https://github.com/emersion/xdg-desktop-portal-wlr/issues/91 is fixed, to check whether I actually have this condition met in the code. I'll probably keep the condition afterwards.

src/screencast/wlr_screencast.c
zsolt-donca: I'll mark this thread closed since this was implemented. Please reopen if you have further comments on this subject.

danshick:
@zsolt-donca You can comment out https://github.com/emersion/xdg-desktop-portal-wlr/blob/07154bb1e3964bb0b319944b1a725907bd757b00/src/core/config.c#L62 for testing, but we do need to resolve that issue regardless.

zsolt-donca:
I rebased this branch once again on the latest master. I also changed a defensive check with an actual assertion as suggested by @emersion . 

I also found a minor issue with how the fps measurements were logged, as it was printing the elapsed time threshold instead of the actual elapsed time (https://github.com/emersion/xdg-desktop-portal-wlr/pull/74/commits/f6c44ca51359a4cc4394aeeb9b00f973ce31f550). I think the timing varies also because we are capturing a frame only if there is damage. Sample log with the new code:

```
2021/03/05 09:42:32 [DEBUG] - fps_limit: average FPS in the last 5.44 seconds: 2.02
2021/03/05 09:42:37 [DEBUG] - fps_limit: average FPS in the last 5.01 seconds: 2.00
2021/03/05 09:42:43 [DEBUG] - fps_limit: average FPS in the last 5.50 seconds: 2.00
2021/03/05 09:42:48 [DEBUG] - fps_limit: average FPS in the last 5.02 seconds: 1.99
2021/03/05 09:42:53 [DEBUG] - fps_limit: average FPS in the last 5.51 seconds: 2.00
2021/03/05 09:42:59 [DEBUG] - fps_limit: average FPS in the last 5.49 seconds: 2.00
```

I also retested all functionality that I could think of, including the new ones with the config file (e.g. the CLI argument takes precedence over the config file, if both are missing then there is no limit).

This should now addres all feedback that I received so far. What should be the next step? Should I squash all my commits into a single one so that it can be merged?

src/screencast/fps_limit.c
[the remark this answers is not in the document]
zsolt-donca: I replaced the check with an assert statement, see https://github.com/emersion/xdg-desktop-portal-wlr/pull/74/commits/f06bc743f00736b8cbd5d25701e1671f62dba000 . I did some further testing, with various screensharing sessions going on and off, and didn't find any issue. I am marking this comment as resolved, please consider reopening if you have further comments.

danshick:
> I think the timing varies also because we are capturing a frame only if there is damage.

Yup, exactly what I'd expect.

> I also retested all functionality...

Thanks!

> This should now addres all feedback that I received so far. What should be the next step? Should I squash all my commits into a single one so that it can be merged?

Yeah, normally at this point, I'd do an interactive rebase to try to reduce the overall number of commits. We tend towards a rebase and merge strategy, and for a reasonably big PR, it is okay if it comes over as more than one commit.

I'd want @emersion to take a final look and I'll start running your PR as a daily driver for a couple of days to make sure I don't bump into any major bugs, but I think this should be approved and merged shortly.

emersion (approved):
Code LGTM, thanks!

@zsolt-donca, can you rebase and split your work into logical commits?

@danshick, if testing goes well, feel free to merge!

zsolt-donca:
Are you guys okay with me squashing the whole thing into a single commit for this pull request? I can't reasonably think of a better way, especially now that the master branch has support for config files.

Just a small question: what should I set in `config.sample` to `max_fps`? Especially now that we are considering installing it as per https://github.com/emersion/xdg-desktop-portal-wlr/issues/92, we should probably put the default values, which for `max_fps` would be `0`, meaning no limit.

emersion:
> Are you guys okay with me squashing the whole thing into a single commit for this pull request?

OK. There are a few unrelated style changes, but it's no big deal.

> Just a small question: what should I set in config.sample to max_fps?

We can always adjust it if/when we decide to ship it.

danshick:
@zsolt-donca Lemme know when you've done any last minute cleanup/squashing and I'll merge this. I haven't seen any crashing or issues over the past couple of days.

zsolt-donca:
@danshick 
> Lemme know when you've done any last minute cleanup/squashing...

I just squashed the whole thing into a single commit. This should be it from me. I'd appreciate if you could give feedback on the commit message, in particular if it is okay with regards to the [contributing guide](https://github.com/swaywm/wlroots/blob/master/CONTRIBUTING.md#commit-messages).

emersion:
Ah, thanks for taking the time to write a good commit message, greatly appreciated! Yes, it looks good to me.

Since @danshick gave their greenlight, let's just merge this :)

Thanks a lot!

zsolt-donca:
@emersion @danshick I would also like to thank you for your guidance and meticulous reviews! It was fun working on this, I enjoyed it. :smile:

danshick:
> It was fun working on this, I enjoyed it. :smile: 

We're happy to have your contributions!

Issue #67, opened 2020-11-11
Man pages


emersion:
We should probably add a man page for official docs. I'd prefer to keep distribution-specific and app-specific instructions in the wiki.

danshick:
I'm planning to write these soon. I'll post a PR when ready for feedback.

danshick:
I think we can close this with #98, but I do still hope to add to them in the future.

PR #59, opened 2020-10-08
Add graphical display chooser

columbarius:
This provides the posibility to choose the screencast display via a graphical interface provided either by slurp, wofi or dmenu. If a display is given via, the commandline parameter, it is prioritised, in case all selectors fail, it will fall back to the current default. For slurp to work properly https://github.com/emersion/slurp/pull/64 is required.

Note: while this is not merged and man pages are not ready, the locations for a config file are in order:

    $XDG_CONFIG_HOME/xdg-desktop-portal-wlr/$XDG_CURRENT_DESKTOP
    $XDG_CONFIG_HOME/xdg-desktop-portal-wlr/config
    $HOME/.config/xdg-desktop-portal-wlr/$XDG_CURRENT_DESKTOP
    $HOME/.config/xdg-desktop-portal-wlr/config
    /etc/xdg/xdg-desktop-portal-wlr/$XDG_CURRENT_DESKTOP
    /etc/xdg/xdg-desktop-portal-wlr/config
    /etc/xdg-desktop-portal-wlr/$XDG_CURRENT_DESKTOP
    /etc/xdg-desktop-portal-wlr/config

The system wide config can be in another place, if SYSCONFDIR differs from /etc

PR #75, opened 2020-12-30
Add support for config file

### LINKED ISSUE #60: Config file format
To offer custom options for an outputchooser we need a config file #59 .
To make it easily readable, we should stick to a plaintext, or sway like config.
I could think of an ini like file, where each section could represent a different portal.

Is there something easy adoptable or an often used library?

columbarius:
Resolves #60 

Note: while this is not merged and man pages are not ready, the locations for a config file are in order:
* $XDG_CONFIG_HOME/xdg-desktop-portal-wlr/$XDG_CURRENT_DESKTOP
* $XDG_CONFIG_HOME/xdg-desktop-portal-wlr/config
* $HOME/.config/xdg-desktop-portal-wlr/$XDG_CURRENT_DESKTOP
* $HOME/.config/xdg-desktop-portal-wlr/config
* /etc/xdg/xdg-desktop-portal-wlr/$XDG_CURRENT_DESKTOP
* /etc/xdg/xdg-desktop-portal-wlr/config
* /etc/xdg-desktop-portal-wlr/$XDG_CURRENT_DESKTOP
* /etc/xdg-desktop-portal-wlr/config

The system wide config can be in another place, if SYSCONFDIR differs from /etc

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -13,6 +13,6 @@ struct xdpw_config {
 
 void print_config(enum LOGLEVEL loglevel, struct xdpw_config *config);
 void finish_config(struct xdpw_config *config);
-void init_config(const char *configfile, struct xdpw_config *config);
+void init_config(const char **configfile, struct xdpw_config *config);
 
 #endif
--- a/include/config.h
+++ b/include/config.h
@@ -13,6 +13,6 @@ struct xdpw_config {
 
 void print_config(enum LOGLEVEL loglevel, struct xdpw_config *config);
 void finish_config(struct xdpw_config *config);
-void init_config(char *configfile, struct xdpw_config *config);
+void init_config(const char *configfile, struct xdpw_config *config);
 
 #endif
--- a/include/config.h
+++ b/include/config.h
@@ -13,6 +13,6 @@ struct xdpw_config {
 
 void print_config(enum LOGLEVEL loglevel, struct xdpw_config *config);
 void finish_config(struct xdpw_config *config);
-void init_config(const char **configfile, struct xdpw_config *config);
+void init_config(char ** const configfile, struct xdpw_config *config);
 
 #endif
--- a/include/config.h
+++ b/include/config.h
@@ -1,0 +1,18 @@
+#ifndef CONFIG_H
+#define CONFIG_H
+
+#include "logger.h"
+
+struct config_screencast {
+	char *output_name;
+};
+
+struct xdpw_config {
+	struct config_screencast screencast_conf;
+};
+
+void print_config(enum LOGLEVEL loglevel, struct xdpw_config *config);
+void finish_config(struct xdpw_config *config);
+void init_config(char *configfile, struct xdpw_config *config);
+
+#endif
--- a/include/logger.h
+++ b/include/logger.h
@@ -3,6 +3,8 @@
 
 #include <stdio.h>
 
+#define DEFAULT_LOGLEVEL ERROR
+
 enum LOGLEVEL { QUIET, ERROR, WARN, INFO, DEBUG, TRACE };
 
 struct logger_properties {
--- a/meson.build
+++ b/meson.build
@@ -30,7 +30,7 @@ rt = cc.find_library('rt')
 pipewire = dependency('libpipewire-0.3', version: '>= 0.3.2')
 wayland_client = dependency('wayland-client')
 wayland_protos = dependency('wayland-protocols', version: '>=1.14')
-iniparser = cc.find_library('iniparser')
+iniparser = cc.find_library('iniparser', dirs: [join_paths(get_option('prefix'),get_option('libdir'))])
 
 if get_option('sd-bus-provider') == 'auto'
 	assert(get_option('auto_features').auto(), 'sd-bus-provider must not be set to auto since auto_features != auto')
--- a/meson.build
+++ b/meson.build
@@ -20,12 +20,17 @@ add_project_arguments(cc.get_supported_arguments([
 	'-D_POSIX_C_SOURCE=200809L',
 ]), language: 'c')
 
+prefix = get_option('prefix')
+sysconfdir = get_option('sysconfdir')
+add_project_arguments('-DSYSCONFDIR="@0@"'.format(join_paths(prefix, sysconfdir)), language : 'c')
+
 inc = include_directories('include')
 
 rt = cc.find_library('rt')
 pipewire = dependency('libpipewire-0.3', version: '>= 0.3.2')
 wayland_client = dependency('wayland-client')
 wayland_protos = dependency('wayland-protocols', version: '>=1.14')
+iniparser = cc.find_library('iniparser')
 
 if get_option('sd-bus-provider') == 'auto'
 	assert(get_option('auto_features').auto(), 'sd-bus-provider must not be set to auto since auto_features != auto')
@@ -59,6 +64,7 @@ executable(
 	files([
 		'src/core/main.c',
 		'src/core/logger.c',
+		'src/core/config.c',
 		'src/core/request.c',
 		'src/core/session.c',
 		'src/screenshot/screenshot.c',
@@ -73,6 +79,7 @@ executable(
 		sdbus,
 		pipewire,
 		rt,
+		iniparser,
 	],
 	include_directories: [inc],
 	install: true,
--- a/src/core/config.c
+++ b/src/core/config.c
@@ -10,7 +10,7 @@
 #include <iniparser.h>
 
 void print_config(enum LOGLEVEL loglevel, struct xdpw_config *config) {
-	logprint(loglevel, "config: Outputname  %s",config->screencast_conf.output_name);
+	logprint(loglevel, "config: Outputname  %s", config->screencast_conf.output_name);
 }
 
 // NOTE: calling finish_config won't prepare the config to be read again from config file
@@ -22,7 +22,8 @@ void finish_config(struct xdpw_config *config) {
 	free(&config->screencast_conf.output_name);
 }
 
-static void getstring_from_conffile(dictionary *d, char *key, char **dest, char *fallback) {
+static void getstring_from_conffile(dictionary *d,
+		const char *key, char **dest, const char *fallback) {
 	if (*dest != NULL) {
 		return;
 	}
@@ -55,7 +56,7 @@ static char *config_path(char *prefix, char *filename) {
 	return path;
 }
 
-static void config_parse_file(char *configfile, struct xdpw_config *config) {
+static void config_parse_file(const char *configfile, struct xdpw_config *config) {
 	dictionary *d = iniparser_load(configfile);
 
 	// screencast
@@ -99,7 +100,7 @@ static char *get_config_path(void) {
 	return NULL;
 }
 
-void init_config(char *configfile, struct xdpw_config *config) {
+void init_config(const char *configfile, struct xdpw_config *config) {
 	if (configfile == NULL) {
 		configfile = get_config_path();
 	}
--- a/src/core/config.c
+++ b/src/core/config.c
@@ -1,0 +1,109 @@
+#include "config.h"
+#include "xdpw.h"
+#include "logger.h"
+
+#include <dictionary.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <iniparser.h>
+
+void print_config(enum LOGLEVEL loglevel, struct xdpw_config *config) {
+	logprint(loglevel, "config: Outputname  %s",config->screencast_conf.output_name);
+}
+
+// NOTE: calling finish_config won't prepare the config to be read again from config file
+// with init_config since to pointers and other values won't be reset to NULL, or 0
+void finish_config(struct xdpw_config *config) {
+	logprint(DEBUG, "config: destroying config");
+
+	// screencast
+	free(&config->screencast_conf.output_name);
+}
+
+static void getstring_from_conffile(dictionary *d, char *key, char **dest, char *fallback) {
+	if (*dest != NULL) {
+		return;
+	}
+	const char *c = iniparser_getstring(d, key, fallback);
+	if (c == NULL) {
+		return;
+	}
+	// Allow keys without value as default
+	if (strcmp(c, "") != 0) {
+		*dest = strdup(c);
+	} else {
+		*dest = fallback ? strdup(fallback) : NULL;
+	}
+}
+
+static bool file_exists(const char *path) {
+	return path && access(path, R_OK) != -1;
+}
+
+static char *config_path(char *prefix, char *filename) {
+	if (!prefix || !prefix[0] || !filename || !filename[0]) {
+		return NULL;
+	}
+
+	char *config_folder = "xdg-desktop-portal-wlr";
+
+	size_t size = 3 + strlen(prefix) + strlen(config_folder) + strlen(filename);
+	char *path = calloc(size, sizeof(char));
+	snprintf(path, size, "%s/%s/%s", prefix, config_folder, filename);
+	return path;
+}
+
+static void config_parse_file(char *configfile, struct xdpw_config *config) {
+	dictionary *d = iniparser_load(configfile);
+
+	// screencast
+	getstring_from_conffile(d, "screencast:output_name", &config->screencast_conf.output_name, NULL);
+
+	iniparser_freedict(d);
+	logprint(DEBUG, "config: config file parsed");
+	print_config(DEBUG, config);
+}
+
+static char *get_config_path(void) {
+	const char *home = getenv("HOME");
+	size_t size_fallback = 1 + strlen(home) + strlen("/.config");
+	char *config_home_fallback = calloc(size_fallback, sizeof(char));
+	snprintf(config_home_fallback, size_fallback, "%s/.config", home);
+
+	char *prefix[4];
+	prefix[0] = getenv("XDG_CONFIG_HOME");
+	prefix[1] = config_home_fallback;
+	prefix[2] = SYSCONFDIR "/xdg";
+	prefix[3] = SYSCONFDIR;
+
+	char *config[2];
+	config[0] = getenv("XDG_CURRENT_DESKTOP");
+	config[1] = "config";
+
+	for (size_t i = 0; i < 4; i++) {
+		for (size_t j = 0; j < 2; j++) {
+			char *path = config_path(prefix[i], config[j]);
+			if (!path) {
+				continue;
+			}
+			logprint(TRACE, "config: trying config file %s", path);
+			if (file_exists(path)) {
+				return path;
+			}
+			free(path);
+		}
+	}
+
+	return NULL;
+}
+
+void init_config(char *configfile, struct xdpw_config *config) {
+	if (configfile == NULL) {
+		configfile = get_config_path();
+	}
+
+	logprint(INFO, "config: using config file %s", configfile);
+	config_parse_file(configfile, config);
+}
--- a/src/core/config.c
+++ b/src/core/config.c
@@ -57,7 +57,13 @@ static char *config_path(char *prefix, char *filename) {
 }
 
 static void config_parse_file(const char *configfile, struct xdpw_config *config) {
-	dictionary *d = iniparser_load(configfile);
+	dictionary *d = NULL;
+	if (configfile) {
+		logprint(INFO, "config: using config file %s", *configfile);
+		d = iniparser_load(configfile);
+	} else {
+		logprint(INFO, "config: no config file found");
+	}
 	if (configfile && !d) {
 		logprint(ERROR, "config: unable to load config file %s", configfile);
 	}
@@ -108,10 +114,5 @@ void init_config(const char **configfile, struct xdpw_config *config) {
 		*configfile = get_config_path();
 	}
 
-	if (*configfile) {
-		logprint(INFO, "config: using config file %s", *configfile);
-	} else {
-		logprint(INFO, "config: no config file found");
-	}
 	config_parse_file(*configfile, config);
 }
--- a/src/core/config.c
+++ b/src/core/config.c
@@ -109,7 +109,7 @@ static char *get_config_path(void) {
 	return NULL;
 }
 
-void init_config(const char **configfile, struct xdpw_config *config) {
+void init_config(char ** const configfile, struct xdpw_config *config) {
 	if (*configfile == NULL) {
 		*configfile = get_config_path();
 	}
--- a/src/core/config.c
+++ b/src/core/config.c
@@ -10,7 +10,7 @@
 #include <iniparser.h>
 
 void print_config(enum LOGLEVEL loglevel, struct xdpw_config *config) {
-	logprint(loglevel, "config: Outputname  %s", config->screencast_conf.output_name);
+	logprint(loglevel, "config: outputname  %s", config->screencast_conf.output_name);
 }
 
 // NOTE: calling finish_config won't prepare the config to be read again from config file
@@ -58,6 +58,9 @@ static char *config_path(char *prefix, char *filename) {
 
 static void config_parse_file(const char *configfile, struct xdpw_config *config) {
 	dictionary *d = iniparser_load(configfile);
+	if (configfile && !d) {
+		logprint(ERROR, "config: unable to load config file %s", configfile);
+	}
 
 	// screencast
 	getstring_from_conffile(d, "screencast:output_name", &config->screencast_conf.output_name, NULL);
@@ -100,11 +103,15 @@ static char *get_config_path(void) {
 	return NULL;
 }
 
-void init_config(const char *configfile, struct xdpw_config *config) {
-	if (configfile == NULL) {
-		configfile = get_config_path();
+void init_config(const char **configfile, struct xdpw_config *config) {
+	if (*configfile == NULL) {
+		*configfile = get_config_path();
 	}
 
-	logprint(INFO, "config: using config file %s", configfile);
-	config_parse_file(configfile, config);
+	if (*configfile) {
+		logprint(INFO, "config: using config file %s", *configfile);
+	} else {
+		logprint(INFO, "config: no config file found");
+	}
+	config_parse_file(*configfile, config);
 }
--- a/src/core/logger.c
+++ b/src/core/logger.c
@@ -5,17 +5,6 @@
 #include <string.h>
 #include <time.h>
 
-static int NUM_LEVELS = 6;
-
-static const char *loglevels[] = {
-	"QUIET",
-	"ERROR",
-	"WARN",
-	"INFO",
-	"DEBUG",
-	"TRACE"
-};
-
 static struct logger_properties logprops;
 
 void init_logger(FILE *dst, enum LOGLEVEL level) {
@@ -24,14 +13,42 @@ void init_logger(FILE *dst, enum LOGLEVEL level) {
 }
 
 enum LOGLEVEL get_loglevel(const char *level) {
-	int i;
-	for (i = 0; i < NUM_LEVELS; i++) {
-		if (!strcmp(level, loglevels[i])) {
-			return (enum LOGLEVEL) i;
-		}
+	if (strcmp(level, "QUIET") == 0) {
+		return QUIET;
+	} else if (strcmp(level, "ERROR") == 0) {
+		return ERROR;
+	} else if (strcmp(level, "WARN") == 0) {
+		return WARN;
+	} else if (strcmp(level, "INFO") == 0) {
+		return INFO;
+	} else if (strcmp(level, "DEBUG") == 0) {
+		return DEBUG;
+	} else if (strcmp(level, "TRACE") == 0) {
+		return TRACE;
 	}
+
 	fprintf(stderr, "Could not understand log level %s\n", level);
 	abort();
+}
+
+char *print_loglevel(enum LOGLEVEL loglevel) {
+	switch (loglevel) {
+	case QUIET:
+		return "QUIET";
+	case ERROR:
+		return "ERROR";
+	case WARN:
+		return "WARN";
+	case INFO:
+		return "INFO";
+	case DEBUG:
+		return "DEBUG";
+	case TRACE:
+		return "TRACE";
+	}
+	fprintf(stderr, "Could not find log level %d\n", loglevel);
+	abort();
+	return NULL;
 }
 
 void logprint(enum LOGLEVEL level, char *msg, ...) {
@@ -56,7 +73,7 @@ void logprint(enum LOGLEVEL level, char *msg, ...) {
 
 	fprintf(logprops.dst, "%s", timestr);
 	fprintf(logprops.dst, " ");
-	fprintf(logprops.dst, "[%s]", loglevels[level]);
+	fprintf(logprops.dst, "[%s]", print_loglevel(level));
 	fprintf(logprops.dst, " - ");
 
 	va_start(args, msg);
--- a/src/core/logger.c
+++ b/src/core/logger.c
@@ -31,7 +31,7 @@ enum LOGLEVEL get_loglevel(const char *level) {
 	abort();
 }
 
-char *print_loglevel(enum LOGLEVEL loglevel) {
+static const char *print_loglevel(enum LOGLEVEL loglevel) {
 	switch (loglevel) {
 	case QUIET:
 		return "QUIET";
@@ -48,7 +48,6 @@ char *print_loglevel(enum LOGLEVEL loglevel) {
 	}
 	fprintf(stderr, "Could not find log level %d\n", loglevel);
 	abort();
-	return NULL;
 }
 
 void logprint(enum LOGLEVEL level, char *msg, ...) {
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -43,7 +43,7 @@ static int handle_name_lost(sd_bus_message *m, void *userdata, sd_bus_error *ret
 int main(int argc, char *argv[]) {
 	struct xdpw_config config = {0};
 	char *configfile = NULL;
-	enum LOGLEVEL loglevel = ERROR;
+	enum LOGLEVEL loglevel = DEFAULT_LOGLEVEL;
 	bool replace = false;
 
 	static const char* shortopts = "l:o:c:rh";
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -24,6 +24,8 @@ static int xdpw_usage(FILE* stream, int rc) {
 		"                                     QUIET, ERROR, WARN, INFO, DEBUG, TRACE\n"
 		"    -o, --output=<name>              Select output to capture.\n"
 		"                                     metadata (performs no conversion).\n"
+		"    -c, --config=<config file>	      Select config file.\n"
+		"                                     (default is $XDG_CONFIG_HOME/xdg-desktop-portal-wlr/config)\n"
 		"    -r, --replace                    Replace a running instance.\n"
 		"    -h, --help                       Get help (this text).\n"
 		"\n";
@@ -39,14 +41,16 @@ static int handle_name_lost(sd_bus_message *m, void *userdata, sd_bus_error *ret
 }
 
 int main(int argc, char *argv[]) {
-	const char* output_name = NULL;
+	struct xdpw_config config = {0};
+	char *configfile = NULL;
 	enum LOGLEVEL loglevel = ERROR;
 	bool replace = false;
 
-	static const char* shortopts = "l:o:rh";
+	static const char* shortopts = "l:o:c:rh";
 	static const struct option longopts[] = {
 		{ "loglevel", required_argument, NULL, 'l' },
 		{ "output", required_argument, NULL, 'o' },
+		{ "config", required_argument, NULL, 'c' },
 		{ "replace", no_argument, NULL, 'r' },
 		{ "help", no_argument, NULL, 'h' },
 		{ NULL, 0, NULL, 0 }
@@ -62,7 +66,10 @@ int main(int argc, char *argv[]) {
 			loglevel = get_loglevel(optarg);
 			break;
 		case 'o':
-			output_name = optarg;
+			config.screencast_conf.output_name = strdup(optarg);
+			break;
+		case 'c':
+			configfile = strdup(optarg);
 			break;
 		case 'r':
 			replace = true;
@@ -75,6 +82,7 @@ int main(int argc, char *argv[]) {
 	}
 
 	init_logger(stderr, loglevel);
+	init_config(configfile, &config);
 
 	int ret = 0;
 
@@ -111,12 +119,13 @@ int main(int argc, char *argv[]) {
 		.screencast_source_types = MONITOR,
 		.screencast_cursor_modes = HIDDEN | EMBEDDED,
 		.screencast_version = XDP_CAST_PROTO_VER,
+		.config = &config,
 	};
 
 	wl_list_init(&state.xdpw_sessions);
 
 	xdpw_screenshot_init(&state);
-	ret = xdpw_screencast_init(&state, output_name);
+	ret = xdpw_screencast_init(&state);
 	if (ret < 0) {
 		logprint(ERROR, "xdpw: failed to initialize screencast");
 		goto error;
@@ -217,6 +226,8 @@ int main(int argc, char *argv[]) {
 	}
 
 	// TODO: cleanup
+	finish_config(&config);
+	free(configfile);
 
 	return EXIT_SUCCESS;
 
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -42,7 +42,7 @@ static int handle_name_lost(sd_bus_message *m, void *userdata, sd_bus_error *ret
 
 int main(int argc, char *argv[]) {
 	struct xdpw_config config = {0};
-	const char *configfile = NULL;
+	char *configfile = NULL;
 	enum LOGLEVEL loglevel = DEFAULT_LOGLEVEL;
 	bool replace = false;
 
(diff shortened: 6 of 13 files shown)


danshick:
Nice. What should the syntax be for a command to determine the output instead of a fixed output string? If the config had both, which would take precedence?

I don't think it is important to implement arbitrary output command handling now, but i think we should have a plan for the syntax and behavior.

columbarius:
> Nice. What should the syntax be for a command to determine the output instead of a fixed output string? If the config had both, which would take precedence?

I think that depends, if we want to have a set of hardcoded default choosers or not. If we have one, we need a way to hard override the chooser with a given monitor. If you could disable all choosers (like adding an empty chooser entry) it would be irrelevant, if you first skip the choosers and then take the set monitor or choose the set monitor in the first place. So i would let the hardcoded output be priority.

> I don't think it is important to implement arbitrary output command handling now, but i think we should have a plan for the syntax and behavior.

For an arbitrary output chooser, we would just need two options in the file. chooser_cmd and chooser_type. Those would override the hardcoded ones. Question is, if we want to fail screencast, if no monitor is selected, or if we still pick the "first" one?

I could rebase #59 onto this MR.

danshick:
> I could rebase #59 onto this MR.

I was thinking of landing this one first, then going back to #59 and just adding the needed config entries.

My initial thoughts for config would be

```ini
chooser_type = first | fixed | command

chooser_cmd = <string representation of command to run, should accept list of outputs on stdin and output the choice on stdout, ignored for all types but command>

chooser_fixed_value = <string representation of output string>
```

I wonder if we should define an ini section for these settings.

columbarius:
I added my thought on how to config the chooser in #59.
We could adapt it in a way that type=none would use output_name or chooser_value to select a monitor directly.

src/core/logger.c
```diff
@@ -18,6 +18,24 @@ static const char *loglevels[] = {
 
 static struct logger_properties logprops;
 
+FILE* open_logger_file(char *file_name) {
```
danshick, line 21: IMO, I don't think we should support log files. Users can redirect stderr and stdout to a file if desired. @emersion, any opinion one way or the other?

src/core/config.c
```diff
@@ -0,0 +1,141 @@
...
+};
+
+void print_config(struct xdpw_config *config) {
+	printf("Configfile: %s\n",config->conf.configfile);
+	printf("Logfile: %s\n",config->logger_conf.logfile);
+	printf("Loglevel: %d\n",config->logger_conf.loglevel);
+	printf("Output_Name: %s\n",config->screencast_conf.output_name);
+}
+
+void free_configstring(char **dest) {
```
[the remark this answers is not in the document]
columbarius: does free(pointer) set pointer to the value NULL?

src/core/logger.c
columbarius: I added this, since it was simple and gave me a way to test the path expansion stuff. Can remove it, if wanted.

src/core/config.c
```diff
@@ -0,0 +1,141 @@
...
+static const char *loglevels[] = {
+	"QUIET",
+	"ERROR",
+	"WARN",
+	"INFO",
+	"DEBUG",
+	"TRACE"
+};
+
+void print_config(struct xdpw_config *config) {
```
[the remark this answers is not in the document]
columbarius: The current problem about logging the creation of the config is, that in the config file there are informations about the logger. So we have to either initialize the logger before with some dummy value, and overwrite it later, or we cant log the config parsing and initialize the logger afterwards.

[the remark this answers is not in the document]
danshick: Makes sense, that went right over my head.

src/core/config.c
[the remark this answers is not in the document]
danshick: No, it just frees the underlying memory. Doing it manually after a free call is a good way to convert use-after-free errors to null pointer exceptions, and in rare cases, prevent a double free.

src/core/logger.c
danshick: This one is just my opinion and there may be some good reason to support file logging that I'm not thinking of, but I always expect services to log to stdout and stderr. That said, if xdpw isn't, it's probably due to a user's choice.

My two biggest concerns are around unneeded additional code surface area for bugs, and distros creating default configs that do unexpected things. Neither of those concerns seem likely.

src/core/config.c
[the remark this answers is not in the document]
columbarius: i would like to do that, since i use null pointers to detect, if a default value was set before. Maybe later some more complex things are done to the config.

src/core/logger.c
columbarius: xdpw is usually started by xdg-desktop-portal. If it writes stuff to stdout/stderr, are they catched by journald?

danshick: > xdpw is usually started by xdg-desktop-portal.

Technically xdpw is started via d-bus activation. xdp makes a d-bus call, which is associated with a particular systemd user service definition, which is automatically started. If you run `systemctl --user status xdg-desktop-portal-wlr.service`, you'll see that it is running and it's last few log lines will be abbreviated there.

Writing to a log file would break this behavior.

(All of this assumes systemd and journald. Technically this all can work without systemd. D-bus activation should still start xdpw, but I'm unfamiliar with this circumstance and haven't tested it. In either scenario, xdp wouldn't be the process parent)

columbarius: I would suggest, we just avoid this issue by removing the logfile option, until someone needs it.

columbarius:
I would also look at /etc/xdg/xdg-desktop-portal-wlr/{DESKTOP,config} before /etc/xdg-desktop-portal-wlr/... for a config to be compliant with xdg spec.
@danshick @emersion is there still something to work on, or can I squash and clean everything up for review?

danshick:
> I would also look at /etc/xdg/xdg-desktop-portal-wlr/{DESKTOP,config} before /etc/xdg-desktop-portal-wlr/... for a config to be compliant with xdg spec.

Makes sense.

> @danshick @emersion is there still something to work on, or can I squash and clean everything up for review?

Nope, looks good to me.

src/core/config.c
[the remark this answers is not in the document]
emersion: I'd prefer _not_ to do that, because it's not idiomatic C. As a reader, it requires more effort to figure out that `free_configstring` is just a wrapper around `free` with some special cases.

If you're interested in use-after-free, you can build with ASan.

src/core/config.c
```diff
@@ -0,0 +1,139 @@
...
+}
+
+static bool file_exists(const char *path) {
+	return path && access(path, R_OK) != -1;
+}
+
+static char *expand_path(char *path, bool realloc) {
+	if (path != NULL) {
+		wordexp_t p;
+		if (wordexp(path, &p, WRDE_UNDEF) == 0) {
```
emersion, line 82: This doesn't work if a directory name contains spaces.

src/core/config.c
```diff
@@ -0,0 +1,139 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+#include <wordexp.h>
+#include "iniparser.h"
+
+#define SYSCONFDIR "/etc"
+
+static const char *loglevels[] = {
```
emersion, line 16: This is duplicated elsewhere. Maybe we can just use the one from `logger.c`, e.g. by introducing a helper function?

src/core/config.c
```diff
@@ -0,0 +1,139 @@
...
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+#include <wordexp.h>
+#include "iniparser.h"
+
+#define SYSCONFDIR "/etc"
```
emersion, line 14: We need to grab that from Meson's options.

src/core/config.c
```diff
@@ -0,0 +1,139 @@
...
+
+	char *config_home = getenv("XDG_CONFIG_HOME");
+	if (!config_home || !*config_home) {
+		config_paths[0] = "$HOME/.config/xdg-desktop-portal-wlr/$XDG_CURRENT_DESKTOP";
+		config_paths[1] = "$HOME/.config/xdg-desktop-portal-wlr/config";
+	}
+
+	char *path;
+	for (size_t i = 0; i < sizeof(config_paths) / sizeof(char *); ++i) {
+		path = expand_path((char*)config_paths[i], false);
```
emersion, line 123: The cast is unnecessary.

src/core/main.c
```diff
@@ -50,10 +53,13 @@ int main(int argc, char *argv[]) {
 
 		switch (c) {
 		case 'l':
-			loglevel = get_loglevel(optarg);
+			config.logger_conf.loglevel = get_loglevel(optarg);
 			break;
 		case 'o':
-			output_name = optarg;
+			config.screencast_conf.output_name = optarg;
```
emersion, line 59: We need to `strdup` here.

src/core/config.c
```diff
@@ -0,0 +1,139 @@
...
+
+void destroy_config(struct xdpw_config *config) {
+	free_configstring(&config->conf.configfile);
+
+	// screencast
+	free_configstring(&config->screencast_conf.output_name);
+}
+
+void getstring_from_conffile(dictionary *d, char *key, char **dest, char *def) {
+	if (*dest == NULL) {
```
emersion, line 47: Instead of these nested `if`s, can return early.

src/core/config.c
```diff
@@ -0,0 +1,139 @@
...
+				*dest = strdup(c);
+			} else {
+				*dest = strdup(def);
+			}
+		}
+	}
+}
+
+void getenum_from_conffile(dictionary *d, char *key, int *dest, const char *array[] , int num_array, int def) {
+	if (*dest == 0) {
```
emersion, line 61: Ditto

include/config.h
```diff
@@ -0,0 +1,28 @@
+#ifndef CONFIG_H
+#define CONFIG_H
+
+#include "logger.h"
+
+struct config_general {
```
emersion, line 6: Do we really need that many structs? Can't we just have a single struct with two fields?

meson.build
```diff
@@ -26,6 +26,7 @@ rt = cc.find_library('rt')
 pipewire = dependency('libpipewire-0.3', version: '>= 0.3.2')
 wayland_client = dependency('wayland-client')
 wayland_protos = dependency('wayland-protocols', version: '>=1.14')
+iniparser = cc.find_library('iniparser')
```
emersion, line 29: Is there no pkg-config file for this library?

columbarius: Sadly [no](https://bugs.archlinux.org/task/45589?string=iniparser&project=1&type%5B0%5D=&sev%5B0%5D=&pri%5B0%5D=&due%5B0%5D=&reported%5B0%5D=&cat%5B0%5D=&status%5B0%5D=&percent%5B0%5D=&opened=&dev=&closed=&duedatefrom=&duedateto=&changedfrom=&changedto=&openedfrom=&openedto=&closedfrom=&closedto=).

include/config.h
columbarius: No, it's more like a personal preference, to have one struct for each portal to create some kind of namespace separation.  Currently all options could be in one struct.

src/core/config.c
columbarius: good idea

src/core/config.c
columbarius: What would be the solution to fix that? That's the way sway expands it's config path.

src/core/config.c
columbarius: good idea. wasn't myself happy with that.

src/core/main.c
columbarius: So there wasn't a strdup used before, because the output option would accidentality be used last?

src/core/config.c
columbarius: It was needed to strip the const mondifier of config_paths[], but i removed that.

meson.build
jbeich: Fails on FreeBSD (and likely other BSDs) as packages are not in the default search path of base compiler in order to avoid tainting build environment. See also https://wiki.freebsd.org/WarnerLosh/UsrLocal#Library_paths

```
$ meson setup _build
[...]
meson.build:29:0: ERROR: C shared or static library 'iniparser' not found

$ pkg info -l iniparser
iniparser-4.1:
        /usr/local/include/dictionary.h
        /usr/local/include/iniparser.h
        /usr/local/lib/libiniparser.a
        /usr/local/lib/libiniparser.so
        /usr/local/lib/libiniparser.so.1
        /usr/local/share/licenses/iniparser-4.1/LICENSE
        /usr/local/share/licenses/iniparser-4.1/MIT
        /usr/local/share/licenses/iniparser-4.1/catalog.mk
```

Maybe try the following
```suggestion
iniparser = cc.find_library('iniparser', dirs: [join_paths(get_option('prefix'), get_option('libdir'))])
```

emersion: This should ne a change you submit to meson, not to each and every project using it.

jbeich: See also https://github.com/mesonbuild/meson/issues/7746

src/core/config.c
columbarius: done

columbarius:
@emersion
I think I addressed all open points except the nested structs of the config struct. I like it more that way, but if you want, I can flatten them. Ready for a new review, if you have time.

src/core/config.c
emersion: This replaces multiple spaces with a single one, and other whitespace characters with a space.

columbarius: This should solve the problem better. Libiniparser removes the quotes when parsing strings, so no path should be quoted, which does break wordexp. I first tried a solution using $IFS to disable separating on whitespaces as indicated [here](https://linux.die.net/man/3/wordexp). But that doesn't seemed to be implemented by the gnu version of libc. So I hope this behaves well on most implementations.

src/core/config.c
```diff
@@ -0,0 +1,133 @@
...
+}
+
+static bool file_exists(const char *path) {
+	return path && access(path, R_OK) != -1;
+}
+
+static char *expand_path(char *path, bool realloc) {
+	if (path != NULL) {
+		char *expand = (char*)calloc(strlen(path)+3, sizeof(char));
+		// Quote string to preserve whitespaces
```
emersion, line 68: This breaks paths which contain quotes.

I really don't think we should be using `wordexp` here.

columbarius: can't we just use the original version, since we only apply this function to hardcoded strings and the only point of error afaics is if someone has a space or another problematic special character as SYSCONFDIR

emersion: The original version breaks when any expanded variable contains a space. I don't think that's reasonable.

columbarius: Yeah makes sense, sry I thought wordexp would only seperate strings on spaces in the to expand string and not in the expanded string and didn't tested it. 

I changed it to assemble the path directly from environment variables.

include/config.h
```diff
@@ -0,0 +1,28 @@
...
+	char *configfile;
+};
+
+struct config_logger {
+	enum LOGLEVEL loglevel;
```
emersion, line 11: I wonder, is there a use-case for setting the log level in the config file? The logging level mostly exists for debugging purposes, which means manually starting xdpw.

For instance, Sway doesn't have an option to set it from the config file.

include/config.h
```diff
@@ -0,0 +1,28 @@
+#ifndef CONFIG_H
+#define CONFIG_H
+
+#include "logger.h"
+
+struct config_general {
+	char *configfile;
```
emersion, line 7: Do we really need this in the config file struct?

include/logger.h
```diff
@@ -12,6 +14,7 @@ struct logger_properties {
 
 void init_logger(FILE *dst, enum LOGLEVEL level);
 enum LOGLEVEL get_loglevel(const char *level);
+const char** get_loglevels();
```
emersion, line 17: Style: `**` on the right

src/core/logger.c
```diff
@@ -34,6 +35,10 @@ enum LOGLEVEL get_loglevel(const char *level) {
 	abort();
 }
 
+const char** get_loglevels() {
```
emersion, line 38: Style: `**` on the right

src/core/config.c
```diff
@@ -0,0 +1,137 @@
...
+static char *config_path(char *prefix, char *filename) {
+	if (!prefix || !*prefix || !filename || !*filename) {
+		return NULL;
+	}
+
+	char *config_folder = "xdg-desktop-portal-wlr";
+
+	size_t length = 3 + strlen(prefix) + strlen(config_folder) + strlen(filename);
+	char *path = (char*)calloc(length, sizeof(char));
+	strcat(path, prefix);
```
emersion, line 75: Can we use `snprintf` here? Too easy to trip on bufefr overflows with `strcat` .

src/core/config.c
```diff
@@ -0,0 +1,137 @@
...
+
+static char *config_path(char *prefix, char *filename) {
+	if (!prefix || !*prefix || !filename || !*filename) {
+		return NULL;
+	}
+
+	char *config_folder = "xdg-desktop-portal-wlr";
+
+	size_t length = 3 + strlen(prefix) + strlen(config_folder) + strlen(filename);
+	char *path = (char*)calloc(length, sizeof(char));
```
emersion, line 74: Style: unnecessary cast

src/core/config.c
```diff
@@ -0,0 +1,137 @@
...
+
+void print_config(enum LOGLEVEL loglevel, struct xdpw_config *config) {
+	logprint(loglevel, "config: Configfile %s",config->conf.configfile);
+	logprint(loglevel, "config: Loglevel %d",config->logger_conf.loglevel);
+	logprint(loglevel, "config: Outputname  %s",config->screencast_conf.output_name);
+}
+
+// NOTE: calling destroy_config won't prepare the config to be read again from config file
+// with init_config since to pointers and other values won't be reset to NULL, or 0
+void destroy_config(struct xdpw_config *config) {
```
emersion, line 22: Style: rename to `finish_config`, since this doesn't `free` the config.

src/core/config.c
```diff
@@ -0,0 +1,137 @@
...
+
+	iniparser_freedict(d);
+	logprint(DEBUG, "config: config file parsed");
+	print_config(DEBUG, config);
+}
+
+static char *get_config_path(void) {
+	char *home = getenv("HOME");
+	char *config_home_fallback = (char*)calloc(1 + strlen(home) + strlen("/.config"), sizeof(char));
+	strcat(config_home_fallback, home);
```
emersion, line 100: Ditto

src/core/config.c
```diff
@@ -0,0 +1,137 @@
...
+	// screencast
+	free(&config->screencast_conf.output_name);
+}
+
+void getstring_from_conffile(dictionary *d, char *key, char **dest, char *def) {
+	if (*dest != NULL) {
+		return;
+	}
+	const char *c = iniparser_getstring(d, key, def);
+	if (c != NULL) {
```
emersion, line 36: Can be simplified with an early return:

```c
if (c == NULL) {
    return;
}
```

src/core/config.c
```diff
@@ -0,0 +1,137 @@
...
+void destroy_config(struct xdpw_config *config) {
+	logprint(DEBUG, "config: destroying config");
+
+	free(&config->conf.configfile);
+
+	// screencast
+	free(&config->screencast_conf.output_name);
+}
+
+void getstring_from_conffile(dictionary *d, char *key, char **dest, char *def) {
```
emersion, line 31: `def` sounds like "definition", maybe "fallback" would be a better name?

src/core/config.c
```diff
@@ -0,0 +1,137 @@
...
+		// Allow keys without value as default
+		if (strcmp(c, "") != 0) {
+			*dest = strdup(c);
+		} else {
+			*dest = strdup(def);
+		}
+	}
+}
+
+void getenum_from_conffile(dictionary *d, char *key, int *dest, const char *array[] , int num_array, int def) {
```
emersion, line 46: This sounds like unnecessary complexity here. If we want to keep the log level config option, why not have a function that converts a string to a log level?

src/core/config.c
```diff
@@ -0,0 +1,137 @@
...
+	}
+	*dest = def;
+}
+
+static bool file_exists(const char *path) {
+	return path && access(path, R_OK) != -1;
+}
+
+static char *config_path(char *prefix, char *filename) {
+	if (!prefix || !*prefix || !filename || !*filename) {
```
emersion, line 67: Style: prefer `str[0]` to `*str` for readability.

src/core/config.c
columbarius: We also need this later for the chooser type in #59 so we need a function to convert a string to an enum via a string array. So we could read the string from the config file and convert it with a seperate function, or do both at once.

include/config.h
columbarius: My idea was to put every option we have into one central config structure. If we want, we can read it from the configfile, of not, we wont.

src/core/config.c
emersion: I don't see a real difference between an array and a function that parses a string. At least, the function ensures the array items aren't mixed up.

include/config.h
columbarius: Same idea, it's one option, which will only be set by commandline parameter (or maybe environment variable). This should (can) not be read from the configfile, but it also serves as a placeholder for general settings, which might/will come later. 
Also it's nice for debugging, if you can dump the state/all settings of xdpw.

emersion: I'd rather have this as a separate variable.

src/core/config.c
columbarius: you mean the problem, that someone could change the order of the array, which indexes should correspond to the enum values?

This would mean, we need one function for every enum?

emersion: > you mean the problem, that someone could change the order of the array, which indexes should correspond to the enum values?

Yeah. Or just add a new entry out-of-order.

> This would mean, we need one function for every enum?

One function, one array, no real difference. Better keep things simple.

include/config.h
columbarius: ok also regarding the loglevel: the config should only contain configs for each portal?

emersion: What do you mean?

columbarius: i would remove the loglevel from config and the configfile setting with it's corresponding structs, so only the output_name for the screecast struct would stay

emersion: Yeah, that sounds good to me. We'll add more config options as we go.

src/core/config.c
columbarius: sure. i guess i should not replace the array with a function pointer to the new compare functions and keep the nice syntax?

include/config.h
emersion: (Side note, it's always easier to add new config options, than to remove existing ones. Removing config options means we break user configurations.)

src/core/config.c
emersion: Just keep it simple. No function pointer, just an intermediate `char *` variable that we feed to the function parsing the string.

columbarius: so sad. ok no problem.

src/core/config.c
columbarius: Can I resolv this with the new implementation via getenv?

columbarius: sry wrong thread

src/core/config.c
columbarius: Can I resolv this with the new implementation via getenv?

emersion: Yeah sure, the new implementation is completely different.

include/config.h
emersion: I don't think that's worthwhile if that's just adding code for no real purpose.

danshick: I kind of like this because starting xdpw manually can be a pain. This would change verbosity in journald if it is started via dbus activation as a systemd user service.

This would mean people running master could keep log verbosity up all the time without overriding the service file.

src/core/config.c
columbarius: Note: we totally relied on this array before

include/config.h
emersion: I'm not sure I understand the use-case here.

You almost never want DEBUG/TRACE levels when running with D-Bus activation, because that'd fill up your hard drive. When running on master, you likely don't have the version you compiled installed, otherwise it would conflict with the one from your package manager. Even if you want DEBUG/TRACE and you have master installed to /usr, then you'd need to edit your config file, restart xdpw (how?), reproduce the bug, then use journalctl to watch/collect the logs, then revert the config file change.

It sounds a lot easier to just run xdpw from the build directory, with the log level flag, redirecting the output to a file.

src/core/config.c
emersion: My main beef is the generic function for enums.

include/config.h
danshick: @emersion I think the problem is that when we ask people in issues to give us logs, they often get confused by the process of killing all instances of xdp/xdpw and starting both from the command line in such a way that the right instances are picked up and used. They're not often building from source. I've seen people suggesting overriding the systemd units just to get logs. If you're developing xdpw, running from the command line is a must, but for regular users, this is confusing.

I think your recent patch to add the `-r` flag will help a lot with this and may invalidate my concerns. Maybe, in addition to this, improved documentation would help in the FAQ or README.

Even when I'm testing a PR, I typically hack a PKGBUILD together that points to another fork so that I can rely on d-bus activation for a few days to test out the stability of the changes. It is nice to have logging up a little (not TRACE level, but DEBUG isn't *too* bad). This helps for more pervasive issues that are harder to catch other than by regular use.

The more I write this out, the more I think wanting a config file option for this is the exception and not the rule. Overriding the systemd unit may not be such a bad thing to have to do in the exceptional case.

columbarius: I think adding the -r flag will make this issue invalid, since the default loglevel is error, so the log will hopefully contain the reason of failure and for a more detailed log the installed version of xdpw can be used with --reload -l DEBUG/TRACE.
The only usecase left would be people which want reduce the loglevel to QUIET.

src/core/config.c
```diff
@@ -0,0 +1,117 @@
...
+	getstring_from_conffile(d, "screencast:output_name", &config->screencast_conf.output_name, NULL);
+
+	iniparser_freedict(d);
+	logprint(DEBUG, "config: config file parsed");
+	print_config(DEBUG, config);
+}
+
+static char *get_config_path(void) {
+	char *home = getenv("HOME");
+	char *config_home_fallback = (char*)calloc(1 + strlen(home) + strlen("/.config"), sizeof(char));
```
emersion, line 80: Style nit: unnecessary cast

src/core/config.c
```diff
@@ -0,0 +1,117 @@
...
+	// screencast
+	getstring_from_conffile(d, "screencast:output_name", &config->screencast_conf.output_name, NULL);
+
+	iniparser_freedict(d);
+	logprint(DEBUG, "config: config file parsed");
+	print_config(DEBUG, config);
+}
+
+static char *get_config_path(void) {
+	char *home = getenv("HOME");
```
emersion, line 79: This can be `const`

src/core/config.c
```diff
@@ -0,0 +1,117 @@
...
+	prefix[0] = getenv("XDG_CONFIG_HOME");
+	prefix[1] = config_home_fallback;
+	prefix[2] = SYSCONFDIR "/xdg";
+	prefix[3] = SYSCONFDIR;
+
+	char *config[2];
+	config[0] = getenv("XDG_CURRENT_DESKTOP");
+	config[1] = "config";
+
+	char *path;
```
emersion, line 93: This can be defined inside the loop.

src/core/config.c
```diff
@@ -0,0 +1,117 @@
...
+	for (size_t i = 0; i < 4; i++) {
+		for (size_t j = 0; j < 2; j++) {
+			path = config_path(prefix[i], config[j]);
+			if (path) {
```
emersion, line 97: We can avoid one indentation level with an early `continue`.

```c
if (!path) {
    continue;
}
```

src/core/config.c
```diff
@@ -0,0 +1,117 @@
...
+static char *config_path(char *prefix, char *filename) {
+	if (!prefix || !prefix[0] || !filename || !filename[0]) {
+		return NULL;
+	}
+
+	char *config_folder = "xdg-desktop-portal-wlr";
+
+	size_t length = 3 + strlen(prefix) + strlen(config_folder) + strlen(filename);
+	char *path = calloc(length, sizeof(char));
+	snprintf(path, strlen(path), "%s/%s/%s", prefix, config_folder, filename);
```
emersion, line 57: `strlen(path)` is zero, I'm surprised that works.

src/core/config.c
```diff
@@ -0,0 +1,117 @@
...
+}
+
+static char *config_path(char *prefix, char *filename) {
+	if (!prefix || !prefix[0] || !filename || !filename[0]) {
+		return NULL;
+	}
+
+	char *config_folder = "xdg-desktop-portal-wlr";
+
+	size_t length = 3 + strlen(prefix) + strlen(config_folder) + strlen(filename);
```
emersion, line 55: Style nit: I prefer to use the term "size" for allocation size, and "length" for number of characters not including the NULL terminator.

src/core/logger.c
```diff
@@ -5,35 +5,56 @@
 #include <string.h>
 #include <time.h>
 
-static int NUM_LEVELS = 6;
-
-static const char *loglevels[] = {
```
emersion, line 10: Hm, sorry, I didn't realize the array was already there.

danshick: Yeah, that was my addition a while ago to avoid using printf everywhere. I've got no reservations about improving it.

src/core/config.c
columbarius: oops ... 
Documentation (https://www.cplusplus.com/reference/cstdio/snprintf/) says the string can be of length n-1, so probably INT_MAX

src/core/logger.c
columbarius: So should we use the hardcoded one, or the array solution? I'm tending more to the array version and reverting the last commit, but in the end I don't care that much. Your choice.

danshick: The array does feel a bit more fragile than what you have now. I'll defer on this one.

columbarius:
rebased and squashed the commits for cleanup

src/core/config.c
```diff
@@ -0,0 +1,108 @@
...
+
+	iniparser_freedict(d);
+	logprint(DEBUG, "config: config file parsed");
+	print_config(DEBUG, config);
+}
+
+static char *get_config_path(void) {
+	const char *home = getenv("HOME");
+	char *config_home_fallback = calloc(1 + strlen(home) + strlen("/.config"), sizeof(char));
+	snprintf(config_home_fallback, strlen(config_home_fallback), "%s/.config", home);
```
David96, line 72: This doesn't work since strlen(config_home_fallback) is 0.

columbarius: Forgot the second snprintf. Thanks for finding!

oranenj:
As a user testing this out, my primary gripe with this PR is that it needs documentation. I had to run xdpw with tracelogging on to figure out what files it actually tries to load. :/ A short manual page would be ideal.

Also, could reasonable defaults be hardcoded into the system? Like, if there's something to choose, use slurp automatically unless configured otherwise, and if it fails, default to picking the first display? I generally prefer software to do something sane without having to configure it.

columbarius:
@oranenj Added a note about the locations at the description of this PR. Documentation will either follow in the wiki or with #67 as soon as this is merged.

I'm also preferring sane defaults and if you don't have a config file the output chooser should try 3 different choosers (I admit, we could differentiate if the screencast was declined, or if the chooser was not installed), but please feel free to suggest your idea at #59.

danshick:
@oranenj

Re: documentation, man pages are planned. I've been working on them for what feels like a decade, but I keep getting side tracked. They'll come out soon and we will update them with this PR if they land first, and they will include this PR if they land second. Also, this PR will likely get merged with the output selection PR, so both will need to be ready before either is stabilized. In the future, updating docs should be a criteria for accepted PRs, but I doubt docs will ever get written before something is ready to be merged.

I like sane defaults as well, but I'm not a big fan of hardcoding the names of other programs into a codebase. We've already got a hard dependency on grim (for screenshots), so slurp doesn't feel like a big stretch. I'm very wary of adding any others, but I might be in the minority on this one. I think we will always gracefully fall back to the first display and have a CLI flag that takes precedence over any chooser, just in case.

oranenj:
@danshick I totally agree with you on dependencies. In this case though, slurp wouldn't be as much a dependency as an extremely convenient integration, and since it's sort of part of the "wlroots ecosystem" it feels justified. If you don't feel like hardcoding it in C, one option might be to ship a "default configuration" in /usr/share somewhere that's not meant to be edited (and will be overwritten by package managers) which will be used as a last resort.

danshick:
> one option might be to ship a "default configuration" in /usr/share somewhere that's not meant to be edited (and will be overwritten by package managers) which will be used as a last resort.

I like this idea a lot actually. Also lets different distros ship a custom default config if they so choose.

What does everyone else think? @emersion @columbarius?

columbarius:
Depends... if we make slurp (with the specific commit) a hard dependency, we could ship a default config in /user/share with an slurp entry. If not, we would ship an almost empty config file such that xdpw falls back to the default path. That way, I would not ship a config, but try to create the most sane default possible and rely on the distribution to ship something in etc reflecting their choice.

oranenj:
I might be a good idea to nudge distributions towards using the defaults from upstream. Otherwise, you might end up having to support users using very different configurations just because distros happen to pick whatever they think is best at the moment of packaging if upstream has no opinion on the matter.

columbarius:
Since we are writing stuff for a library which could be used to write window managers and desktop environments i don't think there is something like a good upstream default, since we can assume very little as preinstalled. With a config file, we could only support one output chooser, while currently we have three assumptions as default, if the config has no configured chooser.
Afaics we have the options:

* Ship a upstream config with probably slurp configured and have slurp as a dependency. If slurp fails, or is not installed xdpw will deny all screencast since the chooser failed. If a specific wm/de wants to ship it's own chooser, it can ship a config file, which will override ours.
* Ship no upstream config file or an empty one and try a list of hardcoded somehow reasonable choosers (eg. slurp, wofi, bemenu, (dmenu in case of wayland support)). If nothing is installed take the first monitor. But this fallback is only reasonable in this single case. If a wm/de chooses otherwise, the same custom config as above should be true.

src/core/logger.c
```diff
@@ -24,16 +13,44 @@ void init_logger(FILE *dst, enum LOGLEVEL level) {
...
+	case INFO:
+		return "INFO";
+	case DEBUG:
+		return "DEBUG";
+	case TRACE:
+		return "TRACE";
+	}
+	fprintf(stderr, "Could not find log level %d\n", loglevel);
+	abort();
+	return NULL;
```
emersion, line 51: Nit: `abort` is noreturn, no need for this statement.

src/core/logger.c
```diff
@@ -24,16 +13,44 @@ void init_logger(FILE *dst, enum LOGLEVEL level) {
...
+		return DEBUG;
+	} else if (strcmp(level, "TRACE") == 0) {
+		return TRACE;
 	}
+
 	fprintf(stderr, "Could not understand log level %s\n", level);
 	abort();
 }
 
+char *print_loglevel(enum LOGLEVEL loglevel) {
```
emersion, line 34: This can be `const char *` to make it clear the caller doesn't own the memory.

src/core/config.c
```diff
@@ -0,0 +1,109 @@
...
+				return path;
+			}
+			free(path);
+		}
+	}
+
+	return NULL;
+}
+
+void init_config(char *configfile, struct xdpw_config *config) {
```
emersion, line 102: `configfile` can be `const`.

src/core/config.c
```diff
@@ -0,0 +1,109 @@
...
+	if (configfile == NULL) {
+		configfile = get_config_path();
```
emersion, line 104: This memory is never free'd. This wouldn't be the biggest memory leak in xdpw, that said.

emersion, line 104: Maybe we should check `configfile != NULL` at this point. `get_config_path` can return NULL.

src/core/config.c
```diff
@@ -0,0 +1,109 @@
...
+// NOTE: calling finish_config won't prepare the config to be read again from config file
+// with init_config since to pointers and other values won't be reset to NULL, or 0
+void finish_config(struct xdpw_config *config) {
+	logprint(DEBUG, "config: destroying config");
+
+	// screencast
+	free(&config->screencast_conf.output_name);
+}
+
+static void getstring_from_conffile(dictionary *d, char *key, char **dest, char *fallback) {
```
emersion, line 25: `key` and `fallback` can be `const`

src/core/config.c
```diff
@@ -0,0 +1,109 @@
...
+
+#include <dictionary.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <iniparser.h>
+
+void print_config(enum LOGLEVEL loglevel, struct xdpw_config *config) {
+	logprint(loglevel, "config: Outputname  %s",config->screencast_conf.output_name);
```
emersion, line 13: Style nit: missing space after comma

emersion (changes requested):
Overall looks almost good for merging. I think this will be my last review round.

src/core/config.c
columbarius: Shouldn't this memory be freed [here at the end of main](https://github.com/emersion/xdg-desktop-portal-wlr/pull/75/files#diff-03b3e8b6554bb8ccd539ad2e547d9ef13f80428101bdc01b4d6e9ea5f685fe7cR230)?
Well better not make that a contest ^^'

columbarius: Yeah, but afaics libiniparser could handle NULL as a path to the config file and will insert the defaults as if this file is empty. If you want to avoid this, I could write another function setting the default values manually.

src/core/config.c
columbarius: done

src/core/config.c
columbarius: done

src/core/config.c
columbarius: done

src/core/logger.c
columbarius: done + added static, since this is only used in this file

src/core/logger.c
columbarius: done

src/core/config.c
emersion: This assignment only changes the local `configfile` variable. The variable in `main` is unaffected.

src/core/config.c
```diff
@@ -0,0 +1,110 @@
...
+	char *config_folder = "xdg-desktop-portal-wlr";
+
+	size_t size = 3 + strlen(prefix) + strlen(config_folder) + strlen(filename);
+	char *path = calloc(size, sizeof(char));
+	snprintf(path, size, "%s/%s/%s", prefix, config_folder, filename);
+	return path;
+}
+
+static void config_parse_file(const char *configfile, struct xdpw_config *config) {
+	dictionary *d = iniparser_load(configfile);
```
emersion, line 60: We should check for a NULL `d` here, in case `iniparser_load` failed.

src/core/config.c
emersion: Hm, looking at the code, it doesn't seem like it?

https://github.com/ndevilla/iniparser/blob/f858275f7f307eecba84c2f5429483f9f28007f8/src/iniparser.c#L732

emersion: (IMHO just set the defaults in `main`)

columbarius: It happens here https://github.com/ndevilla/iniparser/blob/f858275f7f307eecba84c2f5429483f9f28007f8/src/iniparser.c#L419

About defaulting in main is, i can't detect if something is a default or was set via commandline parameter.

src/core/config.c
columbarius: Same here, if ```d``` is ```NULL``` iniparser_get* will just return the fallback.

emersion: OK, but if iniparser failed to read the file we want a proper error message

src/core/config.c
columbarius: damn, yeah screwed that up. Should work now.

src/core/config.c
columbarius: Added logging, but don't know how to supress the log from iniparser. We could check before and skip it when the config file is null, or just have the line in the log.

emersion: Hm, which log from iniparser?

columbarius: There is the line
iniparser: cannot open (NIL) or path to file
in stderr if iniparser_load fails.

src/core/main.c
```diff
@@ -217,6 +226,8 @@ int main(int argc, char *argv[]) {
 	}
 
 	// TODO: cleanup
+	finish_config(&config);
+	free((char *)configfile);
```
emersion, line 230: Something that we need to free can't be `const`

src/core/config.c
emersion: Maybe don't call `iniparser_load` if `configfile == NULL`?

columbarius: I moved all checks of configfile to the part where we load it with iniparser_load, so we don't have to check for the same things multiple times.

src/core/main.c
columbarius: Hope i got the const and not const declaration right this time.

emersion (approved):
Thanks!

columbarius:
Thanks for the squash and merge!

PR #82, opened 2021-02-02
[screenshot] destroy request when done

nullobsi:
destroy the request when it's finished.

- fixes issue where a single application requesting multiple screenshots does not work

### DIFF
--- a/src/screenshot/screenshot.c
+++ b/src/screenshot/screenshot.c
@@ -57,6 +57,7 @@ static int method_screenshot(sd_bus_message *msg, void *data,
 	// TODO: choose a better path
 	const char path[] = "/tmp/out.png";
 	if (!exec_screenshooter(path)) {
+        xdpw_request_destroy(req);
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


src/screenshot/screenshot.c
```diff
@@ -57,6 +57,7 @@ static int method_screenshot(sd_bus_message *msg, void *data,
 	// TODO: choose a better path
 	const char path[] = "/tmp/out.png";
 	if (!exec_screenshooter(path)) {
+        xdpw_request_destroy(req);
```
emersion, line 60: Style: we use tabs for indentation

nullobsi: oops - my editor wasn't set to tabs, let me fix it

emersion:
I wonder if that's the right thing to do. Right now `xdpw_request_destroy` is only called from the `Close` method handler, and multiple screencast sessions work properly. Are you sure it's not a bug in your client?

nullobsi:
according to the spec, Close should be called to cancel the request and close all dialogs associated with the request. I tried calling close from the client, but it never got rid of the object path. Trying to call the screenshot again led to it trying to set the vtable on an already existing object, and failing.

emersion:
What if the client calls `Close`, but at the same time xdpw destroys the object? This seems racy.

nullobsi:
Close is meant to abort the request. The spec says that after calling close, a Response signal will not be emitted. I don't know how the other desktop portals handle this, though.

nullobsi:
`xdg-desktop-portal-gtk` gets rid of the Request object path even if Close is not called.

nullobsi:
while this does seem to be a client program issue (must use token to ensure unique object path,) it doesn't make sense for such an object to persist even beyond the application closing. In fact, calling Close() does not even work-- xdg-desktop-portal destroys its proxy object before the program gets a chance to call it.

dmayle:
subscribe

danshick:
> ~~subscribe~~
Please use the notification feature rather than spamming the comments.

WhyNotHugo:
The same issue seems to exist for screencasts, a bunch of `/org/freedesktop/portal/desktop/request/1_2191/webrtcXXXXXXX` are left behind too.

Issue #81, opened 2021-01-28
screenshare freeze


tuxPT:
I don't know if this problem is from xdg-desktop-portal or from this implementation. But when I share for some time, after some minutes the screenshare freezes(the image does not change, like a photo) with discord, zoom and firefox, the only ones I use for screenshare. I currently workaround this by stopping and reshare again in these programs.

danshick:
Hmmm, I'd be curious to see some logs. Would you feel comfortable running xdp and xdpw manually (explained in the FAQ) with the `-l DEBUG` flag?

David96:
Sounds like #69 (where I still wasn't able to get meaningful WebRTC logs, sorry) to me, except that Zoom afaik doesn't use electron and therefore maybe doesn't use WebRTC which would mean there's another implementation which seems to fail.

Which brings me to a slightly off topic question: how did you manage to get screensharing to work with zoom and discord without using their web clients? Passing the WebRTCPipeWireCapturer feature flag to electron for some reason still doesn't enable screensharing in discord for me.

tuxPT:
https://gitlab.com/jamedjo/gnome-dbus-emulation-wlr

tuxPT:
but it isn't from that because firefox webrtc(jitsi, teamsweb, zoomweb, discordweb) also freezes.
Din't provide any logs beacuaseI haven't share any screen yet, but seems the bug from #69 from the simptoms

danshick:
Yeah, I don't want to muddy the waters on this issue by bringing in clients not built to use xdg-desktop-portal screensharing.

If/when you can share logs, we can take a closer look to see if this resembles the symptoms from #69.

I'd be curious if your setup shares anything in common with others who have this problem that mine does not. I've been screensharing recently for 3+ hours at a time with no issues.

junglerobba:
I am seeing the same problem, and with clients using the portal directly on sway/wlroots master and pipewire 0.3.21 on Arch Linux with a Radeon 5700XT.
On Firefox and OBS with the [PR for Wayland support](https://github.com/obsproject/obs-studio/pull/3338) and [obs-xdg-portal](https://gitlab.gnome.org/feaneron/obs-xdg-portal) the screenshare freezes almost instantly. The Chrome Flatpak seems to be working so far however.

https://gist.github.com/junglerobba/ee2ccc37df050af9eb374b40f451ec73
[...]

PR #87, opened 2021-02-11
Fix freeze on skipped frame

emersion:
When a frame is skipped, we weren't destroying the wlr frame. Since
this function kicks off the next frame capture, the stream was getting
stuck.

Disclaimer: I can't reproduce, please test.

Closes: https://github.com/emersion/xdg-desktop-portal-wlr/issues/81

### DIFF
--- a/src/screencast/wlr_screencast.c
+++ b/src/screencast/wlr_screencast.c
@@ -113,7 +113,6 @@ static struct wl_buffer *create_shm_buffer(struct xdpw_screencast_instance *cast
 
 static void wlr_frame_buffer_chparam(struct xdpw_screencast_instance *cast,
 		uint32_t format, uint32_t width, uint32_t height, uint32_t stride) {
-
 	logprint(DEBUG, "wlroots: reset buffer");
 	cast->simple_frame.width = width;
 	cast->simple_frame.height = height;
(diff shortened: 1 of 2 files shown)


src/screencast/wlr_screencast.c
```diff
@@ -190,7 +190,7 @@ static void wlr_frame_ready(void *data, struct zwlr_screencopy_frame_v1 *frame,
 
 	if (!cast->quit && !cast->err && cast->pwr_stream_state) {
 		pw_loop_signal_event(cast->ctx->state->pw_loop, cast->event);
-		return ;
+		return;
```
hhirtz, line 193: This change is completly unrelated. Can you remove it from the patch.

emersion: Split into its own commit.

David96:
I can confirm that screen sharing now continues when the `[WARN] - pipewire: out of buffers` appears which so far never worked (froze after the first warning), so I think it fixes the issue. Thanks!

hkupty:
Just tested the branch, solves the issue for me as well.

danshick:
LGTM

danshick (approved):
LGTM

emersion:
Thanks for the review, @danshick!

Issue #91, opened 2021-03-04
Segmentation fault when passing when specifying the log level and having a config file


zsolt-donca:
There is a segmentation fault when passing when specifying a log level as a command-line argument and passing INFO, DEBUG, or TRACE. It seems to work with WARN, ERROR, and QUIET.

This only happens if I happen to have a config file (which I have under `~/.config/xdg-desktop-portal-wlr/config`). There is no issue if I remove the config file.

I am building the latest master.

gdb gives me the following stack trace:
```
#0  0x00007fea25722537 in __strlen_avx2 () from /usr/lib/libc.so.6
No symbol table info available.
#1  0x00007fea2562e408 in __vfprintf_internal () from /usr/lib/libc.so.6
No symbol table info available.
#2  0x00007fea2562ebb4 in buffered_vfprintf () from /usr/lib/libc.so.6
No symbol table info available.
#3  0x000055903dd2b419 in logprint (level=INFO, msg=0x55903dd33749 "config: using config file %s") at ../src/core/logger.c:79
        args = {{gp_offset = 24, fp_offset = 48, overflow_arg_area = 0x7fff3b6c0610, reg_save_area = 0x7fff3b6c0550}}
        timestr = "2021/03/04 19:58:21\000\060\000\000\000p\005l;\377\177\000\000\260\004l;\377\177\000\000\000\032\224\326\357QH\267\004\000\000\000\377\177\000\000\060\246\322=\220U\000\000\000\000\000\000\000\000\000\000\320R\273?\220U\000\000)7\323=\220U\000\000\066\070\323=\220U", '\000' <repeats 18 times>, "[\000\000\000w\000\000\000\000\032\224\326\357QH\267\002\000\000\000\000\000\000\000\340\f\002\000\000\000\000\000\060S\273?\220U\000\000\000\000\000\000\000\000\000\000@pt%\352\177\000\000\000\004l;\377\177\000\000\000\000\000\000\000\000\000\000\065\324d%\352\177\000\000[\000\000\000w\000\000"
        t = 1614880701
        tmp = 0x7fea25785ec0 <_tmbuf>
#4  0x000055903dd2b69a in config_parse_file (configfile=0x55903fbb5330 "/home/zdonca/.config/xdg-desktop-portal-wlr/config", config=0x7fff3b6c0690) at ../src/core/config.c:62
        d = 0x0
[...]

PR #93, opened 2021-03-04
Fix segfault in config_parse_file

emersion:
configfile is a `char *`. %s needs a `char *`, so we shouldn't
dereference the pointer here.

Closes: https://github.com/emersion/xdg-desktop-portal-wlr/issues/91


danshick:
Heh, you're a little faster than me. I'll delete my branch.

danshick (approved):
LGTM

PR #95, opened 2021-03-04
Fix link to trigger-screen-shot.py

danshick:
The original snippet was deleted from the GNOME GitLab instance. I had an old copy lying around and reuploaded it as a gist for safe keeping. Link is updated to point at this gist.


emersion:
Thanks!

PR #100, opened 2021-03-15
Simplify config paths

emersion:
Drop some config paths from our lookups. See each commit for an explanation.

cc @columbarius

### DIFF
--- a/src/core/config.c
+++ b/src/core/config.c
@@ -91,16 +91,20 @@ static char *get_config_path(void) {
 	char *config_home_fallback = calloc(size_fallback, sizeof(char));
 	snprintf(config_home_fallback, size_fallback, "%s/.config", home);
 
-	const char *prefix[3];
-	prefix[0] = getenv("XDG_CONFIG_HOME");
-	prefix[1] = config_home_fallback;
-	prefix[2] = SYSCONFDIR "/xdg";
+	const char *config_home = getenv("XDG_CONFIG_HOME");
+	if (config_home == NULL || config_home[0] == '\0') {
+		config_home = config_home_fallback;
+	}
+
+	const char *prefix[2];
+	prefix[0] = config_home;
+	prefix[1] = SYSCONFDIR "/xdg";
 
 	const char *config[2];
 	config[0] = getenv("XDG_CURRENT_DESKTOP");
 	config[1] = "config";
 
-	for (size_t i = 0; i < 3; i++) {
+	for (size_t i = 0; i < 2; i++) {
 		for (size_t j = 0; j < 2; j++) {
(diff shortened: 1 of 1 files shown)


columbarius:
* About /etc/xdg-desktop-portal-wlr/ vs. /etc/xdg/xdg-desktop-portal-wlr/:
/etc/xdg/... is the correct directory following the [xdg base directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). I added /etc additionally, since it is used by sway and I don't know, if every unix system we are targeting supports the xdg spec. I have no problem with dropping /etc/xdg-... on my side.
* Constify config_path arguments:
lgtm
*  ~/.config fallback:
It's the usual pattern to have some kind of backwards compatibility, but since we didn't had a release, which uses ~/.config, despite $XDG_CONFIG_HOME is set to another location, I'm fine with dropping this redundant fallback.

emersion:
> I don't know, if every unix system we are targeting supports the xdg spec

I was wondering whether people actually used `/etc/xdg`. It seems like at least some software does it:

```
> ls /etc/xdg
autostart/  kritarc  nvim/  systemd/  termite/
```

src/core/config.c
```diff
@@ -91,16 +91,20 @@ static char *get_config_path(void) {
 	char *config_home_fallback = calloc(size_fallback, sizeof(char));
 	snprintf(config_home_fallback, size_fallback, "%s/.config", home);
 
-	const char *prefix[3];
-	prefix[0] = getenv("XDG_CONFIG_HOME");
-	prefix[1] = config_home_fallback;
-	prefix[2] = SYSCONFDIR "/xdg";
+	const char *config_home = getenv("XDG_CONFIG_HOME");
+	if (config_home == NULL) {
```
columbarius, line 95: Nit: config_home could be empty (== ""?), if I read correctly.
>  $XDG_CONFIG_HOME defines the base directory relative to which user specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

emersion: Good catch! Fixed.

columbarius:
> > I don't know, if every unix system we are targeting supports the xdg spec
> 
> I was wondering whether people actually used `/etc/xdg`. It seems like at least some software does it:
> 
> ```
> > ls /etc/xdg
> autostart/  kritarc  nvim/  systemd/  termite/
> ```

I was questioning the same, but I quite like the idea, to separate "user facing onfigs" from "server configs".

danshick:
I'm happy to follow the freedesktop spec here. LGTM.

emersion:
OK, seems like we all agree then. Let's merge this.

PR #101, opened 2021-03-15
Add exec_before and exec_after

David96:
I personally find this very useful to disable notifications during a screencast so I thought I'd just put it out there.

It's the first time I touch this codebase so I'm not sure whether I put everything in the right place.

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -6,6 +6,8 @@
 struct config_screencast {
 	char *output_name;
 	double max_fps;
+	char *exec_before;
+	char *exec_after;
 };
 
 struct xdpw_config {
--- a/src/core/config.c
+++ b/src/core/config.c
@@ -20,6 +20,8 @@ void finish_config(struct xdpw_config *config) {
 
 	// screencast
 	free(&config->screencast_conf.output_name);
+	free(&config->screencast_conf.exec_before);
+	free(&config->screencast_conf.exec_after);
 }
 
 static void getstring_from_conffile(dictionary *d,
@@ -79,6 +81,8 @@ static void config_parse_file(const char *configfile, struct xdpw_config *config
 	// screencast
 	getstring_from_conffile(d, "screencast:output_name", &config->screencast_conf.output_name, NULL);
 	getdouble_from_conffile(d, "screencast:max_fps", &config->screencast_conf.max_fps, 0);
+	getstring_from_conffile(d, "screencast:exec_before", &config->screencast_conf.exec_before, NULL);
+	getstring_from_conffile(d, "screencast:exec_after", &config->screencast_conf.exec_after, NULL);
 
 	iniparser_freedict(d);
 	logprint(DEBUG, "config: config file parsed");
--- a/src/screencast/screencast.c
+++ b/src/screencast/screencast.c
@@ -18,8 +18,36 @@
 static const char object_path[] = "/org/freedesktop/portal/desktop";
 static const char interface_name[] = "org.freedesktop.impl.portal.ScreenCast";
 
+void exec_with_shell(char *command) {
+	pid_t pid = fork();
+	if (pid < 0) {
+		perror("fork");
+	} else if (pid == 0) {
+		char *const argv[] = {
+			"sh",
+			"-c",
+			command,
+			NULL,
+		};
+		execvp("sh", argv);
+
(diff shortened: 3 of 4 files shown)


src/core/session.c
```diff
@@ -28,6 +28,12 @@ static int method_close(sd_bus_message *msg, void *data,
 
 	sd_bus_message_unref(reply);
 
+	char *exec_after = sess->screencast_instance->ctx->state->config->screencast_conf.exec_after;
```
David96, line 31: Can screencast_instance be NULL here?

danshick: Off the top of my head, I have to say it might be.

More importantly, see my note in the main conversation regarding why this might not be the right place for this logic.

danshick:
You may not want to tie your exec_before and exec_after to individual screencast instances.

Example:

- Share one screen and exec_before to disable notifications
- Share another screen starting a second instance that runs exec_before again (redundant)
- Close either session and run exec_after
- This breaks the assumption the unclosed instance made that it had disabled notifications

You probably need to introduce a way to check the existence of other instances to see if exec_before is needed in the first place, or if exec_after is free to run without disrupting another session.

All of these instances are going to be referenced here:
https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/include/screencast_common.h#L63

And this data structure is initialized here which is called shortly after the app starts:
https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/src/screencast/screencast.c#L441

You could also look at the list of xdp sessions here, each of which points to an instance (althought it may not be a unique one, but that shouldn't matter to you):
https://github.com/emersion/xdg-desktop-portal-wlr/blob/f3bd5195e4ea00e9a99b8095a4c250dbdc65de0c/include/xdpw.h#L17

src/core/session.c
```diff
@@ -28,6 +28,12 @@ static int method_close(sd_bus_message *msg, void *data,
...
+	if (exec_after) {
+		logprint(INFO, "xdpw: executing %s after screencast", exec_after);
+		system(exec_after);
```
emersion, line 34: This blocks the whole xdpw process while the command is being executed. If think it'd be preferable to fork_exec instead.

https://github.com/emersion/xdg-desktop-portal-wlr/pull/59 has some code for this (although the pipe communication is not needed here).

David96:
Alright, thanks a lot for your comments!
screenshot.c also contains code for fork & exec, is it OK if I basically duplicate that or would you prefer to introduce a helper function in something like util.c? It's slightly different though in that it waits for the exit code, which I would make dependent on a parameter then.

Can I always assume that if an instance is created a screencast will also be started right after and if there are 0 refs left that the instance will get destroyed?

emersion:
I think it's fine to duplicate, it should be just a few LoC.

David96:
Alright, I think I addressed all your comments.

danshick:
This looks correct now. @emersion may still have some feedback regarding our contribution guidelines and I usually like to daily drive a PR for a couple days to test its stability, but I think this is ready.

@emersion - are you okay with having exec_before and exec_after as config file only options or should they get accompanying CLI flags?

David96:
Forgot about the command line arguments, I added them since so far all options have command line args so it's more consistent.

David96:
Rebased on master and added the options to the newly introduced man page.

danshick (approved):
Built and tested. Works great! LGTM.

emersion:
I'm wondering whether we want to keep adding flags for config options. The existing flags are there because of historical reasons: they were introduced before config file support. I'd personally prefer to transition away from CLI flags for configuration knobs.

danshick:
@emersion - Agreed. Before 1.0, we should clean house on flags for everything except logging, replace, configfile, and help IMO. Those seem to be the ones that would be reasonably necessary from the CLI.

@David96 - Sorry for the rework, but would you mind rebasing out the CLI flag commit? I know I was the one who brought it up, but it was meant more as a question rather than a request. Apologies.

After that I'm good to merge it.

David96:
No problem at all, that's why I left them as a separate commit. I also squashed the update to the man page.

danshick:
Looks great! Thanks.

emersion (approved):
LGTM as well, thanks!

PR #103, opened 2021-03-16
pipewire: Fix building EnumFormat spa object

### LINKED ISSUE #102: Broken EnumFormat object when format is without alpha
Possible issue at lines:
https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/src/screencast/pipewire_screencast.c#L178-L179
the size of the enum doesn't match the number of elements. But I don't know if it's valid API and thus a pipewire server/library bug.

`pw-cli` shows (notice truncated output):
```
  Object: size 96, type Spa:Pod:Object:Param:Format (262147), id Spa:Enum:ParamId:EnumFormat (3)
    Prop: key Spa:Pod:Object:Param:Format:mediaType (1), flags 00000000
      Id 2        (Spa:Enum:MediaType:video)
    Prop: key Spa:Pod:Object:Param:Format:mediaSubtype (2), flags 00000000
      Id 1        (Spa:Enum:MediaSubtype:raw)
    Prop: key Spa:Pod:Object:Param:Format:Video:format (131073), flags 00000000
      Choice: type Spa:Enum:Choice:Enum, flags 00000000 24 4
        Id 7        (Spa:Enum:VideoFormat:RGBx)
        Id 7        (Spa:Enum:VideoFormat:RGBx)
```

And the screencast doesn't work. This is with:
```
wlroots master
xdg-desktop-portal-wlr-0.2.0
pipewire-0.3.23
```

Previous discussion: https://github.com/swaywm/wlroots/issues/2784

st3r4g:
Make sure that in SPA_POD_CHOICE_ENUM_Id(n_vals, ...) we always pass
exactly n_vals arguments.
Fix #102

Tested on wlroots 0.12 (format with alpha) and wlroots master (format without alpha)

### DIFF
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -166,27 +166,34 @@ void xdpw_pwr_stream_init(struct xdpw_screencast_instance *cast) {
 	enum spa_video_format format = xdpw_format_pw_from_wl_shm(cast);
 	enum spa_video_format format_without_alpha =
 		xdpw_format_pw_strip_alpha(format);
-	uint32_t n_formats = 1;
-	if (format_without_alpha != SPA_VIDEO_FORMAT_UNKNOWN) {
-		n_formats++;
-	}
 
-	const struct spa_pod *param = spa_pod_builder_add_object(&b,
-		SPA_TYPE_OBJECT_Format, SPA_PARAM_EnumFormat,
-		SPA_FORMAT_mediaType,       SPA_POD_Id(SPA_MEDIA_TYPE_video),
-		SPA_FORMAT_mediaSubtype,    SPA_POD_Id(SPA_MEDIA_SUBTYPE_raw),
-		SPA_FORMAT_VIDEO_format,    SPA_POD_CHOICE_ENUM_Id(n_formats + 1,
-			format, format, format_without_alpha),
-		SPA_FORMAT_VIDEO_size,      SPA_POD_CHOICE_RANGE_Rectangle(
+	struct spa_pod_frame f;
+	spa_pod_builder_push_object(&b, &f, SPA_TYPE_OBJECT_Format, SPA_PARAM_EnumFormat);
+	spa_pod_builder_add(&b, SPA_FORMAT_mediaType, SPA_POD_Id(SPA_MEDIA_TYPE_video), 0);
+	spa_pod_builder_add(&b, SPA_FORMAT_mediaSubtype, SPA_POD_Id(SPA_MEDIA_SUBTYPE_raw), 0);
+	if (format_without_alpha != SPA_VIDEO_FORMAT_UNKNOWN) {
+		spa_pod_builder_add(&b, SPA_FORMAT_VIDEO_format,
+			SPA_POD_CHOICE_ENUM_Id(3, format, format, format_without_alpha), 0);
+	} else {
+		spa_pod_builder_add(&b, SPA_FORMAT_VIDEO_format,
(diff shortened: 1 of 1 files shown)


src/screencast/pipewire_screencast.c
```diff
@@ -166,27 +166,32 @@ void xdpw_pwr_stream_init(struct xdpw_screencast_instance *cast) {
...
+	spa_pod_builder_add(&b, SPA_FORMAT_VIDEO_size,
+		SPA_POD_CHOICE_RANGE_Rectangle(
 			&SPA_RECTANGLE(cast->simple_frame.width, cast->simple_frame.height),
 			&SPA_RECTANGLE(1, 1),
 			&SPA_RECTANGLE(4096, 4096)),
-		// variable framerate
-		SPA_FORMAT_VIDEO_framerate, SPA_POD_Fraction(&SPA_FRACTION(0, 1)),
-		SPA_FORMAT_VIDEO_maxFramerate, SPA_POD_CHOICE_RANGE_Fraction(
+		0);
+	spa_pod_builder_add(&b, SPA_FORMAT_VIDEO_framerate, SPA_POD_Fraction(&SPA_FRACTION(0, 1)), 0);
```
emersion, line 187: This adds a `0`. Is this something we were missing too?

emersion: Oh, no, just an arg for `spa_pod_builder_add`.

st3r4g: I simply expanded the `spa_pod_builder_add_object` macro that was been used before, so no difference. You can see the macro definition in `/usr/include/spa-0.2/spa/pod/builder.h`

src/screencast/pipewire_screencast.c
```diff
@@ -166,27 +166,32 @@ void xdpw_pwr_stream_init(struct xdpw_screencast_instance *cast) {
 	enum spa_video_format format = xdpw_format_pw_from_wl_shm(cast);
 	enum spa_video_format format_without_alpha =
 		xdpw_format_pw_strip_alpha(format);
-	uint32_t n_formats = 1;
+
+	struct spa_pod_frame f;
+	spa_pod_builder_push_object(&b, &f, SPA_TYPE_OBJECT_Format, SPA_PARAM_EnumFormat);
+	spa_pod_builder_add(&b, SPA_FORMAT_mediaType, SPA_POD_Id(SPA_MEDIA_TYPE_video), 0);
```
emersion, line 172: Would it make more sense to replace these trailing `0` with `NULL`?

st3r4g: I've put `0` because that's what it is inside the macro that I expanded, but sure I can change it

st3r4g: in the pipewire [docs](https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/doc/spa/pod.md) they use `0` though

emersion: Ah, actually you were correct, 0 is better. `spa_pod_builder_add` accepts an arbitrary number of `SPA_FORMAT_*` enum values, and zero ends the list. Since enum values are integers, 0 makes more sense.

st3r4g: no problem

emersion (approved):
LGTM. Thanks for taking the time to dig up the correct way to do it.

danshick:
I'll test this shortly and as long as it builds and runs fine we can merge.

Thanks for the fix!

danshick (approved):
Builds, installs, and runs fine. LGTM

PR #105, opened 2021-04-02
Json chooser

columbarius:
Adds the chooser_type json, which calls the chooser with the output_list formated as json with a script parsing this object and calling a dmenu chooser.

Depends: #59

### DIFF
--- a/contrib/dmenu-chooser-json.sh
+++ b/contrib/dmenu-chooser-json.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+
+chooser=$1
+jq  '.monitor | .[] | .name + ": " + .make + " " + .model' | tr -d '"' | ${chooser:="wofi -d"} | cut -d : -f 1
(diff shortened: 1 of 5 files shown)


emersion:
A script can just get info from the Wayland connection. I don't think should do the work for them.

emersion:
Note to self: maybe it would be nice to add JSON output support to wayland-info for this kind of use-case.

columbarius:
> Note to self: maybe it would be nice to add JSON output support to wayland-info for this kind of use-case.

That was mainly the reason why I wrote this. 
The other one was to test, how  easy it would be to add another chooser_type and it showed that the [switch trick](https://github.com/emersion/xdg-desktop-portal-wlr/pull/105/commits/e645e83040fd09a30a2a9735718e44b4a01cf81e#diff-efc44f948f6346032394f951b0770e739779264c544dac3351249eaff6411bd4R382-R398) no longer works. 
Also this way we could do windows, when wlroots screencopy has support for that.

But feel free to close the MR if it's currently out of scope.

Issue #76, opened 2021-01-02
Monitor hotplug


columbarius:
Support hotplug in for monitors

Problems:

- [ ] Close stream graceful on when unplugging a monitor [1](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/src/screencast/screencast.c#L35) [2](https://github.com/emersion/xdg-desktop-portal-wlr/blob/dafb25fdb116bfd17fa32fb2a4a9c5f4e23e4506/src/screencast/wlr_screencast.c#L50) [3](https://github.com/emersion/xdg-desktop-portal-wlr/blob/dafb25fdb116bfd17fa32fb2a4a9c5f4e23e4506/src/screencast/wlr_screencast.c#L199)
- [ ] Added monitor is missing name (nil), or name contains garbage.

danshick:
@columbarius Will the implementation for #137 help with this as well? I know proper stream cleanup will be a necessary part of that PR, so maybe it makes sense to address this at the same time.

columbarius:
Could be. We should be able to shutdown more gracefully. The interesting thing is, if we are fast enough to signal pipewire that we are stopping the stream (or just stopping the wlroots part) before we are trying to request a screencopy for a monitor, which doesn't exist anymore. I just don't know, what each component can handle. I think we have to take a look at it after the restructure.

columbarius:
All points are done, so monitor hotplug should be working.

PR #109, opened 2021-04-15
screencast: fix monitor hotplug

columbarius:
This MR aims to fix the corrupted output list after a monitor hotplug.

Related to #76

### DIFF
--- a/src/screencast/wlr_screencast.c
+++ b/src/screencast/wlr_screencast.c
@@ -543,7 +543,13 @@ struct xdpw_wlr_output *xdpw_wlr_output_find(struct xdpw_screencast_context *ctx
 }
 
 static void wlr_remove_output(struct xdpw_wlr_output *out) {
+	free(out->name);
+	free(out->make);
+	free(out->model);
+	zxdg_output_v1_destroy(out->xdg_output);
+	wl_output_destroy(out->output);
 	wl_list_remove(&out->link);
+	free(out);
 }
 
 static void wlr_registry_handle_add(void *data, struct wl_registry *reg,
@@ -552,7 +558,7 @@ static void wlr_registry_handle_add(void *data, struct wl_registry *reg,
 
 	logprint(DEBUG, "wlroots: interface to register %s  (Version: %u)",interface, ver);
 	if (!strcmp(interface, wl_output_interface.name)) {
-		struct xdpw_wlr_output *output = malloc(sizeof(*output));
+		struct xdpw_wlr_output *output = calloc(1, sizeof(*output));
 
 		output->id = id;
 		logprint(DEBUG, "wlroots: |-- registered to interface %s (Version %u)", interface, WL_OUTPUT_VERSION);
--- a/src/screencast/wlr_screencast.c
+++ b/src/screencast/wlr_screencast.c
@@ -303,13 +303,21 @@ static void wlr_add_xdg_output_listener(struct xdpw_wlr_output *output,
 		output);
 }
 
+static void wlr_init_xdg_output(struct xdpw_screencast_context *ctx,
+		struct xdpw_wlr_output *output) {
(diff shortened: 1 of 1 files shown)


src/screencast/wlr_screencast.c
```diff
@@ -543,7 +543,11 @@ struct xdpw_wlr_output *xdpw_wlr_output_find(struct xdpw_screencast_context *ctx
 }
 
 static void wlr_remove_output(struct xdpw_wlr_output *out) {
+	free(out->name);
+	free(out->make);
+	free(out->model);
 	wl_list_remove(&out->link);
+	free(out);
 }
```
columbarius, line 551: Should out->xdg_output also be freed here? It's returned from [zxdg_output_manager_v1_get_xdg_output](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/protocols/xdg-output-unstable-v1.xml#L71).

columbarius:
The second issue it the missing monitor name. This is the result of calling the [wlr_xdg_output_listener](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/src/screencast/wlr_screencast.c#L299) only on [xdpw_wlr_screencopy_init](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/src/screencast/wlr_screencast.c#L618) and not for any output added afterwards. Maybe we should add the call to the wlr_xdg_output_listener in [wlr_output_handle_done](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/src/screencast/wlr_screencast.c#L264) but we need to screencast_context for that and I don't know, if a roundtrip is needed.

emersion:
In `wlr_registry_handle_remove`, there's no guarantee that the removed global is an output. We should check the return value of `xdpw_wlr_output_find`.

src/screencast/wlr_screencast.c
emersion: Yes, via `zxdg_output_v1_destroy`.

emersion: We should also free the `wl_output`, via `wl_output_destroy`.

emersion:
> Maybe we should add the call to the wlr_xdg_output_listener in wlr_output_handle_done but we need to screencast_context for that and I don't know, if a roundtrip is needed.

We don't need to wait for `wl_output.done`. Also, `wl_output.done` is sent when the output properties change, we don't want to re-create the `xdg_output` object when that happens.

An issue is that the order of globals is undefined, there's no guarantee that the `xdg_output_manager` will come before the `wl_output` globals.

I think we can create the `xdg_output` right after binding to the `wl_output`, assuming the `xdg_output_manager` is available. Then on init only create the `xdg_output` for each `wl_output` if it doesn't already exist.

src/screencast/wlr_screencast.c
columbarius: thanks. added it.

columbarius:
> > Maybe we should add the call to the wlr_xdg_output_listener in wlr_output_handle_done but we need to screencast_context for that and I don't know, if a roundtrip is needed.
> 
> We don't need to wait for `wl_output.done`. Also, `wl_output.done` is sent when the output properties change, we don't want to re-create the `xdg_output` object when that happens.
> 
> An issue is that the order of globals is undefined, there's no guarantee that the `xdg_output_manager` will come before the `wl_output` globals.
> 
> I think we can create the `xdg_output` right after binding to the `wl_output`, after checking that the `xdg_output_manager` is available. Then on init only create the `xdg_output` for each `wl_output` if it doesn't already exist.

Thanks for the explanation.

Current state: Output list doesn't corrupt anymore. Removing the output while screencasting will segfault xdpw. We could try to stop the corresponding screencast instances, when we remove the output, but that's probably  something for a followup mr.

emersion (approved):
LGTM, thanks!

PR #110, opened 2021-04-17
Screenshot: Implement `interactive` option

### LINKED ISSUE #45: Screenshot: Allow user to take screenshot of region
As the title says. Basically, have a way to select a region for screenshot instead of the entire screen. I'm not sure what the best mechanism for the user to select what kind of screenshot to take is though. Perhaps a simple dialog to choose between region, output, everything or cancel. Or have wait to take the screenshot until the user presses "ENTER" and use different modifiers to select a subset of the screen to capture (ex. Ctrl+click to select an output and Shift+drag to select a region)

junglerobba:
This adds reading the screenshot message options and if the `interactive` option is set, prompts the user to select a screen region with slurp before taking the screenshot.
Would close #45 

I am still very unfamiliar with C, so any comments or corrections are appreciated.
Also I wasn't quite sure about the approach, since for screencasting the chooser command is configurable. In the case of using slurp + grim their coordinate formats are compatible, so no parsing needs to be done, but I'm not sure if potential other chooser applications would be expected to output their coordinates in the same format or if the result needs to be parsed.

### DIFF
--- a/src/screenshot/screenshot.c
+++ b/src/screenshot/screenshot.c
@@ -35,17 +35,78 @@ static bool exec_screenshooter(const char *path) {
 
 	return stat == 0;
 }
+static bool exec_screenshooter_interactive(const char *path) {
+	pid_t pid = fork();
+	if (pid < 0) {
+		perror("fork");
+		return false;
+	} else if (pid == 0) {
+		char cmd[strlen(path) + 25];
+		snprintf(cmd, sizeof(cmd), "grim -g \"$(slurp)\" -- %s", path);
+		execl("/bin/sh", "/bin/sh", "-c", cmd, NULL);
+		perror("execl");
+		exit(127);
+	}
+
+	int stat;
+	if (waitpid(pid, &stat, 0) < 0) {
+		perror("waitpid");
+		return false;
+	}
+
+	return stat == 0;
+}
 
 static int method_screenshot(sd_bus_message *msg, void *data,
 		sd_bus_error *ret_error) {
 	int ret = 0;
+
+	bool interactive = false;
 
 	char *handle, *app_id, *parent_window;
 	ret = sd_bus_message_read(msg, "oss", &handle, &app_id, &parent_window);
 	if (ret < 0) {
 		return ret;
 	}
-	// TODO: read options
+
+	ret = sd_bus_message_enter_container(msg, 'a', "{sv}");
+	if (ret < 0) {
+		return ret;
+	}
+	char *key;
+	int inner_ret = 0;
+	while ((ret = sd_bus_message_enter_container(msg, 'e', "sv")) > 0) {
+		inner_ret = sd_bus_message_read(msg, "s", &key);
+		if (inner_ret < 0) {
+			return inner_ret;
+		}
+
+		if (strcmp(key, "interactive") == 0) {
+			bool mode;
+			sd_bus_message_read(msg, "v", "b", &mode);
+			logprint(DEBUG, "dbus: option interactive: %x", mode);
+			interactive = mode;
+		} else if (strcmp(key, "modal") == 0) {
+			bool modal;
+			sd_bus_message_read(msg, "v", "b", &modal);
+			logprint(DEBUG, "dbus: option modal: %x", modal);
+		} else {
+			logprint(WARN, "dbus: unknown option %s", key);
+			sd_bus_message_skip(msg, "v");
+		}
+
+		inner_ret = sd_bus_message_exit_container(msg);
+		if (inner_ret < 0) {
+			return inner_ret;
+		}
+	}
+	if (ret < 0) {
+		return ret;
+	}
+	ret = sd_bus_message_exit_container(msg);
+	if (ret < 0) {
+		return ret;
+	}
 
 	// TODO: cleanup this
 	struct xdpw_request *req =
@@ -56,7 +117,10 @@ static int method_screenshot(sd_bus_message *msg, void *data,
 
 	// TODO: choose a better path
 	const char path[] = "/tmp/out.png";
-	if (!exec_screenshooter(path)) {
+	if (interactive && !exec_screenshooter_interactive(path)) {
+		return -1;
+	}
+	if (!interactive && !exec_screenshooter(path)) {
 		return -1;
 	}
 


columbarius:
Thanks for your MR. 

What screenshot tool are you using? I tried ksnip and fireshot from flathub. It looks like only fireshot supports interactive mode and the buttons are only drawn on the focused window in sway, but that might be a sway thing. 

Other than that it worked great!

src/screenshot/screenshot.c
```diff
@@ -35,17 +35,137 @@ static bool exec_screenshooter(const char *path) {
...
+	int chooser_out[2];
+	if (pipe(chooser_out) == -1) {
+		perror("pipe chooser_out");
+		return false;
+	}
+
+	pid_t pid = spawn_chooser(chooser_out);
+	if (pid < 0) {
+		logprint(ERROR, "Failed to fork chooser");
+		return false;
```
columbarius, line 93: chooser_out[0] needs to be closed

junglerobba:
I couldn't find any tool that uses the interactive option, so I just used the [python script](https://gist.github.com/danshick/3446dac24c64ce6172eced4ac255ac3d#file-trigger-screen-shot-py-L35) that is linked in the contribution notes and added `'interactive': True` to the options.

src/screenshot/screenshot.c
```diff
@@ -35,17 +35,137 @@ static bool exec_screenshooter(const char *path) {
...
+	}
+
+	FILE *f = fdopen(chooser_out[0], "r");
+	if (f == NULL) {
+		perror("fopen pipe chooser_out");
+		return false;
```
columbarius, line 99: chooser_out[0] needs to be closed

src/screenshot/screenshot.c
```diff
@@ -54,9 +174,20 @@ static int method_screenshot(sd_bus_message *msg, void *data,
...
+		if (coords == NULL || strlen(coords) == 0) {
+			return -1;
+		}
+	}
+
 	// TODO: choose a better path
 	const char path[] = "/tmp/out.png";
-	if (!exec_screenshooter(path)) {
+	if (interactive && !exec_screenshooter_with_coordinates(path, coords)) {
+		return -1;
```
columbarius, line 188: coords has to be freed after usage

columbarius:
> I couldn't find any tool that uses the interactive option, so I just used the [python script](https://gist.github.com/danshick/3446dac24c64ce6172eced4ac255ac3d#file-trigger-screen-shot-py-L35) that is linked in the contribution notes and added `'interactive': True` to the options.

Thanks added that to my toolbox.

src/screenshot/screenshot.c
```diff
@@ -35,17 +35,140 @@ static bool exec_screenshooter(const char *path) {
...
 		return ret;
 	}
-	// TODO: read options
+
+	sd_bus_message_enter_container(msg, 'a', "{sv}");
+	if (ret < 0) {
+		return ret;
+	}
+	logprint(INFO, "dbus: handle: %s", handle);
+	logprint(INFO, "dbus: app_id: %s", app_id);
```
emersion, line 137: INFO is for end-user messages. I think this should be TRACE or DEBUG, or removed altogether.

emersion:
Can't we just `grim -g "$(slurp)"` directly?

src/screenshot/screenshot.c
```diff
@@ -35,17 +35,140 @@ static bool exec_screenshooter(const char *path) {
...
+	if (waitpid(pid, &stat, 0) < 0) {
+		perror("waitpid");
+		return false;
+	}
+
+	return stat == 0;
+}
+
+static pid_t spawn_chooser(int chooser_out[2]) {
+	pid_t pid = fork();
```
emersion, line 62: This leaves a zombie process behind, because we don't `waitpid` this new child.

junglerobba:
> Can't we just grim -g "$(slurp)" directly?
The idea was to potentially make the chooser program configurable, similar to screencasting. But I guess that might not really be necessary?

src/screenshot/screenshot.c
```diff
@@ -35,17 +35,144 @@ static bool exec_screenshooter(const char *path) {
...
+			bool modal;
+			sd_bus_message_read(msg, "v", "b", &modal);
+			logprint(INFO, "dbus: option modal: %x", modal);
+		} else {
+			logprint(WARN, "dbus: unknown option %s", key);
+			sd_bus_message_skip(msg, "v");
+		}
+
+		innerRet = sd_bus_message_exit_container(msg);
+		if (ret < 0) {
```
columbarius, line 165: shouldn't that be innerRet?

src/screenshot/screenshot.c
```diff
@@ -54,9 +181,25 @@ static int method_screenshot(sd_bus_message *msg, void *data,
 		return -ENOMEM;
 	}
 
+	char *coordinates;
+	if (interactive) {
+		coordinates = exec_coordinates_selector();
+		if (coordinates == NULL || strlen(coordinates) == 0) {
```
columbarius, line 187: Nit: use `coordinates[0] != '\0'` or `!coordinates[0]` instead of `strlen(coordinates) == 0`

src/screenshot/screenshot.c
```diff
@@ -35,17 +35,144 @@ static bool exec_screenshooter(const char *path) {
...
+	if (waitpid(pid, &stat, 0) < 0) {
+		perror("waitpid");
+		return false;
+	}
+
+	close(chooser_out[1]);
+	return stat == 0;
+}
+
+static char *exec_coordinates_selector() {
```
columbarius, line 88: Nit: use exec_ only for functions, which fork

columbarius:
> > Can't we just grim -g "$(slurp)" directly?
> 
> The idea was to potentially make the chooser program configurable, similar to screencasting. But I guess that might not really be necessary?

Currently I don't know any alternative to slurp and the only reason for a custom chooser, that i can see, would be to customize a possible chooser prompt (see #121) for e.g. language, but correct me, if i'm missing sth.

I think it would be best to just use `grim -g "$(slurp)"` directly, but maybe keep these patches around in a seperate branch (junglerobba:screenshot-interactive-old).

Sry for the late review.

cc: @emersion  Any objections?

src/screenshot/screenshot.c
```diff
@@ -35,17 +35,78 @@ static bool exec_screenshooter(const char *path) {
...
 	return stat == 0;
 }
+static bool exec_screenshooter_interactive(const char *path) {
+	pid_t pid = fork();
+	if (pid < 0) {
+		perror("fork");
+		return false;
+	} else if (pid == 0) {
+		char cmd[strlen(path) + 25];
+		sprintf(cmd, "grim -g \"$(slurp)\" -- %s", path);
```
columbarius, line 45: use snprintf(cmf, sizeof(cmf), ...)

columbarius (approved):
Thanks for your changes.

LGTM

src/screenshot/screenshot.c
```diff
@@ -35,17 +35,78 @@ static bool exec_screenshooter(const char *path) {
...
 	return stat == 0;
 }
+static bool exec_screenshooter_interactive(const char *path) {
+	pid_t pid = fork();
+	if (pid < 0) {
+		perror("fork");
+		return false;
+	} else if (pid == 0) {
+		char cmd[strlen(path) + 25];
+		snprintf(cmd, sizeof(cmd), "grim -g \"$(slurp)\" -- %s", path);
```
emersion, line 45: Note: this will fall apart if the path contains spaces or other special characters, but doesn't matter for now because the path is hardcoded.

In the future, we may want to set an env variable containing the path, and use `sh`'s variable expansion to ensure there are no quoting shenanigans.

src/screenshot/screenshot.c
```diff
@@ -35,17 +35,78 @@ static bool exec_screenshooter(const char *path) {
...
 		return ret;
 	}
-	// TODO: read options
+
+	ret = sd_bus_message_enter_container(msg, 'a', "{sv}");
+	if (ret < 0) {
+		return ret;
+	}
+	char *key;
+	int innerRet = 0;
```
emersion, line 77: Style: we use snake_case instead of camelCase.

src/screenshot/screenshot.c
```diff
@@ -35,17 +35,78 @@ static bool exec_screenshooter(const char *path) {
...
+	while ((ret = sd_bus_message_enter_container(msg, 'e', "sv")) > 0) {
+		innerRet = sd_bus_message_read(msg, "s", &key);
+		if (innerRet < 0) {
+			return innerRet;
+		}
+
+		if (strcmp(key, "interactive") == 0) {
+			bool mode;
+			sd_bus_message_read(msg, "v", "b", &mode);
+			logprint(INFO, "dbus: option interactive: %x", mode);
```
emersion, line 87: This should be DEBUG or TRACE imho.

src/screenshot/screenshot.c
```diff
@@ -35,17 +35,78 @@ static bool exec_screenshooter(const char *path) {
...
+			interactive = mode;
+		} else if (strcmp(key, "modal") == 0) {
+			bool modal;
+			sd_bus_message_read(msg, "v", "b", &modal);
+			logprint(INFO, "dbus: option modal: %x", modal);
```
emersion, line 92: Same here.

emersion (changes requested):
LGTM apart from these nits!

emersion (approved):
LGTM, thanks!

Also thanks @columbarius for the review!

PR #112, opened 2021-04-17
wlr_screencast: remove wl_display_dispatch

columbarius:
wl_display_dispatch is unnecessary and can cause hangs, if no new requests are queued


emersion (approved):
LGTM, thanks!

PR #115, opened 2021-04-19
Revert "screencast: workaround corrupted output_list"

columbarius:
This reverts commit 94f1f5d0a8e9387057b643111d9909f25395fa10.

After fixing the #109 we probably can remove that check. Will test it the next days

### DIFF
--- a/src/screencast/wlr_screencast.c
+++ b/src/screencast/wlr_screencast.c
@@ -432,8 +432,7 @@ static bool wlr_output_chooser(struct xdpw_output_chooser *chooser,
 
 	logprint(TRACE, "wlroots: output chooser %s selects output %s", chooser->cmd, name);
 	wl_list_for_each(out, output_list, link) {
-		// TODO: Replugging of outputs can result in a corrupted output_list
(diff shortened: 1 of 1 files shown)


emersion:
Do we require `xdg_output_manager` to be available? If not, then `out->name` can be NULL.

columbarius:
We require `xdg_output_manager` and the name of the output, because that's the value the chooser selects the screen by. To drop `xdg_output_manager` we would have to move to the displays `id` and patch slurp. The dmenu chooser would be useless without external information to present a human readable interface. The only path not requiring `xdg_output_manager` is probably `chooser_type=none` and `output_name=''`.

emersion:
I agree we shouldn't support the case where `xdg_output_manager` isn't available. My question was more about making sure we refuse to start if `xdg_output_manager` isn't available, ie. add a new guard in https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/src/screencast/wlr_screencast.c#L643. Can you do this?

columbarius:
Ahh that makes sense.

emersion (approved):
LGTM, thanks!

PR #118, opened 2021-04-21
Remove -o and -f options

emersion:
These can be specified in the configuration file now.


columbarius:
Nope, LGTM.
I first thought -f might be usefull to try different modes in fast succession, but an open editor and a save-shortcut will do just fine.

PR #119, opened 2021-04-22
Add wlroots to portal manifest UseIn list

emersion:
XDG_CURRENT_DESKTOP can contain a list [1]. Allow users to
set a list such as XDG_CURRENT_DESKTOP=asdf:wlroots. That way, users
can use xdpw in their non-Sway compositor without specifying
XDG_CURRENT_DESKTOP=sway and without having to edit the portal
definition.

[1]: https://github.com/flatpak/xdg-desktop-portal/blob/9a2582d035a397efbb330c53f893d5649c1fd749/src/portal-impl.c#L189

### DIFF
--- a/wlr.portal
+++ b/wlr.portal
@@ -1,4 +1,4 @@
 [portal]
 DBusName=org.freedesktop.impl.portal.desktop.wlr
 Interfaces=org.freedesktop.impl.portal.Screenshot;org.freedesktop.impl.portal.ScreenCast;
-UseIn=sway;Wayfire;river
+UseIn=wlroots;sway;Wayfire;river


danshick:
Makes sense to me. Will we still accept PRs adding individual compositors to the portal definition?

danshick (approved):

columbarius:
Then we probably have to change this https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/src/core/config.c#L119.
Should we try all values of the list, or only the first?

emersion:
Ah, yeah, correctly parsing `XDG_CURRENT_DESKTOP` is a good idea.

PR #120, opened 2021-04-22
Parse XDG_CURENT_DESKTOP as list

columbarius:
Also fixes memory leak in get_config_path

### DIFF
--- a/src/core/config.c
+++ b/src/core/config.c
@@ -119,23 +119,41 @@ static char *get_config_path(void) {
 	prefix[0] = config_home;
 	prefix[1] = SYSCONFDIR "/xdg";
 
-	const char *config[2];
-	config[0] = getenv("XDG_CURRENT_DESKTOP");
-	config[1] = "config";
+	char *xdg_current_desktop = getenv("XDG_CURRENT_DESKTOP");
+	char *config_fallback = "config";
 
+	char *config_list = NULL;
 	for (size_t i = 0; i < 2; i++) {
-		for (size_t j = 0; j < 2; j++) {
-			char *path = config_path(prefix[i], config[j]);
-			if (!path) {
-				continue;
+		if (xdg_current_desktop) {
+			config_list = strdup(xdg_current_desktop);
+			char *config = strtok(config_list, ":");
+			while (config) {
+				char *path = config_path(prefix[i], config);
+				if (!path) {
+					config = strtok(NULL, ":");
+					continue;
+				}
+				logprint(TRACE, "config: trying config file %s", path);
+				if (file_exists(path)) {
+					free(config_list);
+					free(config_home_fallback);
+					return path;
+				}
+				free(path);
+				config = strtok(NULL, ":");
 			}
-			logprint(TRACE, "config: trying config file %s", path);
-			if (file_exists(path)) {
-				free(config_home_fallback);
-				return path;
-			}
-			free(path);
+			free(config_list);
 		}
+		char *path = config_path(prefix[i], config_fallback);
+		if (!path) {
+			continue;
+		}
+		logprint(TRACE, "config: trying config file %s", path);
+		if (file_exists(path)) {
+			free(config_home_fallback);
+			return path;
+		}
+		free(path);
 	}
 
 	free(config_home_fallback);
--- a/src/core/config.c
+++ b/src/core/config.c
@@ -131,12 +131,14 @@ static char *get_config_path(void) {
 			}
 			logprint(TRACE, "config: trying config file %s", path);
 			if (file_exists(path)) {
+				free(config_home_fallback);
 				return path;
 			}
 			free(path);
 		}
 	}
 
+	free(config_home_fallback);
 	return NULL;
 }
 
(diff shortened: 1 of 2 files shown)


src/core/config.c
```diff
@@ -115,30 +115,56 @@ static char *get_config_path(void) {
...
+		xdg_current_desktop = strdup(xdg_current_desktop);
+	}
+
+	const char *config[config_num];
+	if (xdg_current_desktop != NULL) {
+		char *token = strtok(xdg_current_desktop, ":");
+		for (int i = 0; i < config_num - 1; i++) {
+			logprint(TRACE, "xdpw: token: %s (%p)", token, token);
+			config[i] = token;
+			token = strtok(NULL, ":");
```
columbarius, line 138: 1. Is strtok nilpotent (safe to call arbitrary often)?
2. Why is strtok "skipping" delimiters?
    ```
    % XDG_CURRENT_DESKTOP="::sway:wlroots" ./build/xdg-desktop-portal-wlr -l TRACE  
    2021/04/23 01:12:16 [TRACE] - xdpw: token: sway (0x602000000010)
    2021/04/23 01:12:16 [TRACE] - xdpw: token: wlroots (0x602000000015)
    2021/04/23 01:12:16 [TRACE] - xdpw: token: (null) ((nil))
    2021/04/23 01:12:16 [TRACE] - xdpw: token: (null) ((nil))
    2021/04/23 01:12:16 [TRACE] - config: trying config file ~/.config/xdg-desktop-portal-wlr/sway
    2021/04/23 01:12:16 [TRACE] - config: trying config file ~/.config/xdg-desktop-portal-wlr/wlroots
    2021/04/23 01:12:16 [TRACE] - config: trying config file ~/.config/xdg-desktop-portal-wlr/config
    2021/04/23 01:12:16 [TRACE] - config: trying config file /usr/local/etc/xdg/xdg-desktop-portal-wlr/sway
    2021/04/23 01:12:16 [TRACE] - config: trying config file /usr/local/etc/xdg/xdg-desktop-portal-wlr/wlroots
    2021/04/23 01:12:16 [TRACE] - config: trying config file /usr/local/etc/xdg/xdg-desktop-portal-wlr/config
    ```

columbarius: 1. should be according to the manpage
2. default behaviour according to the manpage

columbarius:
... Changed the title, but didn't remove the draft setting

columbarius:
Rebased and "simplified" on the cost of one inline if.

src/core/config.c
```diff
@@ -119,24 +119,46 @@ static char *get_config_path(void) {
...
+	if (xdg_current_desktop && xdg_current_desktop[0] != '\0') {
+		config_num++;
+		for (size_t i = 0; xdg_current_desktop[i] != '\0'; i++) {
+			if (xdg_current_desktop[i] == ':')
+				config_num++;
+		}
+	}
+	xdg_current_desktop = config_num > 1 ? strdup(xdg_current_desktop) : NULL;
+
+	const char *config[config_num];
```
emersion, line 134: I don't think we actually need to build this list? Can't we just call `strtok` unconditionally in the loop that calls `config_path`?

columbarius: we would have to strdup getenv("XDG_CURRENT_DESKTOP") for every prefix in the loop, but that should also work

emersion: Maybe swap the outer and inner loops to avoid this?

columbarius: then we would alternate between user and system config. I would prefere to first check all possible user configs and then the system ones.

danshick:
Apologies for the delay, I can test this tomorrow.

danshick (changes requested):
Do we want to update the documentation [here](https://github.com/emersion/xdg-desktop-portal-wlr/blame/master/xdg-desktop-portal-wlr.5.scd#L13-L20) to reflect this change? Add a line like, "$XDG_CURRNET_DESKTOP can be a colon-separated list which we will process in order".

Otherwise, this runs fine and the implementation looks good.

danshick (approved):

src/core/config.c
columbarius: I changed it to that. It looks a little bit cleaner. Choose which one you prefer and we can use that version

danshick:
I strongly prefer this approach, even though it is a little more verbose. I'll test later today or tomorrow and we can merge.

columbarius:
Squashed and ready for testing and merging

columbarius:
Please don't merge it, when you finished testing. The last commitmessage has to be reworded s/README/doc/.

danshick (approved):

danshick:
Looks good to me. Please go ahead and fixup your commit message, force push, and merge.

src/core/config.c
```diff
@@ -119,24 +119,44 @@ static char *get_config_path(void) {
 	prefix[0] = config_home;
 	prefix[1] = SYSCONFDIR "/xdg";
 
-	const char *config[2];
-	config[0] = getenv("XDG_CURRENT_DESKTOP");
-	config[1] = "config";
+	char *xdg_current_desktop = getenv("XDG_CURRENT_DESKTOP");
+	char *config_fallback = "config";
```
emersion, line 123: Nit: these two can be `const`

emersion (commented):
Feel free to merge!

src/core/config.c
columbarius: done

columbarius:
Thanks for the review!

Issue #96, opened 2021-03-07
missing iniparser include directory when compiling


flopraden:
Hi, 

When I compile the rev 323d89ead757ba43cd6e1da62ffc6074d0ed6525 (current master), I got the following error:
```
[19/20] Compiling C object xdg-desktop-portal-wlr.p/src_core_config.c.o                                                                                                                                                                                                                   
FAILED: xdg-desktop-portal-wlr.p/src_core_config.c.o                                                                                                                                                                                                                                      
ccache cc -Ixdg-desktop-portal-wlr.p -I. -I.. -I../include -Iprotocols -I/usr/local/include/pipewire-0.3 -I/usr/local/include/spa-0.2 -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Werror -std=c11 -g -Wno-missing-braces -Wno-missing-field-initi
alizers -Wno-unused-parameter -D_POSIX_C_SOURCE=200809L '-DSYSCONFDIR="/usr/local/etc"' -DHAVE_LIBSYSTEMD=1 -D_REENTRANT -MD -MQ xdg-desktop-portal-wlr.p/src_core_config.c.o -MF xdg-desktop-portal-wlr.p/src_core_config.c.o.d -o xdg-desktop-portal-wlr.p/src_core_config.c.o -c ../src
/core/config.c                                                                                                                                                                                                                                                                            
../src/core/config.c:5:10: fatal error: dictionary.h: Aucun fichier ou dossier de ce type                                                                                                                                                                                                 
    5 | #include <dictionary.h>                                                                                                                                                                                                                                                           
      |          ^~~~~~~~~~~~~~                                        
```
Adding the include directory for lib iniparser in build.ninja for core/config.c correct the problem.
Since I do not know how to make meson discover and add the correct path, I corrected manually the build.ninja

Best,

emersion:
iniparser doesn't have a pkg-config file, so we can't just use that.

On my system inparser just installs `/usr/include/dictionary.h` :S

jsmaniac:
Just installing right now, I needed to run these commands because on my ubuntu 20.04 it's in `/usr/include/iniparser/dictionary.h` and `/usr/include/iniparser/iniparser.h`

```
cd /tmp
git clone https://github.com/emersion/xdg-desktop-portal-wlr
[...]

PR #126, opened 2021-05-01
Use inih instead of libiniparser

columbarius:
Resolves: #96 

Note: inih doesn't escape quotes around a value, so this will break some configs.

#118 has made parsing much simpler.

### DIFF
--- a/src/core/config.c
+++ b/src/core/config.c
@@ -3,12 +3,11 @@
 #include "logger.h"
 #include "screencast_common.h"
 
-#include <dictionary.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
-#include <iniparser.h>
+#include <ini.h>
 
 void print_config(enum LOGLEVEL loglevel, struct xdpw_config *config) {
 	logprint(loglevel, "config: outputname  %s", config->screencast_conf.output_name);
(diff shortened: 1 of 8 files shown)


danshick:
Can you give an example of broken configs due to quoting? Does the documentation need to be updated? Can we add the escaping?

columbarius:
In the example [config](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/contrib/config.sample#L4) i used
```
chooser_cmd="slurp -f %o -o"
```
which will result now in an error, because sh can't find `"slurp -f %o -o"`.
Those quotes are not necessary and I can't remember any reason, why i added them in the first place.

So the problem is not inih not adding quotes, but libiniparser ignoring some quotes and with a sample recommending this "wrong"/unneeded quotes.

I just updated the example and the documentation with this pr.

emersion:
Yeah, I think it's OK to do this breaking change.

src/core/config.c
```diff
@@ -28,63 +30,57 @@ void finish_config(struct xdpw_config *config) {
...
+	} else if (MATCH("screencast", "chooser_cmd")) {
+		parse_string(&config->screencast_conf.chooser_cmd, value);
+	} else if (MATCH("screencast", "chooser_type")) {
 		char *chooser_type = NULL;
-		getstring_from_conffile(d, "screencast:chooser_type", &chooser_type, "default");
+		parse_string(&chooser_type, value);
 		config->screencast_conf.chooser_type = get_chooser_type(chooser_type);
 		free(chooser_type);
+	} else {
+		logprint(TRACE, "config: skipping invalid key in config file");
```
emersion, line 71: Making this an error would make it easier for users to understand why one of their config options doesn't work (e.g. typo). This is kind of unrelated to this PR though.

src/core/config.c
```diff
@@ -28,63 +30,57 @@ void finish_config(struct xdpw_config *config) {
...
-	getdouble_from_conffile(d, "screencast:max_fps", &config->screencast_conf.max_fps, 0);
-	getstring_from_conffile(d, "screencast:exec_before", &config->screencast_conf.exec_before, NULL);
-	getstring_from_conffile(d, "screencast:exec_after", &config->screencast_conf.exec_after, NULL);
-	getstring_from_conffile(d, "screencast:chooser_cmd", &config->screencast_conf.chooser_cmd, NULL);
-	if (!config->screencast_conf.chooser_type) {
+static int ini_parser_handler(void *data, const char* section, const char* name, const char* value) {
+	struct xdpw_config *config = (struct xdpw_config*)data;
+	logprint(TRACE, "config: parsing setction %s, key %s, value %s", section, name, value);
+
+	#define MATCH(s, n) strcmp(section, s) == 0 && strcmp(name, n) == 0
```
emersion, line 54: This can be a function, bringing type safety in.

emersion: If the `section` repetition is too verbose, we could have one handler per section:

```c
static int handle_ini_screencast(struct xdpw_config *config, const char *name, const char *value);
```

src/core/config.c
```diff
@@ -28,63 +30,57 @@ void finish_config(struct xdpw_config *config) {
...
-	}
-
-	// screencast
-	getstring_from_conffile(d, "screencast:output_name", &config->screencast_conf.output_name, NULL);
-	getdouble_from_conffile(d, "screencast:max_fps", &config->screencast_conf.max_fps, 0);
-	getstring_from_conffile(d, "screencast:exec_before", &config->screencast_conf.exec_before, NULL);
-	getstring_from_conffile(d, "screencast:exec_after", &config->screencast_conf.exec_after, NULL);
-	getstring_from_conffile(d, "screencast:chooser_cmd", &config->screencast_conf.chooser_cmd, NULL);
-	if (!config->screencast_conf.chooser_type) {
+static int ini_parser_handler(void *data, const char* section, const char* name, const char* value) {
```
emersion, line 50: Style: `*` on the right

src/core/config.c
emersion: About macros: https://drewdevault.com/2017/03/15/How-I-learned-to-stop-worrying-and-love-C.html

src/core/config.c
columbarius: This only works nice together with [`logprint(TRACE, "config: parsing setction %s, key %s, value %s", section, name, value);`](https://github.com/emersion/xdg-desktop-portal-wlr/pull/126/files/1b7b22ec2665cdd11c02ae96ec2df79cf6cef8f2#diff-4189891b601c97c24b3f16f7fea437d38af30d2d1c34fc8b692e7b96b153a37cR52)

emersion (approved):
LGTM, thanks!

columbarius:
Thanks for review and merging

PR #127, opened 2021-05-03
Add -r flag to slurp

emersion:
This prevents the user from selecting a custom region, and forces
them to select a full output.


columbarius (approved):
Sth. broken with the build system? Otherwise LGTM.

emersion:
Thanks for the review!

PR #131, opened 2021-05-03
readme: misc. improvements

emersion:
_See individual commits._

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,11 @@ ninja -C build install
 
 ## Running
 
-Make sure `XDG_CURRENT_DESKTOP=sway` is set.
+Make sure `XDG_CURRENT_DESKTOP` is set. Make sure `WAYLAND_DISPLAY` and
+`XDG_CURRENT_DESKTOP` are imported into D-Bus. If you're running Sway, this
+can be added to your config file:
+
+    exec dbus-update-activation-environment --systemd WAYLAND_DISPLAY XDG_CURRENT_DESKTOP=sway
 
 When correctly installed, xdg-desktop-portal should automatically invoke
 xdg-desktop-portal-wlr when needed.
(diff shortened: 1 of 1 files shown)


columbarius (approved):
LGTM

README.md
```diff
@@ -37,13 +41,15 @@ See `man 5 xdg-desktop-portal-wlr`.
...
-At the moment, some command line flags are available for testing, compatibility,
-or output selection. If you need to use one of these flags, you can provide an
-instance of xdpw using the following command:
+At the moment, some command line flags are available for development and
+testing. If you need to use one of these flags, you can start an instance of
+xdpw using the following command:
 
-```/usr/lib/xdg-desktop-portal -r & xdg-desktop-portal-wlr [OPTION...]```
+```sh
+/usr/lib/xdg-desktop-portal -r & xdg-desktop-portal-wlr -r [OPTION...]
```
danshick, line 49: We probably don't even need to bother with the first half anymore (replacing the xdp instance). That was likely only needed to ensure it picked up a new xdpw instance. Now that xdpw is replaceable, we can probably just recommend `xdg-desktop-portal-wlr -r [OPTION...]`

LMK if you guys agree.

danshick (changes requested):

README.md
columbarius: Works for me.

Is there a minimum version of xdp for this to work?

emersion: Ah right, good point.

> Is there a minimum version of xdp for this to work?

I don't think so

danshick: I think this is actually a d-bus feature, so xdp version shouldn't matter.

https://www.freedesktop.org/software/systemd/man/sd_bus_request_name.html

danshick: Eh, well, maybe xdp is doing something, but this was implemented 3 years ago...https://github.com/flatpak/xdg-desktop-portal/commit/0e67f33c7e296784019bce17b5684095791cbdfa

danshick (approved):

PR #132, opened 2021-05-03
Screenshot: Implement PickColor method

### LINKED ISSUE #129: Screenshot: Implement PickColor method
https://github.com/flatpak/xdg-desktop-portal/commit/7f693a7d3918e193488746ccc012588c487363d6

This was already mentioned in some comments, but there was no separate issue for this yet.

junglerobba:
I couldn't find any application to test this method, so i just used the usual python script and modified it a bit: https://gist.github.com/junglerobba/e3985bc66a025aed6beb5edfee1ed900

Not sure if the whole ppm parsing is really necessary, but I felt it was safer to do so.

closes #129

### DIFF
--- a/include/screenshot.h
+++ b/include/screenshot.h
@@ -1,0 +1,10 @@
+#ifndef SCREENSHOT_H
+#define SCREENSHOT_H
+
+
+struct xdpw_ppm_pixel {
+	int max_color_value;
+	unsigned char red, green, blue;
+};
+
+#endif
--- a/src/screenshot/screenshot.c
+++ b/src/screenshot/screenshot.c
@@ -5,6 +5,7 @@
 #include <sys/wait.h>
 #include <unistd.h>
 #include "xdpw.h"
+#include "screenshot.h"
 
 static const char object_path[] = "/org/freedesktop/portal/desktop";
 static const char interface_name[] = "org.freedesktop.impl.portal.Screenshot";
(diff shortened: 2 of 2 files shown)


danshick:
Very cool, thanks!

danshick:
Does this tool work to test it? This issue seems to imply it can *only* use the portal color picker method, which works in our advantage.

https://gitlab.gnome.org/World/gcolor3/-/issues/124

emersion:
IIRC Gimp might be able to use it as well, but not sure if support is merged already.

junglerobba:
> https://gitlab.gnome.org/World/gcolor3/-/issues/124

Just tested with the flatpak, and it works just fine.

include/screenshot.h
```diff
@@ -0,0 +1,15 @@
...
+
+
+struct xdpw_ppm_pixel {
+	unsigned char red, green, blue;
+};
+
+struct xdpw_ppm_image {
+	unsigned int width, height;
+	int max_color_value;
+	struct xdpw_ppm_pixel *data;
```
emersion, line 12: In practice we'll only need a single pixel, maybe we can simplify all of this by just having `red, green, blue` directly embedded here?

src/screenshot/screenshot.c
```diff
@@ -148,9 +149,151 @@ static int method_screenshot(sd_bus_message *msg, void *data,
...
+	if (!fgets(buf, sizeof(buf), f)) {
+		perror("could not read image format");
+		goto error_img;
+	}
+
+	if (buf[0] != 'P' || buf[1] != '6') {
+		logprint(WARN, "Invalid image format");
+		goto error_img;
+	}
+	if (fscanf(f, "%d %d\n", &img->width, &img->height) != 2) {
```
emersion, line 211: Maybe we can just `strcmp` this with "P6 1 1\n"?

danshick:
I also just tested it with gcolor3 without flatpak. It didn't function properly before this change and it does function correctly now, so I'm calling that a success.

include/screenshot.h
junglerobba: Sure, but then we don't really need width/height either, especially if we exit out if those are not 1?

emersion: Yup! We can get rid of those as well.

I don't think there'll be a situation where we'll want to parse more than one pixel worth of PPM data.

junglerobba:
Tried to address all current feedback. I'm not sure yet if I'm quite happy with the ppm parsing code, but it seems to be working fine at least

src/screenshot/screenshot.c
```diff
@@ -148,9 +149,149 @@ static int method_screenshot(sd_bus_message *msg, void *data,
...
+		goto error_img;
+	}
+
+	img->red = pixel[0];
+	img->green = pixel[1];
+	img->blue = pixel[2];
+
+	free(format);
+	fclose(f);
+	close(chooser_out[0]);
```
emersion, line 230: `fclose` already takes care of closing the FD, I think

emersion (changes requested):
LGTM apart from this one comment!

src/screenshot/screenshot.c
```diff
@@ -148,9 +149,149 @@ static int method_screenshot(sd_bus_message *msg, void *data,
...
+
+	return img;
+
+error_img:
+	logprint(WARN, "Invalid image format or size");
+	free(format);
+	free(img);
+	fclose(f);
+	close(chooser_out[0]);
```
emersion, line 239: (Same here)

src/screenshot/screenshot.c
```diff
@@ -148,9 +149,149 @@ static int method_screenshot(sd_bus_message *msg, void *data,
...
+	}
+
+	FILE *f = fdopen(chooser_out[0], "r");
+	if (f == NULL) {
+		perror("fopen pipe chooser_out");
+		close(chooser_out[0]);
+		return NULL;
+	}
+
+	struct xdpw_ppm_pixel *img = malloc(sizeof(struct xdpw_ppm_pixel));
```
emersion, line 199: To simplify the code, we can take a `struct xdpw_ppm_pixel *` as an argument of this function, and fill its fields.

junglerobba: Just to clarify, then the code calling this function would look like this?
```
struct xdpw_ppm_pixel *img = malloc(sizeof(struct xdpw_ppm_pixel));
if (!exec_color_picker(img)) {
	free(img);
	return -1;
}
```

emersion: No need to allocate on the heap:

```c
struct xdpw_ppm_pixel img = {0};
if (!exec_color_picker(&img)) {
    return -1;
}
```

emersion (approved):
LGTM, thanks!

PR #136, opened 2021-05-07
screencast: improve cleanup on error in xdpw_screencast_init()

### LINKED ISSUE #134: Uninitialized list in error path
Hello,

I've just started playing around with xdg-desktop-portal-wlr, and I'm encountering a segfault here:
https://github.com/emersion/xdg-desktop-portal-wlr/blob/efcbcb60aa3bc46b0e290d5f6627783fbb534dcd/src/screencast/wlr_screencast.c#L667

My analysis shows that xdpw_wlr_screencopy_init() is never called for me, hence the two lists are not initialized.

Inspecting the coredump confirms this:
```
gdb) p *ctx
$1 = {state = 0x7ffe0eb63610, pwr_context = 0x13232a0, core = 0x0, output_list = {prev = 0x0, next = 0x0}, registry = 0x0, screencopy_manager = 0x0, 
  xdg_output_manager = 0x0, shm = 0x0, screencast_instances = {prev = 0x0, next = 0x0}}
```

So this goto statement needs to be fixed:
https://github.com/emersion/xdg-desktop-portal-wlr/blob/efcbcb60aa3bc46b0e290d5f6627783fbb534dcd/src/screencast/screencast.c#L477

With best wishes,
Tobias

tobiasjakobi:
My idea how to fix #134. Probably needs review, and is only compile-tested atm.

### DIFF
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -230,6 +230,10 @@ int xdpw_pwr_context_create(struct xdpw_state *state) {
 }
 
 void xdpw_pwr_stream_destroy(struct xdpw_screencast_instance *cast) {
+	if (!cast->stream) {
+		return;
+	}
+
 	logprint(DEBUG, "pipewire: destroying stream");
 	pw_stream_flush(cast->stream, false);
 	pw_stream_disconnect(cast->stream);
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -136,7 +136,7 @@ static const struct pw_stream_events pwr_stream_events = {
 	.param_changed = pwr_handle_stream_param_changed,
 };
 
-void xdpw_pwr_stream_init(struct xdpw_screencast_instance *cast) {
+void xdpw_pwr_stream_create(struct xdpw_screencast_instance *cast) {
 	struct xdpw_screencast_context *ctx = cast->ctx;
 	struct xdpw_state *state = ctx->state;
 
@@ -206,7 +206,7 @@ void xdpw_pwr_stream_init(struct xdpw_screencast_instance *cast) {
 		&param, 1);
 }
 
-int xdpw_pwr_core_connect(struct xdpw_state *state) {
+int xdpw_pwr_context_create(struct xdpw_state *state) {
 	struct xdpw_screencast_context *ctx = &state->screencast;
 
 	logprint(DEBUG, "pipewire: establishing connection to core");
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -229,6 +229,22 @@ int xdpw_pwr_context_create(struct xdpw_state *state) {
 	return 0;
 }
 
+void xdpw_pwr_context_destroy(struct xdpw_state *state) {
+	struct xdpw_screencast_context *ctx = &state->screencast;
+
+	logprint(DEBUG, "pipewire: disconnecting fom core");
+
+	if (ctx->core) {
+		pw_core_disconnect(ctx->core);
+		ctx->core = NULL;
+	}
+
+	if (ctx->pwr_context) {
+		pw_context_destroy(ctx->pwr_context);
+		ctx->pwr_context = NULL;
+	}
+}
+
 void xdpw_pwr_stream_destroy(struct xdpw_screencast_instance *cast) {
 	if (!cast->stream) {
 		return;
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -206,6 +206,18 @@ void xdpw_pwr_stream_create(struct xdpw_screencast_instance *cast) {
 		&param, 1);
 }
 
+void xdpw_pwr_stream_destroy(struct xdpw_screencast_instance *cast) {
+	if (!cast->stream) {
+		return;
+	}
+
+	logprint(DEBUG, "pipewire: destroying stream");
+	pw_stream_flush(cast->stream, false);
+	pw_stream_disconnect(cast->stream);
+	pw_stream_destroy(cast->stream);
+	cast->stream = NULL;
+}
+
 int xdpw_pwr_context_create(struct xdpw_state *state) {
 	struct xdpw_screencast_context *ctx = &state->screencast;
 
@@ -243,16 +255,4 @@ void xdpw_pwr_context_destroy(struct xdpw_state *state) {
 		pw_context_destroy(ctx->pwr_context);
 		ctx->pwr_context = NULL;
 	}
-}
-
-void xdpw_pwr_stream_destroy(struct xdpw_screencast_instance *cast) {
-	if (!cast->stream) {
-		return;
-	}
-
-	logprint(DEBUG, "pipewire: destroying stream");
-	pw_stream_flush(cast->stream, false);
-	pw_stream_disconnect(cast->stream);
-	pw_stream_destroy(cast->stream);
-	cast->stream = NULL;
 }
--- a/src/screencast/screencast.c
+++ b/src/screencast/screencast.c
@@ -134,7 +134,7 @@ static int start_screencast(struct xdpw_screencast_instance *cast) {
 	wl_display_dispatch(cast->ctx->state->wl_display);
 	wl_display_roundtrip(cast->ctx->state->wl_display);
 
-	xdpw_pwr_stream_init(cast);
+	xdpw_pwr_stream_create(cast);
 
 	cast->initialized = true;
 	return 0;
@@ -472,7 +472,7 @@ int xdpw_screencast_init(struct xdpw_state *state) {
 	state->screencast.state = state;
 
 	int err;
-	err = xdpw_pwr_core_connect(state);
+	err = xdpw_pwr_context_create(state);
 	if (err) {
 		goto end;
 	}
--- a/src/screencast/screencast.c
+++ b/src/screencast/screencast.c
@@ -474,19 +474,22 @@ int xdpw_screencast_init(struct xdpw_state *state) {
 	int err;
 	err = xdpw_pwr_context_create(state);
 	if (err) {
-		goto end;
+		goto fail_pipewire;
 	}
 
 	err = xdpw_wlr_screencopy_init(state);
 	if (err) {
-		goto end;
+		goto fail_screencopy;
 	}
 
 	return sd_bus_add_object_vtable(state->bus, &slot, object_path, interface_name,
 		screencast_vtable, state);
 
-end:
-	// TODO: clean up pipewire
+fail_screencopy:
 	xdpw_wlr_screencopy_finish(&state->screencast);
+
+fail_pipewire:
+	xdpw_pwr_context_destroy(state);
+
 	return err;
 }
--- a/include/pipewire_screencast.h
+++ b/include/pipewire_screencast.h
@@ -6,8 +6,8 @@
 #define XDPW_PWR_BUFFERS 1
 #define XDPW_PWR_ALIGN 16
 
-void xdpw_pwr_stream_init(struct xdpw_screencast_instance *cast);
-int xdpw_pwr_core_connect(struct xdpw_state *state);
+void xdpw_pwr_stream_create(struct xdpw_screencast_instance *cast);
 void xdpw_pwr_stream_destroy(struct xdpw_screencast_instance *cast);
+int xdpw_pwr_context_create(struct xdpw_state *state);
 
 #endif
--- a/include/pipewire_screencast.h
+++ b/include/pipewire_screencast.h
@@ -9,5 +9,6 @@
 void xdpw_pwr_stream_create(struct xdpw_screencast_instance *cast);
 void xdpw_pwr_stream_destroy(struct xdpw_screencast_instance *cast);
 int xdpw_pwr_context_create(struct xdpw_state *state);
+void xdpw_pwr_context_destroy(struct xdpw_state *state);
 
 #endif


danshick:
This is a good start. I would consider calling `xdpw_pwr_stream_destroy` on all streams before you destroy the context and disconnect from the core. You'll have to iterate through `ctx->screencast_instances` to derive the arguments to `xdpw_pwr_stream_destroy`, and while you're doing so, you probably want to set their `quit` flags to true to make sure they get cleaned up on the wlr side too if necessary. `xdpw_pwr_stream_destroy` should probably have some nullish checking of `cast->stream` before disconnecting and destroying too.

I would need to take a lot more time to make sure that that approach covers all bases, but I think that would do a pretty good job of cleaning up everything downstream of the pipewire context. I'd be curious to hear from @columbarius as he's probably more knowledgeable about the pipewire API than I am at this point.

*just nitpicking below this point*

I would also point you to [our contribution guidelines](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/CONTRIBUTING.md). As it says, we closely follow sway/wlroots contributing guidelines which discusses [construction and destruction function naming](https://github.com/swaywm/wlroots/blob/master/CONTRIBUTING.md#constructiondestruction-functions).

This is all a long way of saying that we probably want to rename `xdpw_pwr_core_connect` to `xdpw_pwr_context_create` and name your function `xdpw_pwr_context_destroy`. Frankly, we may also want to rename `xdpw_pwr_stream_init` to `xdpw_pwr_stream_create` to be more consistent.

@emersion If we have a void method that takes a struct as an argument, malloc's a thing, but returns nothing and just sets a pointer inside the struct argument to the newly malloc'd thing, would you call that "init/finish" or "create/destroy"? I'm leaning towards the latter because mallocing and freeing is happening in the function and not at the call site, but it isn't completely conformant to either convention in the sway/wlroots contribution guidelines. I guess most important is that we remain consistent.

tobiasjakobi:
OK, let's wait for some additional feedback then! :-)

danshick:
Thanks for working on this BTW! We always appreciate PRs.

tobiasjakobi:
I just updated the merge request. I've split the commit into two parts, first doing the function naming fixup, then the actual fix. The quit flag already seems to be set in xdpw_wlr_screencopy_finish(), so I've just added the stream destroy there. Still need to add the the NULL ptr check in xdpw_pwr_stream_destroy().

@danshick Am I moving into the right direction? :)

src/screencast/screencast.c
```diff
@@ -472,21 +472,25 @@ int xdpw_screencast_init(struct xdpw_state *state) {
 	state->screencast.state = state;
 
 	int err;
-	err = xdpw_pwr_core_connect(state);
+	err = xdpw_pwr_context_create(state);
 	if (err) {
-		goto end;
+		goto fail_context;
```
danshick, line 477: I might call this `fail_pipewire`.

src/screencast/screencast.c
```diff
@@ -472,21 +472,25 @@ int xdpw_screencast_init(struct xdpw_state *state) {
...
-		goto end;
+		goto fail_screencopy;
 	}
 
 	return sd_bus_add_object_vtable(state->bus, &slot, object_path, interface_name,
 		screencast_vtable, state);
 
-end:
+fail_screencopy:
 	// TODO: clean up pipewire
```
danshick, line 489: You can remove this TODO.

danshick (changes requested):
Just a few quick notes looking at this on my phone. Looks good overall. I should have time to build and test this on Tuesday.

src/screencast/screencast.c
tobiasjakobi: Fixed!

src/screencast/screencast.c
tobiasjakobi: Fixed!

tobiasjakobi:
New version pushed, also adds the missing NULL pointer check.

columbarius:
At first glance this looks good. I want to test later how this works with the replace flag. I noticed that before the replace flag there were dead pipewirenodes left when stopping xdpw. So maybe we want to add a xdpw_screencast_destroy calling the error path of init on shutdown or replace, but I need to take a look where exactly.

src/screencast/wlr_screencast.c
```diff
@@ -673,6 +673,7 @@ void xdpw_wlr_screencopy_finish(struct xdpw_screencast_context *ctx) {
 	struct xdpw_screencast_instance *cast, *tmp_c;
 	wl_list_for_each_safe(cast, tmp_c, &ctx->screencast_instances, link) {
 		cast->quit = true;
+		xdpw_pwr_stream_destroy(cast);
```
columbarius, line 676: This shouldn't be necessary, because destroying the stream is done with [xdpw_screencast_instance_destroy](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/src/screencast/screencast.c#L76) and xdpw_wlr_screencopy_init doesn't touch pipewire, so xdpw_wlr_screencopy_finish shouldn't. 

Thing is xdpw_screencast_instance_destroy is only called from [xdpw_wlr_frame_free](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/src/screencast/wlr_screencast.c#L38) which needs a roundtrip through the screencast loop. The pipewire stream is created after the wlroots buffer loop is started, so xdpw_wlr_frame_free should be reached.

src/screencast/pipewire_screencast.c
```diff
@@ -229,7 +229,27 @@ int xdpw_pwr_core_connect(struct xdpw_state *state) {
...
+		ctx->core = NULL;
+	}
+
+	if (ctx->pwr_context) {
+		pw_context_destroy(ctx->pwr_context);
+		ctx->pwr_context = NULL;
+	}
+}
+
 void xdpw_pwr_stream_destroy(struct xdpw_screencast_instance *cast) {
```
columbarius, line 248: Nit: Could you please move this function up, so we have the same order as in the header?

columbarius (changes requested):

columbarius:
The dead pipewire nodes are gone, which is nice, but when we replace xdpw with the -r flag pipewire clients can SEGFAULT, when streaming, which is not nice. We probably need a xdpw_screencast_finish, which goes the error path of xdpw_screencast_init and does quit all open screencast instances, but this can be done in another MR.

danshick:
> You'll have to iterate through `ctx->screencast_instances` to derive the arguments to `xdpw_pwr_stream_destroy`, and while you're doing so, you probably want to set their `quit` flags to true to make sure they get cleaned up on the wlr side too if necessary.

From above, I thought that'd be the case, but...

> The quit flag already seems to be set in `xdpw_wlr_screencopy_finish()`, so I've just added the stream destroy there.

Does the `-r` flag logic not call screencopy finish?

columbarius:
> Does the `-r` flag logic not call screencopy finish?

It should just jump from https://github.com/tobiasjakobi/xdg-desktop-portal-wlr/blob/fixes/src/core/main.c#L204 to https://github.com/tobiasjakobi/xdg-desktop-portal-wlr/blob/fixes/src/core/main.c#L263

columbarius:
> Edit: But per your review comment, setting the quit flag should be enough to allow instance destroy to do its thing.

We would just have to wait for instance destroy to close the pipewire streams before destroying the pw context, etc.

tobiasjakobi:
> > We would just have to wait for instance destroy to close the pipewire streams before destroying the pw context, etc.
> 
> Of course, thanks. I guess we could use a timer to "reschedule" the pipewire cleanup if streams are still hanging out there, or we could just go and do it explicitly before destroying the context like I had originally thought.

Uhm, so what does that mean for me precisely?

src/screencast/pipewire_screencast.c
```diff
@@ -229,7 +229,27 @@ int xdpw_pwr_core_connect(struct xdpw_state *state) {
 	return 0;
 }
 
+void xdpw_pwr_context_destroy(struct xdpw_state *state) {
+	struct xdpw_screencast_context *ctx = &state->screencast;
+
```
danshick, line 234: If we've gotten here, and there are still streams present, we either need to asynchronously defer execution of the remainder of this function until such a time when all streams have been terminated and cleaned up (likely due to the quit or err flags being set), or we need to manually iterate through all streams and destroy them synchronously.

The former approach would use a timer (some example of this in the commits related to the max fps PR). The latter approach would involve iterating through all streams and calling stream destroy for each (and setting their quit flags so that we don't reschedule them).

danshick:
> Uhm, so what does that mean for me precisely?

I just left you a review comment that explains the necessary next steps. If you want to take a stab at implementing it, we'll do our best to guide you with advice and code reviews. If not, one of us can complete this PR when we have time to do so.

danshick (changes requested):

src/screencast/pipewire_screencast.c
tobiasjakobi: Hmm, I can't really follow here. As far as I can see xdpw_pwr_context_destroy() is only called from xdpw_screencast_init() during the error path.

If that happens we either come from goto fail_pipewire, i.e. there shouldn't be any stream around anyway. Or we come from goto fail_screencopy, in which case xdpw_wlr_screencopy_finish() is called, and then xdpw_pwr_context_destroy().

In fact also xdpw_wlr_screencopy_finish() is only ever called from xdpw_screencast_init(), I don't see it referenced anywhere else.

Now you tell me to iterate over the screencast instances in xdpw_pwr_context_destroy() and clean them up. But that is precisely what is done in xdpw_wlr_screencopy_finish().

Maybe I'm missing the big picture here, but to me this just looks like moving code around, without actually changing the order of how things are done.

danshick: I haven't had the time to actually test this or read anything beyond the diffs. I'll have a chance in the next couple weeks to give you a much more thorough review. I am making some assumptions based on @columbarius's comment, that the `-r` flag that does instance replacements hits this codepath when streams are active.

> The dead pipewire nodes are gone, which is nice, but when we replace xdpw with the -r flag pipewire clients can SEGFAULT, when streaming, which is not nice. We probably need a xdpw_screencast_finish, which goes the error path of xdpw_screencast_init and does quit all open screencast instances.

> This xdpw_screencast_finish should wait until all screencopy loops have finished (and destroyed the pipewire stream) and then destroy the context.

There is a race condition somewhere (that has always been an issue, just covered up by the fact that we've been leaving dead pipewire sessions around).

I can give you much more specific feedback later this week when I have time to read through the existing code again.

Thanks for your patience. We appreciate you working with us on this.

danshick: Okay @tobiasjakobi, I've taken the time to understand what @columbarius was saying. Unrelated to your PR, we have an error condition [here](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/src/core/main.c#L263) that can occur with active streams hanging around. We would want to call this context_destroy function there as well.

This is more complicated than I first thought, because we are also outside of our event loop here, so using our timer setup isn't plausible. I think the right thing to do may just be to check if there are any lingering stream objects, and skip the destroy calls and return an error code if there are any. This would allow us to worry about how to use this function in a place where active streams *might* be present.

@columbarius I think our error label in `core/main.c` needs to be a second "simplified" event loop that can continue processing streams to allow us a chance to clean things up, but I don't want to go down that path on this PR. Do you think making `context_destroy` return int instead of void and checking for active streams first would be sufficient to accept this PR? I'd then propose we raise an issue about this and go back in a second PR and implement that improved exit condition.

columbarius: I think we should simplify this commit to just adding context_destroy and handle errors occurring in context_create.

Since with #137 I will restructure most of the screencast_instance stuff to rely more on pipewire, which has well defined signals for buffer creation and destruction. With that we don't have to wait until some loop reached an arbitrary point at which the cleanup is finished. I don't know, how much i have to touch, so i wouldn't spend to much time with that tbh.

danshick: @columbarius In that case, I'm happy to close this review here and merge. Agree?

danshick (approved):

danshick:
@tobiasjakobi I think if you address @columbarius's two review items, we can merge this as is.

columbarius:
@danshick yeah, i'm fine with that after the two review items.

@tobiasjakobi Thanks for your work!

tobiasjakobi:
Hi, thanks for all the feedback. I pushed a new version, which should cover the remaining review items. The only thing that bothers me now, is that the "protect against NULL stream" commit isn't really needed anymore. At least for now.

danshick:
> The only thing that bothers me now, is that the "protect against NULL stream" commit isn't really needed anymore.

Even if unnecessary, I will never complain about making a function idempotent. Thanks for this work. It will prevent some segfaults in the short term and help us to build better support for instance replacement.

I'll start daily driving your branch to test it out and then we can get it merged.

src/screencast/screencast.c
```diff
@@ -472,21 +472,24 @@ int xdpw_screencast_init(struct xdpw_state *state) {
...
+		goto fail_screencopy;
 	}
 
 	return sd_bus_add_object_vtable(state->bus, &slot, object_path, interface_name,
 		screencast_vtable, state);
 
-end:
-	// TODO: clean up pipewire
+fail_screencopy:
 	xdpw_wlr_screencopy_finish(&state->screencast);
```
danshick, line 489: Nit: can you indent this line like the one below it? Labels get no indentation but their contents match the surrounding indentation.

danshick (changes requested):

src/screencast/screencast.c
tobiasjakobi: Not sure what you mean, this is how it looks in my editor:
screenshot_2021-05-23-125422

No indent for the label iself (as you say), one tab for the code itself (matches the surrounding).

danshick: Sorry, the GitHub mobile app lied to me.

danshick (approved):

columbarius (approved):

emersion:
Thanks!

Issue #137, opened 2021-05-20
Proposal: import buffers from pipewire


columbarius:
Currently we allocate a buffer, send it to wlroots for screencopy and then copy it onto the pipewire buffer when requested.

I would like to use the pipewire buffer chain directly. This means:
* using the add_buffer methods to allocate buffers
* using the pipewire on_process method as the main driver with the following process flow:
  * import pw_buffer to xdpw_simplebuffer
  * request screencopy for that buffer
  * export the xdpw_simplebuffer to pw_buffer
  * repeat
 
This might make screencopy_with_damage more inefficient, but i think when we want to drop the memcpy of shm buffers and only send fd's to the buffer around, we have to use more than one buffer to avoid race conditions.

columbarius:
> Can we completely replace MemPtr with MemFd and let PipeWire do the conversion for clients that expect MemPtr?

afaik yes. Otherwise we would have to do one mmap after the allocation, which would not be too bad. Otherwise we can get the fd from a MemPtr pw_buffer, so no problem there.

danshick:
I know that wlroots is no longer y-inverting frames, but can we be sure we won't ever need to do any bit-fiddling again for other compositors or strange graphics cards? That's the only reason I can think to leave MemPtr around as a legacy option long term. I don't think pipewire has any concept of y-inversion or similar.

I would like to be able to do zero frame processing in xdpw, so I support this idea in principle.

emersion:
> I don't think pipewire has any concept of y-inversion or similar.

In theory PipeWire has negative strides for this. But clients don't support it.

We could just y-invert the buffer in-place if we want, before handing it over to PipeWire.

danshick:
> We could just y-invert the buffer in-place if we want, before handing it over to PipeWire.

Yeah, of course. Why didn't I think of that.

columbarius:
> Otherwise we would have to do one mmap after the allocation, which would not be too bad. Otherwise we can get the fd from a MemPtr pw_buffer, so no problem there.

It doesn't work this way. When we allocate our own buffers, we are restricted to MemFd and DmaBuf. But most clients only handling MemPtr do use the MAP_BUFFERS flag, so pipewire does the mmapping. 

As far as i tested it (python script, obs, firefox) it worked without any problem.

columbarius:
Should we move from wl_shm_formats to drm_formats as the common format?

emersion:
[...]

Issue #122, opened 2021-04-24
xdpw should retry after a "out of buffers" error


emersion:
Instead of stalling the screencast, xdpw should continue pumping out frames from the compositor after running out of buffers.

danshick:
Do you mean we should retry every frame until it succeeds? Currently, when we're out of buffers, we free the frame and reregister the wlroots callbacks. We lose a frame but less chance of pushing a stale frame if pipewire lags. The screencast session should still continue.

emersion:
Hmmm. In my experience the screencast became completely stuck after the first "out of buffers" error, without any retry. I'll check again the code and see if we missed a case.

danshick:
@emersion, on the latest build, I currently get a ton of "out of buffers" errors from pipewire and the screencast continues to work. Can you check and see if you can reproduce the original issue?

emersion:
Sorry, I won't have time to investigate this further. I'll re-open if need be.

columbarius:
One thing to reduce out of buffer occurrences is to increase [XDPW_PWR_BUFFERS](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/include/pipewire_screencast.h#L6) so we don't get blocked, if a client uses the only buffer on the bus, but i didn't note any blocks because of that.

emersion:
Ref https://github.com/Plagman/gamescope/pull/219/commits/ef78e2e2707a75b6660809445fab578959a1548e - we need to add a `process` stream handler to know when a buffer becomes available again, and kick off wlr-screencopy again. This means we drop a frame -- but at least we don't stall the whole screen sharing session.

columbarius:
I'm not sure what we want to prioritise:
Do we want to stall the wlroots loop, when pipewire is out of buffers (which can be avoided with more pipewire buffers) or do we want to create a wlroots screencopy_frame for each frame of the screen and waste it by not sending it via pipewire, when no buffers are available.

Version 1:

1. register screencopy callback
2. create screencopy_frame
3. get buffer events
4. import buffer from pipewire
    -  buffer available -> copy frame from wlroots
    -  buffer not available -> goto 6.
5. screencopy successull -> export buffer
6. destroy screencopy_frame 
7. goto 1.


Version 2:
1. register screencopy callback
2. create screencopy_frame
3. get buffer events
4. import buffer from pipewire
    -  buffer available -> copy frame from wlroots
[...]

PR #141, opened 2021-05-26
use pipewire buffers

columbarius:
This MR aims to implement screencast driven by pipewire as explained in #137.

Resolves: #137 

~~Depends: #150~~
~~Depends: #148~~
~~Depends: #46~~
~~Depends: #61~~

### DIFF
--- a/include/screencast_common.h
+++ b/include/screencast_common.h
@@ -123,6 +123,8 @@ struct xdpw_wlr_output {
 
 void randname(char *buf);
 int anonymous_shm_open(void);
+struct wl_buffer *import_wl_shm_buffer(struct xdpw_screencast_instance *cast, int fd,
+	enum wl_shm_format fmt, int width, int height, int stride);
 enum spa_video_format xdpw_format_pw_from_wl_shm(enum wl_shm_format format);
 enum spa_video_format xdpw_format_pw_strip_alpha(enum spa_video_format format);
 
--- a/include/screencast_common.h
+++ b/include/screencast_common.h
@@ -42,6 +42,17 @@ struct xdpw_frame_damage {
 };
 
 struct xdpw_frame {
+	uint32_t size;
+	uint32_t stride;
+	bool y_invert;
+	uint64_t tv_sec;
+	uint32_t tv_nsec;
+	struct xdpw_frame_damage damage;
+	struct wl_buffer *buffer;
+	struct pw_buffer *current_pw_buffer;
+};
+
+struct xdpw_screencopy_frame {
 	uint32_t width;
 	uint32_t height;
 	uint32_t size;
@@ -83,6 +94,7 @@ struct xdpw_screencast_instance {
 	uint32_t refcount;
 	struct xdpw_screencast_context *ctx;
 	bool initialized;
+	struct xdpw_frame current_frame;
 
 	// pipewire
 	struct pw_stream *stream;
@@ -92,14 +104,13 @@ struct xdpw_screencast_instance {
 	uint32_t node_id;
 	bool pwr_stream_state;
 	uint32_t framerate;
-	struct pw_buffer *current_pw_buffer;
 
 	// wlroots
 	struct zwlr_screencopy_frame_v1 *frame_callback;
 	struct xdpw_wlr_output *target_output;
 	uint32_t max_framerate;
 	struct zwlr_screencopy_frame_v1 *wlr_frame;
-	struct xdpw_frame simple_frame;
+	struct xdpw_screencopy_frame screencopy_frame;
 	bool with_cursor;
 	int err;
 	bool quit;
--- a/include/screencast_common.h
+++ b/include/screencast_common.h
@@ -29,6 +29,13 @@ enum xdpw_chooser_types {
   XDPW_CHOOSER_DMENU,
 };
 
+enum xdpw_frame_state {
+  XDPW_FRAME_STATE_NONE,
+  XDPW_FRAME_STATE_RENEG,
+  XDPW_FRAME_STATE_FAILED,
+  XDPW_FRAME_STATE_SUCCESS,
+};
+
 struct xdpw_output_chooser {
 	enum xdpw_chooser_types type;
 	char *cmd;
@@ -89,6 +96,7 @@ struct xdpw_screencast_instance {
 	struct xdpw_screencast_context *ctx;
 	bool initialized;
 	struct xdpw_frame current_frame;
+	enum xdpw_frame_state frame_state;
 
 	// pipewire
 	struct pw_stream *stream;
@@ -108,7 +116,6 @@ struct xdpw_screencast_instance {
 	bool with_cursor;
 	int err;
 	bool quit;
-	bool copied;
 
 	// fps limit
 	struct fps_limit_state fps_limit;
--- a/include/screencast_common.h
+++ b/include/screencast_common.h
@@ -108,6 +108,7 @@ struct xdpw_screencast_instance {
 	bool with_cursor;
 	int err;
 	bool quit;
+	bool copied;
 
 	// fps limit
 	struct fps_limit_state fps_limit;
--- a/include/screencast_common.h
+++ b/include/screencast_common.h
@@ -57,13 +57,7 @@ struct xdpw_screencopy_frame {
 	uint32_t height;
 	uint32_t size;
 	uint32_t stride;
-	bool y_invert;
-	uint64_t tv_sec;
-	uint32_t tv_nsec;
 	enum wl_shm_format format;
-	struct xdpw_frame_damage damage;
-	struct wl_buffer *buffer;
-	void *data;
 };
 
 struct xdpw_screencast_context {
--- a/include/screencast_common.h
+++ b/include/screencast_common.h
@@ -93,6 +93,7 @@ struct xdpw_screencast_instance {
 	uint32_t node_id;
 	bool pwr_stream_state;
 	uint32_t framerate;
+	struct pw_buffer *current_pw_buffer;
 
 	// wlroots
 	struct zwlr_screencopy_frame_v1 *frame_callback;
--- a/include/screencast_common.h
+++ b/include/screencast_common.h
@@ -85,7 +85,6 @@ struct xdpw_screencast_instance {
 	bool initialized;
 
 	// pipewire
-	struct spa_source *event;
 	struct pw_stream *stream;
 	struct spa_hook stream_listener;
 	struct spa_video_info_raw pwr_format;
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -44,6 +44,14 @@ static struct spa_pod *build_format(struct spa_pod_builder *b, enum spa_video_fo
 	return spa_pod_builder_pop(b, &f[0]);
 }
 
+static void pwr_handle_stream_process(void *data) {
+	struct xdpw_screencast_instance *cast = data;
+
+	logprint(TRACE, "pipewire: stream process");
+
+	xdpw_wlr_frame_start(cast);
+}
+
 static void pwr_handle_stream_state_changed(void *data,
 		enum pw_stream_state old, enum pw_stream_state state, const char *error) {
 	struct xdpw_screencast_instance *cast = data;
@@ -176,7 +184,16 @@ static const struct pw_stream_events pwr_stream_events = {
 	.param_changed = pwr_handle_stream_param_changed,
 	.add_buffer = pwr_handle_stream_add_buffer,
 	.remove_buffer = pwr_handle_stream_remove_buffer,
+	.process = pwr_handle_stream_process,
 };
+
+void xdpw_pwr_trigger_process(struct xdpw_screencast_instance *cast) {
+	pw_stream_trigger_process(cast->stream);
+}
+
+bool xdpw_pwr_is_driving(struct xdpw_screencast_instance *cast) {
+	return pw_stream_is_driving(cast->stream);
+}
 
 void xdpw_pwr_dequeue_buffer(struct xdpw_screencast_instance *cast) {
 	logprint(TRACE, "pipewire: dequeueing buffer");
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -13,22 +13,6 @@
 #include "xdpw.h"
 #include "logger.h"
 
-static void writeFrameData(void *pwFramePointer, void *wlrFramePointer,
-		uint32_t height, uint32_t stride, bool inverted) {
-	if (!inverted) {
-		memcpy(pwFramePointer, wlrFramePointer, height * stride);
-		return;
-	}
-
-	for (size_t i = 0; i < (size_t)height; ++i) {
-		void *flippedWlrRowPointer = wlrFramePointer + ((height - i - 1) * stride);
-		void *pwRowPointer = pwFramePointer + (i * stride);
-		memcpy(pwRowPointer, flippedWlrRowPointer, stride);
-	}
-
-	return;
-}
-
 static struct spa_pod *build_format(struct spa_pod_builder *b, enum spa_video_format format,
 		uint32_t width, uint32_t height, uint32_t framerate) {
 	struct spa_pod_frame f[1];
@@ -143,6 +127,7 @@ static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
 		d[0].chunk->offset = 0;
 		d[0].flags = 0;
 		d[0].fd = anonymous_shm_open();
+		d[0].data = NULL;
 
 		if (d[0].fd == -1) {
 			logprint(ERROR, "pipewire: unable to create anonymous filedescriptor");
@@ -158,13 +143,9 @@ static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
 			return;
 		}
 
-		// mmap buffer, so we can use the data_ptr in on_process
-		d[0].data = mmap(NULL, d[0].maxsize, PROT_READ | PROT_WRITE, MAP_SHARED, d[0].fd, d[0].mapoffset);
-		if (d[0].data == MAP_FAILED) {
-			logprint(ERROR, "pipewire: unable to mmap memory");
-			cast->err = 1;
-			return;
-		}
+		// create wl_buffer
+		buffer->user_data = import_wl_shm_buffer(cast, d[0].fd, cast->screencopy_frame.format,
+			cast->screencopy_frame.width, cast->screencopy_frame.height, cast->screencopy_frame.stride);
 	}
 }
 
@@ -174,7 +155,7 @@ static void pwr_handle_stream_remove_buffer(void *data, struct pw_buffer *buffer
 	struct spa_data *d = buffer->buffer->datas;
 	switch (d[0].type) {
 	case SPA_DATA_MemFd:
-		munmap(d[0].data, d[0].maxsize);
+		wl_buffer_destroy(buffer->user_data);
 		close(d[0].fd);
 		break;
 	default:
@@ -196,6 +177,7 @@ void xdpw_pwr_dequeue_buffer(struct xdpw_screencast_instance *cast) {
 	assert(cast->current_frame.current_pw_buffer == NULL);
 	if ((cast->current_frame.current_pw_buffer = pw_stream_dequeue_buffer(cast->stream)) == NULL) {
 		logprint(WARN, "pipewire: out of buffers");
+		cast->current_frame.buffer = NULL;
 		return;
 	}
 
@@ -203,6 +185,7 @@ void xdpw_pwr_dequeue_buffer(struct xdpw_screencast_instance *cast) {
 	struct spa_data *d = spa_buf->datas;
 	cast->current_frame.size = d[0].chunk->size;
 	cast->current_frame.stride = d[0].chunk->stride;
+	cast->current_frame.buffer = cast->current_frame.current_pw_buffer->user_data;
 }
 
 void xdpw_pwr_enqueue_buffer(struct xdpw_screencast_instance *cast) {
@@ -224,16 +207,14 @@ void xdpw_pwr_enqueue_buffer(struct xdpw_screencast_instance *cast) {
 		h->seq = cast->seq++;
 		h->dts_offset = 0;
 	}
-	if (d[0].data == NULL) {
-		logprint(TRACE, "pipewire: data pointer undefined");
-		goto queue;
+
+	if (cast->screencopy_frame.y_invert) {
+		//TODO: Flip buffer or set stride negative
+		cast->err = 1;
 	}
 
-	writeFrameData(d[0].data, cast->screencopy_frame.data, cast->screencopy_frame.height,
-		cast->screencopy_frame.stride, cast->screencopy_frame.y_invert);
-
 	logprint(TRACE, "********************");
-	logprint(TRACE, "pipewire: pointer %p", d[0].data);
+	logprint(TRACE, "pipewire: fd %u", d[0].fd);
 	logprint(TRACE, "pipewire: size %d", d[0].maxsize);
 	logprint(TRACE, "pipewire: stride %d", d[0].chunk->stride);
 	logprint(TRACE, "pipewire: width %d", cast->screencopy_frame.width);
@@ -241,10 +222,10 @@ void xdpw_pwr_enqueue_buffer(struct xdpw_screencast_instance *cast) {
 	logprint(TRACE, "pipewire: y_invert %d", cast->screencopy_frame.y_invert);
 	logprint(TRACE, "********************");
 
-queue:
 	pw_stream_queue_buffer(cast->stream, pw_buf);
 
 	cast->current_frame.current_pw_buffer = NULL;
+	cast->current_frame.buffer = NULL;
 }
 
 void pwr_update_stream_param(struct xdpw_screencast_instance *cast) {
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -7,6 +7,7 @@
 #include <spa/param/video/format-utils.h>
 #include <sys/mman.h>
 #include <unistd.h>
+#include <assert.h>
 
 #include "wlr_screencast.h"
 #include "xdpw.h"
@@ -230,6 +231,58 @@ static const struct pw_stream_events pwr_stream_events = {
 	.add_buffer = pwr_handle_stream_add_buffer,
 	.remove_buffer = pwr_handle_stream_remove_buffer,
 };
+
+void xdpw_pwr_dequeue_buffer(struct xdpw_screencast_instance *cast) {
+	logprint(TRACE, "pipewire: dequeueing buffer");
+
+	assert(cast->current_pw_buffer == NULL);
+	if ((cast->current_pw_buffer = pw_stream_dequeue_buffer(cast->stream)) == NULL) {
+		logprint(WARN, "pipewire: out of buffers");
+		return;
+	}
+}
+
+void xdpw_pwr_enqueue_buffer(struct xdpw_screencast_instance *cast) {
+	logprint(TRACE, "pipewire: exporting buffer");
+
+	struct pw_buffer *pw_buf = cast->current_pw_buffer;
+
+	if (!pw_buf) {
+		logprint(TRACE, "pipewire: no pipewire buffer to queue");
+		return;
+	}
+
+	struct spa_buffer *spa_buf = pw_buf->buffer;
+	struct spa_data *d = spa_buf->datas;
+	struct spa_meta_header *h;
+	if ((h = spa_buffer_find_meta_data(spa_buf, SPA_META_Header, sizeof(*h)))) {
+		h->pts = -1;
+		h->flags = 0;
+		h->seq = cast->seq++;
+		h->dts_offset = 0;
+	}
+	if (d[0].data == NULL) {
+		logprint(TRACE, "pipewire: data pointer undefined");
+		goto queue;
+	}
+
+	writeFrameData(d[0].data, cast->simple_frame.data, cast->simple_frame.height,
+		cast->simple_frame.stride, cast->simple_frame.y_invert);
+
+	logprint(TRACE, "********************");
+	logprint(TRACE, "pipewire: pointer %p", d[0].data);
+	logprint(TRACE, "pipewire: size %d", d[0].maxsize);
+	logprint(TRACE, "pipewire: stride %d", d[0].chunk->stride);
+	logprint(TRACE, "pipewire: width %d", cast->simple_frame.width);
+	logprint(TRACE, "pipewire: height %d", cast->simple_frame.height);
+	logprint(TRACE, "pipewire: y_invert %d", cast->simple_frame.y_invert);
+	logprint(TRACE, "********************");
+
+queue:
+	pw_stream_queue_buffer(cast->stream, pw_buf);
+
+	cast->current_pw_buffer = NULL;
+}
 
 void pwr_update_stream_param(struct xdpw_screencast_instance *cast) {
 	logprint(TRACE, "pipewire: stream update parameters");
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -72,6 +72,9 @@ static void pwr_handle_stream_state_changed(void *data,
 	switch (state) {
 	case PW_STREAM_STATE_STREAMING:
 		cast->pwr_stream_state = true;
+		if (!cast->wlr_frame) {
+			xdpw_wlr_register_cb(cast);
+		}
 		break;
 	default:
 		cast->pwr_stream_state = false;
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -57,7 +57,7 @@ static void pwr_handle_stream_state_changed(void *data,
 	case PW_STREAM_STATE_STREAMING:
 		cast->pwr_stream_state = true;
 		if (!cast->wlr_frame) {
-			xdpw_wlr_register_cb(cast);
+			xdpw_wlr_frame_start(cast);
 		}
 		break;
 	default:
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -150,9 +150,16 @@ static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
 }
 
 static void pwr_handle_stream_remove_buffer(void *data, struct pw_buffer *buffer) {
+	struct xdpw_screencast_instance *cast = data;
+
 	logprint(TRACE, "pipewire: remove buffer event handle");
 
 	struct spa_data *d = buffer->buffer->datas;
+	if (cast->current_frame.current_pw_buffer == buffer) {
+		logprint(TRACE, "pipewire: remove buffer currently in use");
+		cast->current_frame.current_pw_buffer = NULL;
+		cast->current_frame.buffer = NULL;
+	}
 	switch (d[0].type) {
 	case SPA_DATA_MemFd:
 		wl_buffer_destroy(buffer->user_data);
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -100,8 +100,8 @@ static void pwr_handle_stream_param_changed(void *data, uint32_t id,
 		SPA_TYPE_OBJECT_ParamBuffers, SPA_PARAM_Buffers,
 		SPA_PARAM_BUFFERS_buffers, SPA_POD_CHOICE_RANGE_Int(XDPW_PWR_BUFFERS, 1, 32),
 		SPA_PARAM_BUFFERS_blocks,  SPA_POD_Int(1),
-		SPA_PARAM_BUFFERS_size,    SPA_POD_Int(cast->simple_frame.size),
-		SPA_PARAM_BUFFERS_stride,  SPA_POD_Int(cast->simple_frame.stride),
+		SPA_PARAM_BUFFERS_size,    SPA_POD_Int(cast->screencopy_frame.size),
+		SPA_PARAM_BUFFERS_stride,  SPA_POD_Int(cast->screencopy_frame.stride),
 		SPA_PARAM_BUFFERS_align,   SPA_POD_Int(XDPW_PWR_ALIGN),
 		SPA_PARAM_BUFFERS_dataType,SPA_POD_CHOICE_FLAGS_Int(1<<SPA_DATA_MemFd));
 
@@ -133,10 +133,10 @@ static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
 	logprint(TRACE, "pipewire: selected buffertype %u", d[0].type);
 	// Prepare buffer for choosen type
 	if (d[0].type == SPA_DATA_MemFd) {
-		d[0].maxsize = cast->simple_frame.size;
+		d[0].maxsize = cast->screencopy_frame.size;
 		d[0].mapoffset = 0;
-		d[0].chunk->size = cast->simple_frame.size;
-		d[0].chunk->stride = cast->simple_frame.stride;
+		d[0].chunk->size = cast->screencopy_frame.size;
+		d[0].chunk->stride = cast->screencopy_frame.stride;
 		d[0].chunk->offset = 0;
 		d[0].flags = 0;
 		d[0].fd = anonymous_shm_open();
@@ -190,17 +190,22 @@ static const struct pw_stream_events pwr_stream_events = {
 void xdpw_pwr_dequeue_buffer(struct xdpw_screencast_instance *cast) {
 	logprint(TRACE, "pipewire: dequeueing buffer");
 
-	assert(cast->current_pw_buffer == NULL);
-	if ((cast->current_pw_buffer = pw_stream_dequeue_buffer(cast->stream)) == NULL) {
+	assert(cast->current_frame.current_pw_buffer == NULL);
+	if ((cast->current_frame.current_pw_buffer = pw_stream_dequeue_buffer(cast->stream)) == NULL) {
 		logprint(WARN, "pipewire: out of buffers");
 		return;
 	}
+
+	struct spa_buffer *spa_buf = cast->current_frame.current_pw_buffer->buffer;
+	struct spa_data *d = spa_buf->datas;
+	cast->current_frame.size = d[0].chunk->size;
+	cast->current_frame.stride = d[0].chunk->stride;
 }
 
 void xdpw_pwr_enqueue_buffer(struct xdpw_screencast_instance *cast) {
 	logprint(TRACE, "pipewire: exporting buffer");
 
-	struct pw_buffer *pw_buf = cast->current_pw_buffer;
+	struct pw_buffer *pw_buf = cast->current_frame.current_pw_buffer;
 
 	if (!pw_buf) {
 		logprint(TRACE, "pipewire: no pipewire buffer to queue");
@@ -221,22 +226,22 @@ void xdpw_pwr_enqueue_buffer(struct xdpw_screencast_instance *cast) {
 		goto queue;
 	}
 
-	writeFrameData(d[0].data, cast->simple_frame.data, cast->simple_frame.height,
-		cast->simple_frame.stride, cast->simple_frame.y_invert);
+	writeFrameData(d[0].data, cast->screencopy_frame.data, cast->screencopy_frame.height,
+		cast->screencopy_frame.stride, cast->screencopy_frame.y_invert);
 
 	logprint(TRACE, "********************");
 	logprint(TRACE, "pipewire: pointer %p", d[0].data);
 	logprint(TRACE, "pipewire: size %d", d[0].maxsize);
 	logprint(TRACE, "pipewire: stride %d", d[0].chunk->stride);
-	logprint(TRACE, "pipewire: width %d", cast->simple_frame.width);
-	logprint(TRACE, "pipewire: height %d", cast->simple_frame.height);
-	logprint(TRACE, "pipewire: y_invert %d", cast->simple_frame.y_invert);
+	logprint(TRACE, "pipewire: width %d", cast->screencopy_frame.width);
+	logprint(TRACE, "pipewire: height %d", cast->screencopy_frame.height);
+	logprint(TRACE, "pipewire: y_invert %d", cast->screencopy_frame.y_invert);
 	logprint(TRACE, "********************");
 
 queue:
 	pw_stream_queue_buffer(cast->stream, pw_buf);
 
-	cast->current_pw_buffer = NULL;
+	cast->current_frame.current_pw_buffer = NULL;
 }
 
 void pwr_update_stream_param(struct xdpw_screencast_instance *cast) {
@@ -247,10 +252,10 @@ void pwr_update_stream_param(struct xdpw_screencast_instance *cast) {
 		SPA_POD_BUILDER_INIT(params_buffer, sizeof(params_buffer));
 	const struct spa_pod *params[1];
 
-	enum spa_video_format format = xdpw_format_pw_from_wl_shm(cast->simple_frame.format);
+	enum spa_video_format format = xdpw_format_pw_from_wl_shm(cast->screencopy_frame.format);
 
 	params[0] = build_format(&b, format,
-			cast->simple_frame.width, cast->simple_frame.height, cast->framerate);
+			cast->screencopy_frame.width, cast->screencopy_frame.height, cast->framerate);
 
 	pw_stream_update_params(stream, params, 1);
 }
@@ -277,10 +282,10 @@ void xdpw_pwr_stream_create(struct xdpw_screencast_instance *cast) {
 	}
 	cast->pwr_stream_state = false;
 
-	enum spa_video_format format = xdpw_format_pw_from_wl_shm(cast->simple_frame.format);
+	enum spa_video_format format = xdpw_format_pw_from_wl_shm(cast->screencopy_frame.format);
 
 	const struct spa_pod *param = build_format(&b, format,
-			cast->simple_frame.width, cast->simple_frame.height, cast->framerate);
+			cast->screencopy_frame.width, cast->screencopy_frame.height, cast->framerate);
 
 	pw_stream_add_listener(cast->stream, &cast->stream_listener,
 		&pwr_stream_events, cast);
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -60,51 +60,6 @@ static struct spa_pod *build_format(struct spa_pod_builder *b, enum spa_video_fo
 	return spa_pod_builder_pop(b, &f[0]);
 }
 
-static void pwr_on_event(void *data, uint64_t expirations) {
-	struct xdpw_screencast_instance *cast = data;
-	struct pw_buffer *pw_buf;
-	struct spa_buffer *spa_buf;
-	struct spa_meta_header *h;
-	struct spa_data *d;
-
-	logprint(TRACE, "********************");
-	logprint(TRACE, "pipewire: event fired");
-
-	if ((pw_buf = pw_stream_dequeue_buffer(cast->stream)) == NULL) {
-		logprint(WARN, "pipewire: out of buffers");
-		goto out;
-	}
-
-	spa_buf = pw_buf->buffer;
-	d = spa_buf->datas;
-	if ((d[0].data) == NULL) {
-		logprint(TRACE, "pipewire: data pointer undefined");
-		goto out;
-	}
-	if ((h = spa_buffer_find_meta_data(spa_buf, SPA_META_Header, sizeof(*h)))) {
-		h->pts = -1;
-		h->flags = 0;
-		h->seq = cast->seq++;
-		h->dts_offset = 0;
-	}
-
-	writeFrameData(d[0].data, cast->simple_frame.data, cast->simple_frame.height,
-		cast->simple_frame.stride, cast->simple_frame.y_invert);
-
-	logprint(TRACE, "pipewire: pointer %p", d[0].data);
-	logprint(TRACE, "pipewire: size %d", d[0].maxsize);
-	logprint(TRACE, "pipewire: stride %d", d[0].chunk->stride);
-	logprint(TRACE, "pipewire: width %d", cast->simple_frame.width);
-	logprint(TRACE, "pipewire: height %d", cast->simple_frame.height);
-	logprint(TRACE, "pipewire: y_invert %d", cast->simple_frame.y_invert);
-	logprint(TRACE, "********************");
-
-	pw_stream_queue_buffer(cast->stream, pw_buf);
-
-out:
-	xdpw_wlr_frame_free(cast);
-}
-
 static void pwr_handle_stream_state_changed(void *data,
 		enum pw_stream_state old, enum pw_stream_state state, const char *error) {
 	struct xdpw_screencast_instance *cast = data;
@@ -321,11 +276,6 @@ void xdpw_pwr_stream_create(struct xdpw_screencast_instance *cast) {
 		abort();
 	}
 	cast->pwr_stream_state = false;
-
-	/* make an event to signal frame ready */
-	cast->event =
-		pw_loop_add_event(state->pw_loop, pwr_on_event, cast);
-	logprint(DEBUG, "pipewire: registered event %p", cast->event);
 
 	enum spa_video_format format = xdpw_format_pw_from_wl_shm(cast->simple_frame.format);
 
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -208,7 +208,7 @@ void xdpw_pwr_enqueue_buffer(struct xdpw_screencast_instance *cast) {
 		h->dts_offset = 0;
 	}
 
-	if (cast->screencopy_frame.y_invert) {
+	if (cast->current_frame.y_invert) {
 		//TODO: Flip buffer or set stride negative
 		cast->err = 1;
 	}
@@ -219,7 +219,7 @@ void xdpw_pwr_enqueue_buffer(struct xdpw_screencast_instance *cast) {
 	logprint(TRACE, "pipewire: stride %d", d[0].chunk->stride);
 	logprint(TRACE, "pipewire: width %d", cast->screencopy_frame.width);
 	logprint(TRACE, "pipewire: height %d", cast->screencopy_frame.height);
-	logprint(TRACE, "pipewire: y_invert %d", cast->screencopy_frame.y_invert);
+	logprint(TRACE, "pipewire: y_invert %d", cast->current_frame.y_invert);
 	logprint(TRACE, "********************");
 
 	pw_stream_queue_buffer(cast->stream, pw_buf);
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -201,10 +201,7 @@ void xdpw_pwr_enqueue_buffer(struct xdpw_screencast_instance *cast) {
 	struct pw_buffer *pw_buf = cast->current_frame.current_pw_buffer;
 	bool buffer_corrupt = cast->frame_state != XDPW_FRAME_STATE_SUCCESS;
 
-	if (!pw_buf) {
-		logprint(TRACE, "pipewire: no pipewire buffer to queue");
-		return;
-	}
+	assert(pw_buf);
 
 	struct spa_buffer *spa_buf = pw_buf->buffer;
 	struct spa_data *d = spa_buf->datas;
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -199,6 +199,7 @@ void xdpw_pwr_enqueue_buffer(struct xdpw_screencast_instance *cast) {
 	logprint(TRACE, "pipewire: exporting buffer");
 
 	struct pw_buffer *pw_buf = cast->current_frame.current_pw_buffer;
+	bool buffer_corrupt = cast->frame_state != XDPW_FRAME_STATE_SUCCESS;
 
 	if (!pw_buf) {
 		logprint(TRACE, "pipewire: no pipewire buffer to queue");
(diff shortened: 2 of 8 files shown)


columbarius:
Update: works on my machine!

Please take a look if it's going in the right direction. Performance is about doubled (at least from the cpu reading in obs).  

I have to look into the shutdown procedure, if there is sth. to improve.

Sry for the last commit. The changes with using pipewire to allocate buffers, importing then and restructuring the wlroots screencopy handlers were so connected, that i couldn't find a good spot, which didn't produce a memory leak or segfault in an itermediate step, so I added it as once.

danshick:
@emersion 

We need to push a bugfix release before this lands (which may be sooner than I expected, thanks columbarius). People without configs who also don't have a default chooser installed are experiencing errors.

columbarius:
This `zwlr_screencopy_frame_v1@12: error 1: invalid buffer stride` error appears on resolution change.

```
2021/05/28 10:30:10 [TRACE] - ********************
2021/05/28 10:30:10 [TRACE] - wlroots: frame destroyed
2021/05/28 10:30:10 [TRACE] - wlroots: callbacks registered
2021/05/28 10:30:10 [TRACE] - event-loop: got wayland event
2021/05/28 10:30:10 [TRACE] - wlroots: buffer event handler
2021/05/28 10:30:10 [TRACE] - pipewire: importing buffer
2021/05/28 10:30:10 [TRACE] - wlroots: linux_dmabuf event handler
2021/05/28 10:30:10 [TRACE] - wlroots: buffer_done event handler
2021/05/28 10:30:10 [TRACE] - wlroots: shm buffer exists
2021/05/28 10:30:10 [TRACE] - wlroots: frame copied
2021/05/28 10:30:10 [TRACE] - event-loop: got wayland event
2021/05/28 10:30:11 [TRACE] - event-loop: got wayland event
2021/05/28 10:30:11 [TRACE] - wlroots: flags event handler
2021/05/28 10:30:11 [TRACE] - wlroots: damage event handler
2021/05/28 10:30:11 [TRACE] - wlroots: ready event handler
2021/05/28 10:30:11 [TRACE] - pipewire: exporting buffer
2021/05/28 10:30:11 [TRACE] - ********************
2021/05/28 10:30:11 [TRACE] - pipewire: pointer 0x6070001207c0
2021/05/28 10:30:11 [TRACE] - pipewire: size 7680000
2021/05/28 10:30:11 [TRACE] - pipewire: stride 6400
2021/05/28 10:30:11 [TRACE] - pipewire: width 1600
2021/05/28 10:30:11 [TRACE] - pipewire: height 1200
2021/05/28 10:30:11 [TRACE] - pipewire: y_invert 0
2021/05/28 10:30:11 [TRACE] - ********************
2021/05/28 10:30:11 [TRACE] - wlroots: frame destroyed
2021/05/28 10:30:11 [TRACE] - wlroots: callbacks registered
2021/05/28 10:30:11 [TRACE] - event-loop: got wayland event
2021/05/28 10:30:11 [TRACE] - wlroots: buffer event handler
2021/05/28 10:30:11 [TRACE] - pipewire: importing buffer
2021/05/28 10:30:11 [TRACE] - wlroots: buffer properties changed
2021/05/28 10:30:11 [DEBUG] - wlroots: reset buffer
2021/05/28 10:30:11 [TRACE] - wlroots: linux_dmabuf event handler
2021/05/28 10:30:11 [TRACE] - wlroots: buffer_done event handler
2021/05/28 10:30:11 [TRACE] - pipewire: exporting buffer
2021/05/28 10:30:11 [TRACE] - ********************
2021/05/28 10:30:11 [TRACE] - pipewire: pointer 0x6070001207c0
2021/05/28 10:30:11 [TRACE] - pipewire: size 7680000
2021/05/28 10:30:11 [TRACE] - pipewire: stride 6400
2021/05/28 10:30:11 [TRACE] - pipewire: width 1920
2021/05/28 10:30:11 [TRACE] - pipewire: height 1080
2021/05/28 10:30:11 [TRACE] - pipewire: y_invert 0
2021/05/28 10:30:11 [TRACE] - ********************
2021/05/28 10:30:11 [TRACE] - pipewire: stream update parameters
2021/05/28 10:30:11 [TRACE] - wlroots: frame destroyed
2021/05/28 10:30:11 [TRACE] - wlroots: callbacks registered
2021/05/28 10:30:11 [TRACE] - event-loop: got pipewire event
2021/05/28 10:30:11 [TRACE] - event-loop: got wayland event
2021/05/28 10:30:11 [TRACE] - wlroots: buffer event handler
2021/05/28 10:30:11 [TRACE] - pipewire: importing buffer
2021/05/28 10:30:11 [TRACE] - wlroots: linux_dmabuf event handler
2021/05/28 10:30:11 [TRACE] - wlroots: buffer_done event handler
2021/05/28 10:30:11 [TRACE] - wlroots: shm buffer exists
2021/05/28 10:30:11 [TRACE] - wlroots: frame copied
2021/05/28 10:30:11 [TRACE] - event-loop: got pipewire event
2021/05/28 10:30:11 [TRACE] - pipewire: remove buffer event handle
2021/05/28 10:30:11 [TRACE] - pipewire: can't remove buffer. still in use
2021/05/28 10:30:11 [TRACE] - pipewire: stream parameters changed
2021/05/28 10:30:11 [TRACE] - event-loop: got wayland event
zwlr_screencopy_frame_v1@12: error 1: invalid buffer stride
2021/05/28 10:30:11 [ERROR] - wl_display_dispatch failed: Protokollfehler

=================================================================
==3091==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 320 byte(s) in 4 object(s) allocated from:
    #0 0x7f5630a7c459 in __interceptor_calloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cpp:154
    #1 0x7f56309b9afb in wl_proxy_marshal_array_constructor_versioned (/usr/lib/libwayland-client.so.0+0x6afb)
```

columbarius:
> @emersion
> 
> We need to push a bugfix release before this lands (which may be sooner than I expected, thanks columbarius). People without configs who also don't have a default chooser installed are experiencing errors.

I second that part with the bugfix release.

I don't see this MR merged soon, since the change in resolution/renegotiation part needs some work, where I'm not sure how to do it (also important to the point, that pipewire doesn't kills our buffers while we are using them) and I would like to make things right in a way, that it is nicely compatible with both shm and dmabuf depending on the clients preferences. I just had some unexpected time and made those changes to a point, where it works if you don't do anything exceptional.

columbarius:
Ok, i'm restarting the stream after a renegotiation by some kind of refcounting in {add,remove}_buffer. It's not elegant, but it works for now. 

Will have to modify obs again to test what's happening, if a client requests param changes. We might have to add a lock to the imported buffer and register the buffer_destroy call on the release of that lock, to prevent the buffer from being destroyed while wlroots is using it.

We might be able to just drop that buffer without the requeueing the buffer to pipewire beforehand.

Next step would be to rebase the/create a new dmabuf MR on top of the new flow. To test it I have a obs-studio branch ready (https://github.com/columbarius/obs-studio/tree/egl-modifiers).

emersion:
> prevent the buffer from being destroyed while wlroots is using it

It should be fine to destroy the `wl_buffer` while wlroots is writing to it, if we no longer care about the buffer and don't need the result. wlroots ref'counts its buffers.

columbarius:
> It should be fine to destroy the `wl_buffer` while wlroots is writing to it, if we no longer care about the buffer and don't need the result. wlroots ref'counts its buffers.

@emersion do you have an explanation how this error occures (
https://github.com/emersion/xdg-desktop-portal-wlr/pull/141#issuecomment-850255089
) if i destroy the buffer (wl_buffer and the underlying shm buffer)?

emersion:
Can you grab a `WAYLAND_DEBUG=1` log?

columbarius:
2021/06/14 12:10:37 [TRACE] - pipewire: importing buffer
2021/06/14 12:10:37 [TRACE] - wlroots: buffer properties changed
2021/06/14 12:10:37 [DEBUG] - wlroots: reset buffer
[296121.345] zwlr_screencopy_frame_v1@10.linux_dmabuf(875713089, 1600, 1200)
2021/06/14 12:10:37 [TRACE] - wlroots: linux_dmabuf event handler
[296121.386] zwlr_screencopy_frame_v1@10.buffer_done()
2021/06/14 12:10:37 [TRACE] - wlroots: buffer_done event handler
2021/06/14 12:10:37 [TRACE] - pipewire: exporting buffer
2021/06/14 12:10:37 [TRACE] - ********************
2021/06/14 12:10:37 [TRACE] - pipewire: pointer 0x607000018350
2021/06/14 12:10:37 [TRACE] - pipewire: size 8294400
2021/06/14 12:10:37 [TRACE] - pipewire: stride 7680
2021/06/14 12:10:37 [TRACE] - pipewire: width 1600
2021/06/14 12:10:37 [TRACE] - pipewire: height 1200
2021/06/14 12:10:37 [TRACE] - pipewire: y_invert 0
2021/06/14 12:10:37 [TRACE] - ********************
2021/06/14 12:10:37 [TRACE] - pipewire: stream update parameters
[296121.657]  -> zwlr_screencopy_frame_v1@10.destroy()
2021/06/14 12:10:37 [TRACE] - wlroots: frame destroyed
2021/06/14 12:10:37 [TRACE] - fps_limit: elapsed time since the last measurement: 271592476, target 3333333, target not met (ns)
[296121.731]  -> zwlr_screencopy_manager_v1@6.capture_output(new id zwlr_screencopy_frame_v1@3, 1, wl_output@7)
2021/06/14 12:10:37 [TRACE] - wlroots: callbacks registered
2021/06/14 12:10:37 [TRACE] - event-loop: got pipewire event
2021/06/14 12:10:37 [TRACE] - event-loop: got pipewire event
2021/06/14 12:10:37 [TRACE] - pipewire: remove buffer event handle
[296122.453]  -> wl_buffer@9.destroy()
2021/06/14 12:10:37 [TRACE] - pipewire: stream parameters changed
2021/06/14 12:10:37 [TRACE] - event-loop: got wayland event
[296135.786] wl_display@1.delete_id(10)
[296135.817] wl_display@1.delete_id(9)
[296135.835] zwlr_screencopy_frame_v1@3.buffer(0, 1600, 1200, 6400)
2021/06/14 12:10:37 [TRACE] - wlroots: buffer event handler
2021/06/14 12:10:37 [TRACE] - pipewire: importing buffer
2021/06/14 12:10:37 [WARN] - pipewire: out of buffers
[296135.930] zwlr_screencopy_frame_v1@3.linux_dmabuf(875713089, 1600, 1200)
2021/06/14 12:10:37 [TRACE] - wlroots: linux_dmabuf event handler
[296135.967] zwlr_screencopy_frame_v1@3.buffer_done()
2021/06/14 12:10:37 [TRACE] - wlroots: buffer_done event handler
2021/06/14 12:10:37 [TRACE] - wlroots: shm buffer exists
[296136.006]  -> zwlr_screencopy_frame_v1@3.copy_with_damageAddressSanitizer:DEADLYSIGNAL
=================================================================
==17609==ERROR: AddressSanitizer: SEGV on unknown address 0x000067800002 (pc 0x7f079dbe9520 bp 0x55b5f7be72a2 sp 0x7ffeb82a2630 T0)
==17609==The signal is caused by a READ memory access.
    #0 0x7f079dbe9520  (/usr/lib/libwayland-client.so.0+0xa520)
    #1 0x7f079dbe5c18 in wl_proxy_marshal_array_constructor_versioned (/usr/lib/libwayland-client.so.0+0x6c18)
    #2 0x7f079dbe5df2 in wl_proxy_marshal (/usr/lib/libwayland-client.so.0+0x6df2)
    #3 0x55b5f7bd5c80 in zwlr_screencopy_frame_v1_copy_with_damage protocols/wlr-screencopy-unstable-v1-client-protocol.h:487
    #4 0x55b5f7bd67db in wlr_frame_buffer_done ../src/screencast/wlr_screencast.c:133
    #5 0x7f079d4e7acc  (/usr/lib/libffi.so.7+0x6acc)
    #6 0x7f079d4e7039  (/usr/lib/libffi.so.7+0x6039)
    #7 0x7f079dbe8fe3  (/usr/lib/libwayland-client.so.0+0x9fe3)
    #8 0x7f079dbe5562  (/usr/lib/libwayland-client.so.0+0x6562)
    #9 0x7f079dbe6cab in wl_display_dispatch_queue_pending (/usr/lib/libwayland-client.so.0+0x7cab)
    #10 0x55b5f7bccbe8 in main ../src/core/main.c:212
    #11 0x7f079d8b3b24 in __libc_start_main (/usr/lib/libc.so.6+0x27b24)
    #12 0x55b5f7bcb8cd in _start (/path/to/project/sway-project/xdg-desktop-portal-wlr/build/xdg-desktop-portal-wlr+0xb8cd)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (/usr/lib/libwayland-client.so.0+0xa520)
==17609==ABORTING

emersion:
It seems like you're calling `zwlr_screencopy_frame_v1_copy_with_damage` with a `wl_buffer` you've previously destroyed? This is a use-after-free in the xdpw code. Once `wl_buffer_destroy` has been called, the `wl_buffer` must not be used in any subsequent request.

emersion:
Maybe destroying the `zwlr_screencopy_frame_v1` together with the `wl_buffer` can fix it?

columbarius:
> It seems like you're calling `zwlr_screencopy_frame_v1_copy_with_damage` with a `wl_buffer` you've previously destroyed? This is a use-after-free in the xdpw code. Once `wl_buffer_destroy` has been called, the `wl_buffer` must not be used in any subsequent request.

... i forgot to NULL pw_current_buffer and xdpw_simple_buffer.buffer if the import failed .... now i wasn't able to crash it without refcounting and all... Will investigate that later. Thanks!

columbarius:
```
2021/06/14 12:30:58 [TRACE] - wlroots: callbacks registered
2021/06/14 12:30:58 [TRACE] - event-loop: got pipewire event
2021/06/14 12:30:58 [TRACE] - event-loop: got wayland event
[1517378.923] wl_display@1.delete_id(9)
[1517378.941] zwlr_screencopy_frame_v1@11.buffer(0, 1600, 1200, 6400)
2021/06/14 12:30:58 [TRACE] - wlroots: buffer event handler
2021/06/14 12:30:58 [TRACE] - pipewire: importing buffer
2021/06/14 12:30:58 [WARN] - pipewire: out of buffers
[1517379.019] zwlr_screencopy_frame_v1@11.linux_dmabuf(875713089, 1600, 1200)
2021/06/14 12:30:58 [TRACE] - wlroots: linux_dmabuf event handler
[1517379.090] zwlr_screencopy_frame_v1@11.buffer_done()
2021/06/14 12:30:58 [TRACE] - wlroots: buffer_done event handler
2021/06/14 12:30:58 [DEBUG] - wlroots: failed to import buffer
[1517379.136]  -> zwlr_screencopy_frame_v1@11.destroy()
2021/06/14 12:30:58 [TRACE] - wlroots: frame destroyed
2021/06/14 12:30:58 [TRACE] - fps_limit: elapsed time since the last measurement: 261801446, target 3333333, target not met (ns)
[1517379.220]  -> zwlr_screencopy_manager_v1@6.capture_output(new id zwlr_screencopy_frame_v1@9, 1, wl_output@7)
2021/06/14 12:30:58 [TRACE] - wlroots: callbacks registered
2021/06/14 12:30:58 [TRACE] - event-loop: got wayland event
[1517379.392] wl_display@1.delete_id(11)
[1517379.427] zwlr_screencopy_frame_v1@9.buffer(0, 1600, 1200, 6400)
2021/06/14 12:30:58 [TRACE] - wlroots: buffer event handler
2021/06/14 12:30:58 [TRACE] - pipewire: importing buffer
[1517379.503] zwlr_screencopy_frame_v1@9.linux_dmabuf(875713089, 1600, 1200)
2021/06/14 12:30:58 [TRACE] - wlroots: linux_dmabuf event handler
[1517379.551] zwlr_screencopy_frame_v1@9.buffer_done()
2021/06/14 12:30:58 [TRACE] - wlroots: buffer_done event handler
2021/06/14 12:30:58 [TRACE] - wlroots: shm buffer exists
[1517379.601]  -> zwlr_screencopy_frame_v1@9.copy_with_damage(wl_buffer@10)
2021/06/14 12:30:58 [TRACE] - wlroots: frame copied
2021/06/14 12:30:58 [TRACE] - event-loop: got pipewire event
2021/06/14 12:30:58 [TRACE] - pipewire: remove buffer event handle
[1517379.769]  -> wl_buffer@10.destroy()
2021/06/14 12:30:58 [TRACE] - pipewire: stream parameters changed
2021/06/14 12:30:58 [TRACE] - event-loop: got wayland event
[1517379.872] wl_display@1.error(zwlr_screencopy_frame_v1@9, 1, "invalid buffer stride")
zwlr_screencopy_frame_v1@9: error 1: invalid buffer stride
2021/06/14 12:30:58 [ERROR] - wl_display_dispatch failed: Protokollfehler

=================================================================
==23795==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 320 byte(s) in 4 object(s) allocated from:
    #0 0x7f5c59d66459 in __interceptor_calloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cpp:154
    #1 0x7f5c59ca3afb in wl_proxy_marshal_array_constructor_versioned (/usr/lib/libwayland-client.so.0+0x6afb)

Direct leak of 125 byte(s) in 5 object(s) allocated from:
    #0 0x7f5c59d0b3f9 in __interceptor_strdup /build/gcc/src/gcc/libsanitizer/asan/asan_interceptors.cpp:454
    #1 0x5604fe5b2da5 in wlr_output_handle_geometry ../src/screencast/wlr_screencast.c:208
    #2 0x7f5c595a5acc  (/usr/lib/libffi.so.7+0x6acc)
```

Still a use after free.

emersion:
Interesting, there may be a use-after-free in wlroots here. Can you run the compositor with ASan enabled? Maybe print what the stride values are inside wlroots?

columbarius:
It has to be a use after free, because we destroy the buffer, while copy with damage is in progress. You are sure wlroots can/should handle that? (I'm asking, because that looks like sth. you should never do. Handing out a buffer to someone else and destroying while the other one is processing it.)
```
[1517379.601]  -> zwlr_screencopy_frame_v1@9.copy_with_damage(wl_buffer@10) <----
2021/06/14 12:30:58 [TRACE] - wlroots: frame copied
2021/06/14 12:30:58 [TRACE] - event-loop: got pipewire event
2021/06/14 12:30:58 [TRACE] - pipewire: remove buffer event handle
[1517379.769]  -> wl_buffer@10.destroy() <-----
2021/06/14 12:30:58 [TRACE] - pipewire: stream parameters changed
2021/06/14 12:30:58 [TRACE] - event-loop: got wayland event
[1517379.872] wl_display@1.error(zwlr_screencopy_frame_v1@9, 1, "invalid buffer stride") <----
zwlr_screencopy_frame_v1@9: error 1: invalid buffer stride
2021/06/14 12:30:58 [ERROR] - wl_display_dispatch failed: Protokollfehler
```

> Interesting, there may be a use-after-free in wlroots here. Can you run the compositor with ASan enabled? Maybe print what the stride values are inside wlroots?

you mean compiling wlroots and sway with asan?

emersion:
> You are sure wlroots can/should handle that?

Yeah, for instance it's valid for a client to `wl_surface.attach` a buffer and immediately destroy it after.

> you mean compiling wlroots and sway with asan?

Yes. https://github.com/swaywm/sway/wiki/Development-Setup#compiling-as-a-subproject

columbarius:
Ok, i printed the stride of the imported buffer in wlroots and they where different. 

I'm going forward with renaming the xdpw_frame to wlr_screencopy_frame (which will only hold the informations given by the wlr_screencopy protocol) and adding a "new" xdpw_frame, which holds the informations about the imported buffer from pipewire. This way we can check if the imported buffer is compatible with the bufferinformation advertised by wlroots. 
I somehow wanted to postpone this step until we need it for dmabuf screensharing, but it is not worth it.

danshick:
Thanks for cleaning up my poor naming conventions.

columbarius:
> Thanks for cleaning up my poor naming conventions.

Don't get your hopes too high up. I introduced a new xdpw_frame called simple_frame which is used in the interaction with pipewire, but also holds a wlr_buffer (that's why i didn't called it pipewire frame). 
It's currently too hot at my place for creativity, but i'm open to suggestions.

columbarius:
I'm having problems with the fps limter while cleaning the stuff up. I will deactiveate before the flow change and activate it after wards. We might fix that later, or just ignore it in the intermediate commits.

danshick:
I don't wanna break master, but in this PR is fine.

columbarius:
I cleaned the patches up and restructured it a bit, so the steps are smaller and I think we are close to review. Screencast should work at every commit, renegotiation will be broken on some.

danshick:
To be more explicit, I'm fine with broken commits on master for history preservation if they've never been at HEAD in a practical sense. Not sure if @emersion agrees, but if at rebase time, HEAD contains no regressions, I'm happy.

columbarius:
I'm using this and it seems to work fine. Also adding dmabuf supported doesn't look like we would need changes in the flow, so I'm marking this as ready for review.

danshick:
>... so I'm marking this as ready for review.

I'll build and use this branch today.

src/screencast/pipewire_screencast.c
```diff
@@ -170,10 +170,33 @@ static void pwr_handle_stream_param_changed(void *data, uint32_t id,
 	pw_stream_update_params(stream, params, 2);
 }
 
+static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
+	struct xdpw_screencast_instance *cast = data;
+	struct spa_data *d;
```
emersion, line 108: Style: this can be declared below, one the same line as the assignment.

src/screencast/pipewire_screencast.c
```diff
@@ -170,28 +172,59 @@ static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
...
-		d[0].type = SPA_DATA_MemPtr;
+	if ((d[0].type & (1u << SPA_DATA_MemFd)) > 0) {
+		d[0].type = SPA_DATA_MemFd;
 	} else {
 		logprint(ERROR, "pipewire: unsupported buffer type");
 		cast->err = 1;
 		return;
 	}
 
+	logprint(TRACE, "pipewire: selected buffertype %d", d[0].type);
```
emersion, line 123: We should probably use `%u` here

src/screencast/pipewire_screencast.c
```diff
@@ -170,28 +172,59 @@ static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
...
 	// Prepare buffer for choosen type
-	if (d[0].type == SPA_DATA_MemPtr) {
+	if (d[0].type == SPA_DATA_MemFd) {
+		d[0].type = SPA_DATA_MemFd;
```
emersion, line 126: This assignment is a no-op

src/screencast/pipewire_screencast.c
```diff
@@ -170,28 +172,59 @@ static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
...
 		d[0].maxsize = cast->simple_frame.size;
 		d[0].mapoffset = 0;
 		d[0].chunk->size = cast->simple_frame.size;
 		d[0].chunk->stride = cast->simple_frame.stride;
 		d[0].chunk->offset = 0;
-		d[0].flags = 0;
-		d[0].fd = -1;
+		d[0].flags = SPA_DATA_FLAG_READWRITE;
```
emersion, line 132: Shouldn't we make the buffer read-only for our consumers?

src/screencast/pipewire_screencast.c
```diff
@@ -170,28 +172,59 @@ static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
...
+		d[0].fd = anonymous_shm_open();
+
+		if (d[0].fd == -1) {
+			logprint(ERROR, "pipewire: unable to create anonymous filedescriptor");
+			return;
+		}
+
+		if (ftruncate(d[0].fd, d[0].maxsize) < 0) {
+			logprint(ERROR, "pipewire: unable to truncate filedescriptor");
```
emersion, line 141: We should `close` the fd and reset it to -1 here

src/screencast/pipewire_screencast.c
```diff
@@ -170,28 +172,59 @@ static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
...
+		d[0].data = mmap(NULL, d[0].maxsize, PROT_READ | PROT_WRITE, MAP_SHARED, d[0].fd, d[0].mapoffset);
+		if (d[0].data == MAP_FAILED) {
+			logprint(ERROR, "pipewire: unable to mmap memory");
+			return;
+		}
 	}
 }
 
 static void pwr_handle_stream_remove_buffer(void *data, struct pw_buffer *buffer) {
+	struct spa_data *d;
```
emersion, line 152: Style: this can be declared below

emersion:
I've looked at the commits from "screencast: add logging to pwr_handle_stream_param_changed in pipewire_sreencast.c" to "screencast: pipewire change to self allocation of shm buffers".

columbarius:
> In "screencast: pipewire remove uneeded PW_STREAM_FLAG_MAP_BUFFERS flag": why is this flag unneeded?

Maybe unneeded is the wrong word. A commit later do we map our buffers directly after allocating it.
`PW_STREAM_FLAG_MAP_BUFFERS` indicates, that pipewire should mmap any buffer to `d[0].data` which we don't need then anymore. I think the best solution would be to just squash those two commits together.

src/screencast/pipewire_screencast.c
columbarius: Do we care about another client writing in our buffer stuff? I think it should be overwritten by the next screencopy. 
This might help with filter, which want to avoid the overhead of copying every buffer to another one. But we can just set no flag (I think i took that just from the pipewire example).

columbarius:
Thanks for the review!

src/screencast/pipewire_screencast.c
emersion: I'd prefer to avoid setting the buffer as read-write, to allow future optimizations like handling out the same frames to multiple clients.

emersion:
Hm, it seems like I missed #61 and I've reviewed here commits that belong to that other PR. It doesn't seem like #61 has the updated commits. Can you maybe replace #61's commits with the first 5 commits of this PR?

columbarius:
I think i just did that right now

src/screencast/pipewire_screencast.c
columbarius: done

src/screencast/pipewire_screencast.c
columbarius: done

src/screencast/pipewire_screencast.c
columbarius: done

src/screencast/pipewire_screencast.c
columbarius: done

Hubro:
[As requested by columbarius](https://github.com/emersion/xdg-desktop-portal-wlr/issues/154#issuecomment-873896107), here is the TRACE output from xdg-desktop-portal-wlr crashing as soon as I try to start capturing:

```
➜ ./build/xdg-desktop-portal-wlr -r -l TRACE
2021/07/06 00:48:25 [TRACE] - config: trying config file /home/tomas/.config/xdg-desktop-portal-wlr/sway
2021/07/06 00:48:25 [TRACE] - config: trying config file /home/tomas/.config/xdg-desktop-portal-wlr/config
2021/07/06 00:48:25 [TRACE] - config: trying config file /usr/local/etc/xdg/xdg-desktop-portal-wlr/sway
2021/07/06 00:48:25 [TRACE] - config: trying config file /usr/local/etc/xdg/xdg-desktop-portal-wlr/config
2021/07/06 00:48:25 [ERROR] - config: no config file found
2021/07/06 00:48:25 [DEBUG] - config: outputname:  (null)
2021/07/06 00:48:25 [DEBUG] - config: max_fps:  0.000000
2021/07/06 00:48:25 [DEBUG] - config: exec_before:  (null)
2021/07/06 00:48:25 [DEBUG] - config: exec_after:  (null)
2021/07/06 00:48:25 [DEBUG] - config: chooser_cmd: (null)
2021/07/06 00:48:25 [DEBUG] - config: chooser_type: default
2021/07/06 00:48:25 [DEBUG] - dbus: connected
2021/07/06 00:48:25 [DEBUG] - wlroots: wl_display connected
2021/07/06 00:48:25 [DEBUG] - pipewire: pw_loop created
2021/07/06 00:48:25 [DEBUG] - pipewire: establishing connection to core
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register wl_shm  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: |-- registered to interface wl_shm (Version 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register wl_drm  (Version: 2)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwp_linux_dmabuf_v1  (Version: 3)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register wl_compositor  (Version: 4)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register wl_subcompositor  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register wl_data_device_manager  (Version: 3)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwlr_gamma_control_manager_v1  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zxdg_output_manager_v1  (Version: 3)
2021/07/06 00:48:25 [DEBUG] - wlroots: |-- registered to interface zxdg_output_manager_v1 (Version 3)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register org_kde_kwin_idle  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwp_idle_inhibit_manager_v1  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwlr_layer_shell_v1  (Version: 4)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register xdg_wm_base  (Version: 2)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwp_tablet_manager_v2  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register org_kde_kwin_server_decoration_manager  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zxdg_decoration_manager_v1  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwp_relative_pointer_manager_v1  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwp_pointer_constraints_v1  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register wp_presentation  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwlr_output_manager_v1  (Version: 2)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwlr_output_power_manager_v1  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwp_input_method_manager_v2  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwp_text_input_manager_v3  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwlr_foreign_toplevel_manager_v1  (Version: 3)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwlr_export_dmabuf_manager_v1  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwlr_screencopy_manager_v1  (Version: 3)
2021/07/06 00:48:25 [DEBUG] - wlroots: |-- registered to interface zwlr_screencopy_manager_v1 (Version 3)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwlr_data_control_manager_v1  (Version: 2)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwp_primary_selection_device_manager_v1  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register wp_viewporter  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zxdg_exporter_v1  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zxdg_importer_v1  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zxdg_exporter_v2  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zxdg_importer_v2  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwp_virtual_keyboard_manager_v1  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwlr_virtual_pointer_manager_v1  (Version: 2)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwlr_input_inhibit_manager_v1  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwp_keyboard_shortcuts_inhibit_manager_v1  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register wl_seat  (Version: 7)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register zwp_pointer_gestures_v1  (Version: 1)
2021/07/06 00:48:25 [DEBUG] - wlroots: interface to register wl_output  (Version: 3)
2021/07/06 00:48:25 [DEBUG] - wlroots: |-- registered to interface wl_output (Version 1)
2021/07/06 00:48:25 [DEBUG] - wayland: registry listeners run
2021/07/06 00:48:25 [DEBUG] - wayland: xdg output listeners run
2021/07/06 00:48:25 [TRACE] - event-loop: got pipewire event
2021/07/06 00:48:25 [TRACE] - event-loop: got dbus event
2021/07/06 00:48:25 [TRACE] - event-loop: got dbus event
2021/07/06 00:48:25 [TRACE] - event-loop: got pipewire event
2021/07/06 00:48:29 [TRACE] - event-loop: got dbus event
2021/07/06 00:48:31 [TRACE] - event-loop: got dbus event
2021/07/06 00:48:31 [TRACE] - event-loop: got dbus event
2021/07/06 00:48:31 [TRACE] - event-loop: got dbus event
2021/07/06 00:48:31 [INFO] - dbus: create session method invoked
2021/07/06 00:48:31 [INFO] - dbus: request_handle: /org/freedesktop/portal/desktop/request/1_2541/obs1
2021/07/06 00:48:31 [INFO] - dbus: session_handle: /org/freedesktop/portal/desktop/session/1_2541/obs1
2021/07/06 00:48:31 [INFO] - dbus: app_id:
2021/07/06 00:48:31 [TRACE] - event-loop: got dbus event
2021/07/06 00:48:31 [TRACE] - event-loop: got dbus event
2021/07/06 00:48:31 [INFO] - dbus: select sources method invoked
2021/07/06 00:48:31 [INFO] - dbus: request_handle: /org/freedesktop/portal/desktop/request/1_2541/obs2
2021/07/06 00:48:31 [INFO] - dbus: session_handle: /org/freedesktop/portal/desktop/session/1_2541/obs1
2021/07/06 00:48:31 [INFO] - dbus: app_id:
2021/07/06 00:48:31 [INFO] - dbus: option types:1
2021/07/06 00:48:31 [INFO] - dbus: option multiple: 0
2021/07/06 00:48:31 [INFO] - dbus: option cursor_mode:2
2021/07/06 00:48:31 [DEBUG] - dbus: select sources: found matching session /org/freedesktop/portal/desktop/session/1_2541/obs1
2021/07/06 00:48:31 [INFO] - wlroots: capturable output: Samsung Electric Company model: LC49G95T: id: 43 name: DP-1
2021/07/06 00:48:31 [DEBUG] - wlroots: output chooser called
2021/07/06 00:48:31 [DEBUG] - wlroots: output chooser called
2021/07/06 00:48:31 [TRACE] - exec chooser called: cmd slurp -f %o -or, pipe chooser_in (14,15), pipe chooser_out (16,17)
2021/07/06 00:48:32 [TRACE] - wlroots: output chooser slurp -f %o -or selects output DP-1
2021/07/06 00:48:32 [DEBUG] - wlroots: output chooser selects DP-1
2021/07/06 00:48:32 [INFO] - xdpw: screencast instance 0x5613246055f0 has 1 references
2021/07/06 00:48:32 [INFO] - xdpw: 1 active screencast instances
2021/07/06 00:48:32 [INFO] - wlroots: output: DP-1
2021/07/06 00:48:32 [TRACE] - event-loop: got dbus event
2021/07/06 00:48:32 [TRACE] - event-loop: got dbus event
2021/07/06 00:48:32 [INFO] - dbus: start method invoked
2021/07/06 00:48:32 [INFO] - dbus: request_handle: /org/freedesktop/portal/desktop/request/1_2541/obs3
2021/07/06 00:48:32 [INFO] - dbus: session_handle: /org/freedesktop/portal/desktop/session/1_2541/obs1
2021/07/06 00:48:32 [INFO] - dbus: app_id:
2021/07/06 00:48:32 [INFO] - dbus: parent_window:
2021/07/06 00:48:32 [DEBUG] - dbus: start: found matching session /org/freedesktop/portal/desktop/session/1_2541/obs1
2021/07/06 00:48:32 [TRACE] - wlroots: callbacks registered
2021/07/06 00:48:32 [TRACE] - wlroots: buffer event handler
2021/07/06 00:48:32 [TRACE] - wlroots: linux_dmabuf event handler
2021/07/06 00:48:32 [TRACE] - wlroots: buffer_done event handler
2021/07/06 00:48:32 [WARN] - wlroots: failed to import buffer
2021/07/06 00:48:32 [TRACE] - wlroots: frame destroyed
2021/07/06 00:48:32 [INFO] - pipewire: stream state changed to "connecting"
2021/07/06 00:48:32 [INFO] - pipewire: node id is -1
2021/07/06 00:48:32 [TRACE] - event-loop: got pipewire event
2021/07/06 00:48:32 [TRACE] - event-loop: got pipewire event
2021/07/06 00:48:32 [TRACE] - event-loop: got pipewire event
2021/07/06 00:48:32 [INFO] - pipewire: stream state changed to "paused"
2021/07/06 00:48:32 [INFO] - pipewire: node id is 42
2021/07/06 00:48:32 [TRACE] - event-loop: got pipewire event
2021/07/06 00:48:32 [TRACE] - pipewire: stream parameters changed
2021/07/06 00:48:32 [TRACE] - event-loop: got pipewire event
2021/07/06 00:48:32 [TRACE] - pipewire: add buffer event handle
2021/07/06 00:48:32 [TRACE] - pipewire: selected buffertype 2
2021/07/06 00:48:32 [TRACE] - event-loop: got pipewire event
2021/07/06 00:48:32 [TRACE] - event-loop: got pipewire event
2021/07/06 00:48:32 [INFO] - pipewire: stream state changed to "streaming"
2021/07/06 00:48:32 [INFO] - pipewire: node id is 42
2021/07/06 00:48:32 [TRACE] - wlroots: callbacks registered
2021/07/06 00:48:32 [TRACE] - event-loop: got pipewire event
2021/07/06 00:48:32 [TRACE] - event-loop: got wayland event
2021/07/06 00:48:32 [TRACE] - wlroots: buffer event handler
2021/07/06 00:48:32 [TRACE] - wlroots: linux_dmabuf event handler
2021/07/06 00:48:32 [TRACE] - wlroots: buffer_done event handler
2021/07/06 00:48:32 [TRACE] - pipewire: importing buffer
2021/07/06 00:48:32 [DEBUG] - wlroots: pipewire and wlroots metadata are incompatible. Renegotiate stream
2021/07/06 00:48:32 [TRACE] - pipewire: exporting buffer
2021/07/06 00:48:32 [TRACE] - ********************
2021/07/06 00:48:32 [TRACE] - pipewire: pointer 0x56132460ab80
2021/07/06 00:48:32 [TRACE] - pipewire: size 29491200
2021/07/06 00:48:32 [TRACE] - pipewire: stride 20480
2021/07/06 00:48:32 [TRACE] - pipewire: width 5120
2021/07/06 00:48:32 [TRACE] - pipewire: height 1440
2021/07/06 00:48:32 [TRACE] - pipewire: y_invert 0
2021/07/06 00:48:32 [TRACE] - pipewire: damage 0,0 (0x0)
2021/07/06 00:48:32 [TRACE] - pipewire: crop 0,0 (0x0)
2021/07/06 00:48:32 [TRACE] - ********************
2021/07/06 00:48:32 [TRACE] - pipewire: stream update parameters
2021/07/06 00:48:32 [TRACE] - wlroots: frame destroyed
xdg-desktop-portal-wlr: ../src/screencast/fps_limit.c:27: fps_limit_measure_end: Assertion `!timespec_is_zero(&state->frame_last_time)' failed.
zsh: abort (core dumped)  ./build/xdg-desktop-portal-wlr -r -l TRACE
```

columbarius:
Fixed the pipewire metadata check, so we don't renegotiate when  the alpha is striped and i'm only using the fps limiter if a frame was copied before (which means that the measurement was started). We probably want to check for a more reasonable logic here, see https://github.com/emersion/xdg-desktop-portal-wlr/pull/141/commits/a4485f00ed5d994d4836ab8043ea0d51a04656b4.

columbarius:
@Hubro 
thanks for your trace. What confuses me is, why the xdpw wants to renegotiate the stream on the first frame:
>```
>2021/07/06 00:48:32 [INFO] - pipewire: stream state changed to "streaming"
>2021/07/06 00:48:32 [INFO] - pipewire: node id is 42
>2021/07/06 00:48:32 [TRACE] - wlroots: callbacks registered
>2021/07/06 00:48:32 [TRACE] - event-loop: got pipewire event
>2021/07/06 00:48:32 [TRACE] - event-loop: got wayland event
>2021/07/06 00:48:32 [TRACE] - wlroots: buffer event handler
>2021/07/06 00:48:32 [TRACE] - wlroots: linux_dmabuf event handler
>2021/07/06 00:48:32 [TRACE] - wlroots: buffer_done event handler
>2021/07/06 00:48:32 [TRACE] - pipewire: importing buffer
>2021/07/06 00:48:32 [DEBUG] - wlroots: pipewire and wlroots metadata are incompatible. Renegotiate stream
>2021/07/06 00:48:32 [TRACE] - pipewire: exporting buffer
>2021/07/06 00:48:32 [TRACE] - ********************
>```

We do announce the format together with the corresponding format without alpha because of a webrtc issue, but obs should be able to handle the version with alpha afaicr. 
The screencast fails, because it's leaving the loop before the first frame was copied so the fps_limiter was not engaged and fails when it should stop measuring. I added a check if it was engaged so we should not trigger that assume. Could you please try #156 again?

columbarius:
@emersion @jbeich Is the build failing inside spa (pipewire) on our side or a freebsd issue?

emersion:
This _seems_ like a FreeBSD issue…

jbeich:
> This _seems_ like a FreeBSD issue…

See https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=256970

emersion:
Ah, thanks! (Though it's quite ironic that trying to be portable brings us portability issues…)

columbarius:
thanks

columbarius:
improved commit messages

src/screencast/wlr_screencast.c
```diff
@@ -197,6 +202,8 @@ static void wlr_frame_failed(void *data,
 	logprint(TRACE, "wlroots: failed event handler");
 	cast->err = true;
 
+	xdpw_pwr_export_buffer(cast);
```
emersion, line 156: At best, this will copy a bad frame in `writeFrameData`. Instead we should not perform the `memcpy`. I think it would be nice to set the `d[0].chunk->flags = SPA_CHUNK_FLAG_CORRUPTED` to indicate the buffer we enqueue to PipeWire has no meaningful contents (e.g. the v4l2 plugin does this, see `spa/plugins/v4l2/v4l2-utils.c`).

include/screencast_common.h
```diff
@@ -123,6 +123,8 @@ struct xdpw_wlr_output {
 
 void randname(char *buf);
 int anonymous_shm_open(void);
+struct wl_buffer *import_shm_buffer(struct xdpw_screencast_instance *cast, int fd,
```
emersion, line 132: Maybe rename to `import_wl_shm_buffer` to make this clear this imports the shm buffer into the Wayland compositor?

include/screencast_common.h
```diff
@@ -83,6 +94,7 @@ struct xdpw_screencast_instance {
 	uint32_t refcount;
 	struct xdpw_screencast_context *ctx;
 	bool initialized;
+	struct xdpw_frame simple_frame;
```
emersion, line 91: Why is this named "simple"?

src/screencast/wlr_screencast.c
```diff
@@ -128,13 +112,38 @@ static void wlr_frame_buffer_done(void *data,
...
+
+	if (!cast->simple_frame.current_pw_buffer) {
+		logprint(WARN, "wlroots: failed to import buffer");
+		xdpw_wlr_frame_free(cast);
+		return;
+	}
+
+	// Check if announced screencopy information is compatible with pipewire meta
+	if (!(cast->pwr_format.format == xdpw_format_pw_from_wl_shm(cast->screencopy_frame.format) ||
+		cast->pwr_format.format == xdpw_format_pw_strip_alpha(xdpw_format_pw_from_wl_shm(cast->screencopy_frame.format))) ||
```
emersion, line 75: Style: missing indent level for `if` continuation line

src/screencast/wlr_screencast.c
```diff
@@ -128,13 +112,38 @@ static void wlr_frame_buffer_done(void *data,
...
+		if (cast->pwr_stream_state) {
+			pwr_update_stream_param(cast);
 		}
+		xdpw_wlr_frame_free(cast);
+		return;
+	}
+
+	// Check if imported buffer is compatible with announced buffer
+	if (cast->simple_frame.size != cast->screencopy_frame.size ||
+		cast->simple_frame.stride != cast->screencopy_frame.stride) {
```
emersion, line 90: Style: missing indent level for `if` continuation line

src/screencast/wlr_screencast.c
```diff
@@ -128,13 +112,38 @@ static void wlr_frame_buffer_done(void *data,
...
+	}
+
+	if (!cast->simple_frame.current_pw_buffer) {
+		logprint(WARN, "wlroots: failed to import buffer");
+		xdpw_wlr_frame_free(cast);
+		return;
+	}
+
+	// Check if announced screencopy information is compatible with pipewire meta
+	if (!(cast->pwr_format.format == xdpw_format_pw_from_wl_shm(cast->screencopy_frame.format) ||
```
emersion, line 74: This can be simplified:

```c
if (!(A == B || C == D))
```

→

```c
if (A != B && C != D)
```

src/screencast/pipewire_screencast.c
```diff
@@ -155,13 +139,9 @@ static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
...
 
-		// mmap buffer, so we can use the data_ptr in on_process
-		d[0].data = mmap(NULL, d[0].maxsize, PROT_READ | PROT_WRITE, MAP_SHARED, d[0].fd, d[0].mapoffset);
-		if (d[0].data == MAP_FAILED) {
-			logprint(ERROR, "pipewire: unable to mmap memory");
-			cast->err = 1;
-			return;
-		}
+		// create wl_buffer
+		d[0].data = import_shm_buffer(cast, d[0].fd, cast->screencopy_frame.format,
```
emersion, line 143: Hmmmm. `d[0].data` is supposed to be raw pixel data, not a `wl_buffer`.

Instead we should use `pw_buffer.user_data` I think?

src/screencast/pipewire_screencast.c
```diff
@@ -230,13 +212,10 @@ void xdpw_pwr_export_buffer(struct xdpw_screencast_instance *cast) {
...
 	}
-	if ((d[0].data) == NULL) {
-		logprint(TRACE, "pipewire: data pointer undefined");
-		goto queue;
-	}
 
-	writeFrameData(d[0].data, cast->screencopy_frame.data, cast->screencopy_frame.height,
-		cast->screencopy_frame.stride, cast->screencopy_frame.y_invert);
+	if (cast->screencopy_frame.y_invert) {
+		//TODO: Flip buffer or set stride negative
```
emersion, line 217: Error out in the meantime?

src/screencast/wlr_screencast.c
```diff
@@ -146,7 +89,12 @@ static void wlr_frame_buffer_done(void *data,
 		return;
 	}
 
-	zwlr_screencopy_frame_v1_copy_with_damage(frame, cast->screencopy_frame.buffer);
+	if (cast->simple_frame.buffer == NULL) {
+		logprint(ERROR, "wlroots: imported buffer doesn't contain a wlr_buffer");
+		abort();
```
emersion, line 99: This can be turned into an `assert`.

emersion (commented):
Thanks for the improved commit messages. Here's a first round of comment.

src/screencast/pipewire_screencast.c
columbarius: could you please link me the header? I can't find a user_data.

emersion: In `src/pipewire/stream.h`:

```c
struct pw_buffer {
	struct spa_buffer *buffer;	/**< the spa buffer */
	void *user_data;		/**< user data attached to the buffer */
	uint64_t size;			/**< This field is set by the user and the sum of
					  *  all queued buffer is returned in the time info */
};
```

The docs are at https://docs.pipewire.org/structpw__buffer.html but are down atm.

columbarius: thanks!

columbarius:
Fixed the style issues, but didn't added all asserts, since importing a buffer from pipewire before requesting the one from wlroots (fxing #122 ) will change that.
Note: The last commit is not finished, we don't receive any pipewire events about new buffers.

columbarius:
Added both an event and the process callback

columbarius:
@emersion @danshick may one of you check what i did wrong in comparison with https://github.com/Plagman/gamescope/commit/ef78e2e2707a75b6660809445fab578959a1548e? 

Additionally there is a new pw_stream_trigger_process method [previously named _drive](https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/9831f62927f077ddbd68e0ebd6f193876ca94546) which could be usefull for that.

columbarius:
using pipewire events works pretty good until we can use pw_stream_drive.

src/screencast/pipewire_screencast.c
```diff
@@ -231,6 +231,66 @@ static const struct pw_stream_events pwr_stream_events = {
...
+	struct pw_buffer *pw_buf;
+	struct spa_buffer *spa_buf;
+	struct spa_meta_header *h;
+	struct spa_data *d;
+
+	logprint(TRACE, "pipewire: exporting buffer");
+
+	pw_buf = cast->current_pw_buffer;
+
+	if (!pw_buf) {
```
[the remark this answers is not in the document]
columbarius: The idea behind that was to ensure, that running the screencopy buffer_done method while the pipewire stream was not established didn't made a problem, but it would probably be better to catch those cases beforehand.

include/screencast_common.h
columbarius: Because the old one was. Renamed it to current_frame, another good pick might be working_frame. This struct is the central part containing all information, which has to be transferred between wlroots and pipewire.

src/screencast/pipewire_screencast.c
columbarius: Added a naive implementation to reinvert the buffer.

src/screencast/wlr_screencast.c
columbarius: changed

src/screencast/pipewire_screencast.c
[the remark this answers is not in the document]
columbarius: rework took care of that and an assert was added.

src/screencast/wlr_screencast.c
columbarius: flag was added, thanks for noticing! Needs to be added to clients to be respected.

columbarius:
Reworked the wlroots screencopy loop, to have a clear start and end point, where the frame is created and the pipewire buffer is de and enqueued. Does this looks more reasonable?

columbarius:
Updated the MR again to use the pipewire `process` callback to start the stream.

Ready for a new review round from my side. We probably want to squash some commits if the reworked loop is fine.

src/screencast/pipewire_screencast.c
```diff
@@ -231,6 +232,60 @@ static const struct pw_stream_events pwr_stream_events = {
...
+	struct spa_buffer *spa_buf = pw_buf->buffer;
+	struct spa_data *d = spa_buf->datas;
+	struct spa_meta_header *h;
+	if ((h = spa_buffer_find_meta_data(spa_buf, SPA_META_Header, sizeof(*h)))) {
+		h->pts = -1;
+		h->flags = 0;
+		h->seq = cast->seq++;
+		h->dts_offset = 0;
+	}
+	if ((d[0].data) == NULL) {
```
emersion: Style nit: unnecessary parentheses

src/screencast/wlr_screencast.c
```diff
@@ -127,6 +127,14 @@ static void wlr_frame_buffer_done(void *data,
 
 	logprint(TRACE, "wlroots: buffer_done event handler");
 
+	if (!cast->quit && !cast->err && cast->pwr_stream_state) {
+		xdpw_pwr_import_buffer(cast);
+		if(!cast->current_pw_buffer) {
```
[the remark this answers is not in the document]
emersion: This hasn't been addressed.

src/screencast/pipewire_screencast.c
```diff
@@ -141,6 +125,7 @@ static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
 		d[0].chunk->offset = 0;
 		d[0].flags = 0;
 		d[0].fd = anonymous_shm_open();
+		d[0].data = 0;
```
emersion, line 165: Nit: this is a pointer, so should be set to NULL

src/screencast/wlr_screencast.c
```diff
@@ -34,12 +34,17 @@ void xdpw_wlr_frame_free(struct xdpw_screencast_instance *cast) {
...
+			cast->copied = false;
+			if (delay_ns > 0) {
+				xdpw_add_timer(cast->ctx->state, delay_ns,
+					(xdpw_event_loop_timer_func_t) xdpw_wlr_register_cb, cast);
+			} else {
+				xdpw_wlr_register_cb(cast);
+			}
 		} else {
-			xdpw_wlr_register_cb(cast);
+				xdpw_wlr_register_cb(cast);
```
emersion: Style: this has an extra indent

include/screencast_common.h
```diff
@@ -91,13 +91,15 @@ struct xdpw_screencast_instance {
 	struct xdpw_frame current_frame;
 
 	// pipewire
+	struct spa_source *event;
```
emersion: I don't think `spa_source` should be used for this. Instead, we can use `pw_stream_events.process`: it's fired when a PipeWire buffer can be dequeued.

emersion:
Commits up to "screencast: only end the fps measurement when it was started before" look good to me apart from the style issues and nits.

src/screencast/wlr_screencast.c
```diff
@@ -65,6 +65,27 @@ void xdpw_wlr_frame_start(struct xdpw_screencast_instance *cast) {
...
+	cast->wlr_frame = frame;
+
+	cast->screencopy_frame.width = width;
+	cast->screencopy_frame.height = height;
+	cast->screencopy_frame.stride = stride;
+	cast->screencopy_frame.size = stride * height;
+	cast->screencopy_frame.format = format;
+
+	if (zwlr_screencopy_manager_v1_get_version(cast->ctx->screencopy_manager) < 3) {
+		wlr_frame_buffer_done(cast,frame);
```
emersion, line 103: Style: missing space after comma

src/screencast/pipewire_screencast.c
```diff
@@ -219,6 +219,12 @@ void xdpw_pwr_enqueue_buffer(struct xdpw_screencast_instance *cast) {
 		h->dts_offset = 0;
 	}
 
+	if (cast->frame_state == XDPW_FRAME_STATE_SUCCESS) {
+		d[0].chunk->flags = SPA_CHUNK_FLAG_NONE;
+	} else {
+		d[0].chunk->flags = SPA_CHUNK_FLAG_CORRUPTED;
```
emersion, line 257: Sweet. Can we mark the `spa_meta_header` as corrupted as well with `SPA_META_HEADER_FLAG_CORRUPTED`?

src/screencast/pipewire_screencast.c
```diff
@@ -13,6 +13,35 @@
...
+	uint32_t height = d->chunk->size / d->chunk->stride;
+	if (height%2) {
+		N = (height-1)/2;
+	} else {
+		N = height/2;
+	}
+	for (uint32_t i = 0; i < N; i++) {
+		memcpy(tmp_buffer, (data + i * d->chunk->stride), d->chunk->stride);
+		memcpy((data + i * d->chunk->stride), (data + (height - 1 - i) * d->chunk->stride), d->chunk->stride);
+		memcpy((data + (height - 1 - i) * d->chunk->stride), tmp_buffer, d->chunk->stride);
```
emersion, line 38: There are extra parens in here

src/screencast/pipewire_screencast.c
```diff
@@ -13,6 +13,35 @@
...
+static bool pwr_y_invert_buffer(struct pw_buffer *buffer) {
+	uint32_t N;
+	struct spa_data *d = buffer->buffer->datas;
+	void *tmp_buffer = malloc(d->chunk->stride);
+	if (!tmp_buffer) {
+		return false;
+	}
+	void *data = mmap(NULL, d[0].maxsize, PROT_READ | PROT_WRITE, MAP_SHARED,
+			d[0].fd, d[0].mapoffset);
+	if (!data) {
```
emersion, line 25: Should check for `MAP_FAILED` here

src/screencast/pipewire_screencast.c
```diff
@@ -13,6 +13,35 @@
...
+		free(tmp_buffer);
+		return false;
+	}
+	uint32_t height = d->chunk->size / d->chunk->stride;
```
emersion, line 29: There's no guarantee the height will be divisible by the stride, and that the result is the height…

src/screencast/pipewire_screencast.c
```diff
@@ -13,6 +13,35 @@
 #include "xdpw.h"
 #include "logger.h"
 
+static bool pwr_y_invert_buffer(struct pw_buffer *buffer) {
```
emersion, line 16: TBH, I'd prefer to error out when receiving y-inverted buffer, at least for now. There's no known compositor using it right now.

include/screencast_common.h
emersion: Hm, this is kind of reverted in "screencast: drive screencast without pipewire events", then changed again in "screencast: use the pipewire callback process to start screencast". Please squash together these changes.

src/screencast/pipewire_screencast.c
```diff
@@ -196,8 +204,17 @@ static const struct pw_stream_events pwr_stream_events = {
...
+	.process = pwr_handle_stream_process,
 };
 
+void xdpw_pwr_trigger_process(struct xdpw_screencast_instance *cast) {
+	pw_stream_trigger_process(cast->stream);
+}
+
+bool xdpw_pwr_is_driving(struct xdpw_screencast_instance *cast) {
+	return pw_stream_is_driving(cast->stream);
+}
```
emersion, line 216: Do we really need these trivial wrappers?

include/screencast_common.h
columbarius: > I don't think `spa_source` should be used for this. Instead, we can use `pw_stream_events.process`: it's fired when a PipeWire buffer can be dequeued.

`.process` will only be fired for all consumers when the producer enqueues a buffer, at least that was the state this summer when i wrote this. The introduction of the trigger_process call could have changed that.

columbarius:
Rebased onto master, cleaned up the nits and style, removed transforming y-inverted buffers, added dropping the current buffer on `remove_buffer` and fixed the loop on `trigger_process` for PipeWire and the `frame_state` for screencopy.

src/screencast/pipewire_screencast.c
columbarius: we need a wrapper for `pw_stream_trigger_process`, otherwise we cast a function pointer from `int (*)` to `void (*)`

columbarius: We could make `xdpw_pwr_trigger_process` a noop if we aren't the driver (removing the `xdpw_pwr_is_driving` wrapper), but i'm not happy with waiting for a noop.

columbarius: or we put the whole fps_limit logic into the `.process` callback so we limit the fps even when we are not the driver. But that's sth. for a follow up imho.

emersion (approved):
TBH not sure I'll have time to review this in-depth in the foreseeable future. From my previous review it looked good from a high level. Not sure the state enum is 100% necessary but it's not a big deal.

Feel free to merge away. We can always refine the approach in future PRs.

columbarius:
rebased

src/screencast/pipewire_screencast.c
```diff
@@ -72,6 +72,9 @@ static void pwr_handle_stream_state_changed(void *data,
 	switch (state) {
 	case PW_STREAM_STATE_STREAMING:
 		cast->pwr_stream_state = true;
+		if (!cast->wlr_frame) {
+			xdpw_wlr_register_cb(cast);
+		}
```
columbarius, line 69: Checked every commit again and found a race condition only in `screencast: only restart wlroots loop if stream is active` 
when PipeWire sends a `state_changed` changing the state to `PW_STATE_STREAMING` in between the `buffer_done` and the `ready`/`failed` events  of the screencopy loop. This did trigger a new screencopy loop entrance while the old wasn't finished and triggered an assertion in `xdpw_pwr_dequeue_buffer`.

Checking if **NO** screencopy loop is currently running via `cast->wlr_frame == NULL` solved this issue.

columbarius:
This doesn't affect anything else, so ready to merge.

columbarius:
Not really. This MR is the preparation for [hardware buffer support (DmaBuf)](#152). The [cropping branch](#156) is blocked on adding cropping support on the chooser. I will write there what's open if anyone want's to pick it up.

Hubro:
> @Hubro Not really. This MR is the preparation for [hardware buffer support (DmaBuf)](#152). The [cropping branch](#156) is blocked on adding cropping support on the chooser. I will write there what's open if anyone want's to pick it up.

Is it an option to add it as xdg-desktop-portal-wlr specific configuration meanwhile?

emersion:
I'd rather not add temporary config options that we'll need to remove afterwards.

columbarius:
> Is it an option to add it as xdg-desktop-portal-wlr specific configuration meanwhile?

I could add one in the patchset, but it won't be merged until all things are sorted out.
Can we please move the discussion over to #156?

PR #142, opened 2021-06-01
ci: drop FreeBSD workaround after 22af25749e29

jbeich:
Follow up #126. inih exposes required CFLAGS/LDFLAGS via pkg-config, so hooking `/usr/local` is no longer necessary.


emersion (approved):
Thanks!

PR #144, opened 2021-06-07
Handle POLLHUP

emersion:
This makes xdpw exit when either the D-Bus daemon, the Wayland
compositor or Pipewire is shut down. This avoids having dangling
non-functional xdpw instances, e.g. when restarting the compositor.

To test, start Sway nested in Sway, run xdpw with
WAYLAND_DISPLAY=wayland-2, and exit the nested Sway.

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -196,6 +196,19 @@ int main(int argc, char *argv[]) {
 			goto error;
 		}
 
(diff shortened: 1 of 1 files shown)


columbarius:
xdpw gets closed nicely on POLLHUP, but we still don't cleanup the streams nicely, so screencast clients can hang on the last frame. I think the cleanup can be done in another MR.

columbarius (approved):
LGTM!

PR #147, opened 2021-06-13
screencast: use negotiated maxFramerate from pipewire

columbarius:
Read the negotiated maxFramerate from pipewire and use that for our timer to prevent unnecessary screencopies. 

I don't know any usefull client, which announces maxFramerate yet, but hardcoding them in obs-studio worked.


emersion (approved):
LGTM, thanks!

columbarius:
Thanks for merging!

PR #150, opened 2021-06-15
screencast: don't rely on the structure of xdpw_screencast_instance t…

columbarius:
…o convert formats

### DIFF
--- a/include/screencast_common.h
+++ b/include/screencast_common.h
@@ -121,8 +121,7 @@ struct xdpw_wlr_output {
 };
 
 void randname(char *buf);
-enum spa_video_format xdpw_format_pw_from_wl_shm(
-	struct xdpw_screencast_instance *cast);
+enum spa_video_format xdpw_format_pw_from_wl_shm(enum wl_shm_format format);
 enum spa_video_format xdpw_format_pw_strip_alpha(enum spa_video_format format);
 
 enum xdpw_chooser_types get_chooser_type(const char *chooser_type);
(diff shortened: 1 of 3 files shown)


columbarius (commented):
Do we want to change that to SPA_VIDEO_FORMAT_UNKNOWN?

https://github.com/emersion/xdg-desktop-portal-wlr/pull/150/files#diff-7004fde2b3e368e364c57aa5dc98b5ce8e117c8a90022233b6422b978a31b3baL37

danshick:
> Do we want to change that to SPA_VIDEO_FORMAT_UNKNOWN?

What practical information does that signal to pipewire? I suppose in this case the downstream app can't make sense of it. Probably not a good reason to abruptly crash, but at the same time, it does imply we are seeing a format "in the wild" that we've never seen before. I'd like to encourage the bug reports that `abort()` brings.

I think what you're proposing is more correct but what we have may be more practical. I seriously doubt we'll ever hit this default case, so I don't mind much either way.

columbarius:
> What practical information does that signal to pipewire?

None, except the Format doesn't have a representation on the pipewire enums.

It's more that i don't like a helper function like format convertion to crash the portal, or other running streams.
I would prefer to add the abort either here https://github.com/columbarius/xdg-desktop-portal-wlr/blob/cleanup-format_pw_from_shm/src/screencast/pipewire_screencast.c#L166 or just print a critical error and end the stream instead. An error in the log is probably as usefull as a dump?

danshick:
Logging is just as useful but less likely to result in a github issue. Still, I think I agree with you, format negotiation is a bad place to crash out, but if wlroots is offering an unknown format, we're already pretty stuck.

emersion (approved):
LGTM, thanks!

columbarius:
Thanks for merging!

Issue #149, opened 2021-06-14
Segfault when using greetd + Sway


itaranto:
I'm using `greetd` as login manager with the GTK greeter and Sway.  
This setup is more or less explained [here](https://wiki.archlinux.org/title/Greetd#gtkgreet) and [here](https://man.sr.ht/~kennylevinsen/greetd/#setting-up-greetd-with-gtkgreet
).

After a specific upgrade, the greeter now takes a few seconds to start due to xdpw crashing and being restarted several times.

Here's the log with `-l DEBUG`: [xdpw.log](https://github.com/emersion/xdg-desktop-portal-wlr/files/6645012/xdpw.log)

This is what `coredumpctl info` says:
```
[I] ~ $ coredumpctl info
           PID: 3072 (xdg-desktop-por)
           UID: 976 (greeter)
           GID: 976 (greeter)
        Signal: 11 (SEGV)
     Timestamp: Sun 2021-06-13 22:02:18 -03 (14min ago)
  Command Line: /usr/lib/xdg-desktop-portal-wlr -l DEBUG
    Executable: /usr/lib/xdg-desktop-portal-wlr
 Control Group: /user.slice/user-976.slice/user@976.service/app.slice/xdg-desktop-portal-wlr.service
          Unit: user@976.service
     User Unit: xdg-desktop-portal-wlr.service
         Slice: user-976.slice
     Owner UID: 976 (greeter)
       Boot ID: a52b1ef6df664d259ee863b9aa403656
    Machine ID: 3d14cd859eed44e78935812dd176271f
      Hostname: ryzen
       Storage: /var/lib/systemd/coredump/core.xdg-desktop-por.976.a52b1ef6df664d259ee863b9aa403656.3072.1623632538000000.zst (inaccessible)
       Message: Process 3072 (xdg-desktop-por) of user 976 dumped core.

                Stack trace of thread 3072:
                #0  0x00007fe571794515 __strlen_avx2 (libc.so.6 + 0x160515)
                #1  0x000055b5f806a5c2 n/a (xdg-desktop-portal-wlr + 0x45c2)
                #2  0x00007fe57165bb25 __libc_start_main (libc.so.6 + 0x27b25)
                #3  0x000055b5f806b05e n/a (xdg-desktop-portal-wlr + 0x505e)
```

emersion:
Which version if xdpw is this?

danshick:
[...]

PR #151, opened 2021-06-17
config: handle missing HOME env variable gracefully

columbarius:
Fixes: https://github.com/emersion/xdg-desktop-portal-wlr/issues/149


emersion (approved):
LGTM!

Note, if we wanted to have a fallback, we could use `getpwuid`.

columbarius:
Thanks for merging!

PR #152, opened 2021-06-17
dmabuf support next generation

columbarius:
This MR aims to implement screensharing with dmabufs with the new workflow using pipewire buffers.

NOTE:
This is work in progress and will change frequently.

Depends: #141 
Includes: #159 
Supersedes #48 
Resolves: #9

### DIFF
--- a/include/pipewire_screencast.h
+++ b/include/pipewire_screencast.h
@@ -9,7 +9,6 @@
 
 void xdpw_pwr_dequeue_buffer(struct xdpw_screencast_instance *cast);
 void xdpw_pwr_enqueue_buffer(struct xdpw_screencast_instance *cast);
-void xdpw_pwr_swap_buffer(struct xdpw_screencast_instance *cast);
 void pwr_update_stream_param(struct xdpw_screencast_instance *cast);
 void xdpw_pwr_stream_create(struct xdpw_screencast_instance *cast);
 void xdpw_pwr_stream_destroy(struct xdpw_screencast_instance *cast);
--- a/include/screencast_common.h
+++ b/include/screencast_common.h
@@ -91,6 +91,12 @@ struct xdpw_buffer {
 	struct wl_buffer *buffer;
 };
 
+struct xdpw_format_modifier_pair {
+	struct wl_list link;
+	uint32_t fourcc;
+	uint64_t modifier;
+};
+
 struct xdpw_screencast_context {
 
 	// xdpw
@@ -107,6 +113,7 @@ struct xdpw_screencast_context {
 	struct zxdg_output_manager_v1 *xdg_output_manager;
 	struct wl_shm *shm;
 	struct zwp_linux_dmabuf_v1 *linux_dmabuf;
+	struct wl_list format_modifier_pairs;
 
 	// gbm
 	struct gbm_device *gbm;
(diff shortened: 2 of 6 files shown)


columbarius:
@emersion do we want one global gbm_device, or do we want one for each screencast_instance (trying to use the same as the display)?

columbarius:
All right, dmabuf screencasting with implicit modifiers seems to be working. Obs negotiates dmabuf sharing, chromium and the python screencast script negotiate shm copy. Firefox fails, because of a patch to the webrtc code, which looked at the time to be a good idea, but it wasn't. Will try to get it reverted.

columbarius:
We can now read the supported format modifier pairs from wlroots and can allocate buffers with arbitrary modifiers (or at least the one pipewire chooses for now). 

I added a `force_mod_linear` option to force implicit modifiers to always be linear, might be helpfull for clients, which mmap or debugging. The commit should be very easy to drop if we don't need/want that.

Note: this probably contains some errors in the last commits. I pushed it to test the format modifier pair selection.

columbarius:
fixed the format handling.

columbarius:
@emersion 
is there sth. i have to care wrt. multi gpu? like adding an option to force a specific device, or one device per screencast instance/output?

tinywrkb:
https://github.com/flathub/org.chromium.Chromium/pull/135

columbarius:
amazing! thx. for the note!

columbarius:
Since gnome and kde currently only support implicit modifiers and most clients are aiming at that target, do we want to split this MR into one implementing implicit modifiers and a followup extending it to support explicit modifiers?

columbarius:
Added support for multiplanar dmabufs.

We still need to wait for pipewire to let us fixate onto a modifier.

escalade:
I played with this PR a bit for fun, seems to work well but I've run into a 100% cpu on the process sometimes. Since it's draft maybe you aren't interested in any testing for now, but let me know if you want me to produce some logs on the issue.

columbarius:
@escalade Thanks for testing! The 100% cpu usage probably comes from us waiting for a new buffer from pipewire to become available.

Could you please report if you see a lot `out of buffer` lines in your log? 
Also I'm interesting what gpu you have and what modifiers you support. xdpw should print the list in the log on startup.

escalade:
Yes indeed:

2021/08/14 10:45:08 [WARN] - pipewire: out of buffers
2021/08/14 10:45:08 [WARN] - wlroots: failed to dequeue buffer
2021/08/14 10:45:08 [WARN] - pipewire: out of buffers
2021/08/14 10:45:08 [WARN] - wlroots: failed to dequeue buffer
2021/08/14 10:45:08 [WARN] - pipewire: out of buffers

I'm using an Intel Iris Plus 655 GPU. Latest wlroots/sway git and xdpw + this PR. How do I check supported modifiers?

columbarius:
> Yes indeed:
> 
> 2021/08/14 10:45:08 [WARN] - pipewire: out of buffers
> 2021/08/14 10:45:08 [WARN] - wlroots: failed to dequeue buffer
> 2021/08/14 10:45:08 [WARN] - pipewire: out of buffers
> 2021/08/14 10:45:08 [WARN] - wlroots: failed to dequeue buffer
> 2021/08/14 10:45:08 [WARN] - pipewire: out of buffers
> 
> I'm using an Intel Iris Plus 655 GPU. Latest wlroots/sway git and xdpw + this PR. How do I check supported modifiers?

When you start xdpw with loglevel TRACE, you should get some lines `wlroots: format <format> (<modifier>)` when starting.

columbarius:
Updated and rebased onto #159

PR #155, opened 2021-07-01
Fix potential memory corruption when reading bools

GermainZ:
As per `man sd_bus_message_read`, boolean items should be read into an int rather than a bool as the latter can cause memory corruption.

(I'm not sure if this currently causes any issues in xdpw, but I did encounter this bug in a portal I'm writing that's based on xdpw.)


emersion (approved):
LGTM, thanks!

PR #159, opened 2021-07-31
Dmabuf modifierless

columbarius:
Splits the part containing the modifierless codepaths for dmabuf screensharing from #152 

Depends: #141 
Depends: #184

### DIFF
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -44,6 +44,15 @@ static struct spa_pod *build_format(struct spa_pod_builder *b, enum spa_video_fo
 	return spa_pod_builder_pop(b, &f[0]);
 }
 
+static uint32_t build_formats(struct spa_pod_builder *b, struct xdpw_screencast_instance *cast,
+		const struct spa_pod *params[static 1]) {
+	uint32_t param_count = 1;
+	params[0] = build_format(b, xdpw_format_pw_from_drm_fourcc(cast->screencopy_frame_info.format),
+			cast->screencopy_frame_info.width, cast->screencopy_frame_info.height, cast->framerate);
+
+	return param_count;
+}
+
 static void pwr_handle_stream_on_process(void *data) {
 	logprint(TRACE, "pipewire: stream process");
 	struct xdpw_screencast_instance *cast = data;
@@ -268,12 +277,9 @@ void pwr_update_stream_param(struct xdpw_screencast_instance *cast) {
 		SPA_POD_BUILDER_INIT(params_buffer, sizeof(params_buffer));
 	const struct spa_pod *params[1];
 
-	enum spa_video_format format = xdpw_format_pw_from_drm_fourcc(cast->screencopy_frame_info.format);
+	uint32_t n_params = build_formats(&b, cast, params);
 
-	params[0] = build_format(&b, format,
-			cast->screencopy_frame_info.width, cast->screencopy_frame_info.height, cast->framerate);
-
-	pw_stream_update_params(stream, params, 1);
+	pw_stream_update_params(stream, params, n_params);
 }
 
 void xdpw_pwr_stream_create(struct xdpw_screencast_instance *cast) {
@@ -284,6 +290,7 @@ void xdpw_pwr_stream_create(struct xdpw_screencast_instance *cast) {
 
 	uint8_t buffer[1024];
 	struct spa_pod_builder b = SPA_POD_BUILDER_INIT(buffer, sizeof(buffer));
+	const struct spa_pod *params[1];
 
 	char name[] = "xdpw-stream-XXXXXX";
 	randname(name + strlen(name) - 6);
@@ -298,10 +305,7 @@ void xdpw_pwr_stream_create(struct xdpw_screencast_instance *cast) {
 	}
 	cast->pwr_stream_state = false;
 
-	enum spa_video_format format = xdpw_format_pw_from_drm_fourcc(cast->screencopy_frame_info.format);
-
-	const struct spa_pod *param = build_format(&b, format,
-			cast->screencopy_frame_info.width, cast->screencopy_frame_info.height, cast->framerate);
+	uint32_t param_count = build_formats(&b, cast, params);
 
 	pw_stream_add_listener(cast->stream, &cast->stream_listener,
 		&pwr_stream_events, cast);
@@ -311,7 +315,7 @@ void xdpw_pwr_stream_create(struct xdpw_screencast_instance *cast) {
 		PW_ID_ANY,
 		(PW_STREAM_FLAG_DRIVER |
 			PW_STREAM_FLAG_ALLOC_BUFFERS),
-		&param, 1);
+		params, param_count);
 }
 
 void xdpw_pwr_stream_destroy(struct xdpw_screencast_instance *cast) {
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -8,6 +8,7 @@
 #include <sys/mman.h>
 #include <unistd.h>
 #include <assert.h>
+#include <libdrm/drm_fourcc.h>
 
 #include "wlr_screencast.h"
 #include "xdpw.h"
@@ -35,8 +36,10 @@ static struct spa_pod *build_buffer(struct spa_pod_builder *b, uint32_t blocks, 
 }
 
 static struct spa_pod *build_format(struct spa_pod_builder *b, enum spa_video_format format,
-		uint32_t width, uint32_t height, uint32_t framerate) {
-	struct spa_pod_frame f[1];
+		uint32_t width, uint32_t height, uint32_t framerate,
+		uint64_t *modifiers, int modifier_count) {
+	struct spa_pod_frame f[2];
+	int i, c;
 
 	enum spa_video_format format_without_alpha = xdpw_format_pw_strip_alpha(format);
 
@@ -44,11 +47,30 @@ static struct spa_pod *build_format(struct spa_pod_builder *b, enum spa_video_fo
 	spa_pod_builder_add(b, SPA_FORMAT_mediaType, SPA_POD_Id(SPA_MEDIA_TYPE_video), 0);
 	spa_pod_builder_add(b, SPA_FORMAT_mediaSubtype, SPA_POD_Id(SPA_MEDIA_SUBTYPE_raw), 0);
 	/* format */
-	if (format_without_alpha == SPA_VIDEO_FORMAT_UNKNOWN) {
+	if (modifier_count > 0 || format_without_alpha == SPA_VIDEO_FORMAT_UNKNOWN) {
+		// modifiers are defined only in combinations with their format
+		// we should not announce the format without alpha
 		spa_pod_builder_add(b, SPA_FORMAT_VIDEO_format, SPA_POD_Id(format), 0);
 	} else {
 		spa_pod_builder_add(b, SPA_FORMAT_VIDEO_format,
 				SPA_POD_CHOICE_ENUM_Id(3, format, format, format_without_alpha), 0);
+	}
+	/* modifiers */
+	if (modifier_count == 1 && modifiers[0] == DRM_FORMAT_MOD_INVALID) {
+		// we only support implicit modifiers, use shortpath to skip fixation phase
+		spa_pod_builder_prop(b, SPA_FORMAT_VIDEO_modifier, SPA_POD_PROP_FLAG_MANDATORY);
+		spa_pod_builder_long(b, modifiers[0]);
+	} else if (modifier_count > 0) {
+		// build an enumeration of modifiers
+		spa_pod_builder_prop(b, SPA_FORMAT_VIDEO_modifier, SPA_POD_PROP_FLAG_MANDATORY | SPA_POD_PROP_FLAG_DONT_FIXATE);
+		spa_pod_builder_push_choice(b, &f[1], SPA_CHOICE_Enum, 0);
+		// modifiers from the array
+		for (i = 0, c = 0; i < modifier_count; i++) {
+			spa_pod_builder_long(b, modifiers[i]);
+			if (c++ == 0)
+				spa_pod_builder_long(b, modifiers[i]);
+		}
+		spa_pod_builder_pop(b, &f[1]);
 	}
 	spa_pod_builder_add(b, SPA_FORMAT_VIDEO_size,
 		SPA_POD_Rectangle(&SPA_RECTANGLE(width, height)),
@@ -69,7 +91,8 @@ static uint32_t build_formats(struct spa_pod_builder *b, struct xdpw_screencast_
 		const struct spa_pod *params[static 1]) {
 	uint32_t param_count = 1;
 	params[0] = build_format(b, xdpw_format_pw_from_drm_fourcc(cast->screencopy_frame_info[WL_SHM].format),
-			cast->screencopy_frame_info[WL_SHM].width, cast->screencopy_frame_info[WL_SHM].height, cast->framerate);
+			cast->screencopy_frame_info[WL_SHM].width, cast->screencopy_frame_info[WL_SHM].height, cast->framerate,
+			NULL, 0);
 
 	return param_count;
 }
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -233,6 +233,12 @@ static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
 	d[0].flags = 0;
 	d[0].fd = xdpw_buffer->fd;
 	d[0].data = NULL;
+
+	// clients have implemented to check chunk->size if the buffer is valid instead
+	// of using the flags. Until they are patched we should use some arbitrary value.
+	if (xdpw_buffer->buffer_type == DMABUF && d[0].chunk->size == 0) {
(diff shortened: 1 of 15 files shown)


escalade:
This coupled with the OBS EGL modifiers part 1 PR seems to work great! The issue I had with CPU load on the xdpw process is gone, and performance is excellent. Even recording 4k video from firefox/youtube is buttery smooth. Thanks for your hard work on this :)

columbarius:
@escalade Nice, did you change anything else which could reduce the cpu load?

columbarius:
rebased

columbarius:
Backported some useful features from #152 like the `build_formats` function and not stripping the alpha from the announced format when modifiers are used.

Tested and ready to merge from my side.

columbarius:
Updated the man page with a note about forcing linear modifiers and multi-gpu setups.

src/screencast/screencast_common.c
```diff
@@ -53,25 +54,75 @@ struct wl_buffer *import_wl_shm_buffer(struct xdpw_screencast_instance *cast, in
 	return buffer;
 }
 
-enum spa_video_format xdpw_format_pw_from_wl_shm(enum wl_shm_format format) {
+enum wl_shm_format xdpw_format_wl_shm_from_drm_fourcc(uint32_t format) {
+	switch (format) {
```
emersion, line 58: This can be simplified: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/bcefb71cf61b4d12465c8ba41c79ec6b651e045e/render/pixel_format.c#L146

src/screencast/wlr_screencast.c
```diff
@@ -638,6 +673,12 @@ int xdpw_wlr_screencopy_init(struct xdpw_state *state) {
 		return -1;
 	}
 
+	// make sure we have a gbm device
+	ctx->gbm = create_gbm_device();
+	if (!ctx->gbm) {
+		logprint(ERROR, "system doesn't support %s!", "gbm");
```
emersion, line 679: Nit: no need for a `%s` here

src/screencast/xdpw_gbm.c
```diff
@@ -0,0 +1,56 @@
...
+	drmDevice *devices[64];
+	char *render_node = NULL;
+
+	int n = drmGetDevices2(0, devices, sizeof(devices) / sizeof(devices[0]));
+	for (int i = 0; i < n; ++i) {
+		drmDevice *dev = devices[i];
+		if (!(dev->available_nodes & (1 << DRM_NODE_RENDER)))
+			continue;
+
+		render_node = strndup(dev->nodes[DRM_NODE_RENDER], maxlen);
```
emersion, line 20: Shouldn't need a `maxlen` param here? `strdup` should work fine.

src/screencast/xdpw_gbm.c
```diff
@@ -0,0 +1,56 @@
...
+		break;
+	}
+
+	drmFreeDevices(devices, n);
+	return render_node;
+}
+
+struct gbm_device *create_gbm_device(){
```
emersion, line 28: Style: missing space between `)` and `{`

Should use `(void)` to make sure the caller doesn't pass unexpected args.

src/screencast/xdpw_gbm.c
```diff
@@ -0,0 +1,56 @@
...
+		logprint(ERROR, "xdpw: Could not find render node");
+		return NULL;
+	} else {
+		logprint(INFO, "xdpw: Using render node %s",render_node);
+	}
+
+	int fd = open(render_node, O_RDWR);
+	if (fd < 0) {
+		logprint(ERROR, "xdpw: Could not open render node %s",render_node);
+		close(fd);
```
emersion, line 44: Should be `free(render_node)` instead

src/screencast/xdpw_gbm.c
```diff
@@ -0,0 +1,56 @@
...
+		return NULL;
+	}
+
+	free(render_node);
+	gbm = gbm_create_device(fd);
+	close(fd);
```
emersion, line 50: This is incorrect. This will close the device under GBM's nose.

src/screencast/xdpw_gbm.c
```diff
@@ -0,0 +1,56 @@
...
+
+	render_node = gbm_find_render_node(render_node_size);
+	if (render_node == NULL) {
+		logprint(ERROR, "xdpw: Could not find render node");
+		return NULL;
+	} else {
+		logprint(INFO, "xdpw: Using render node %s",render_node);
+	}
+
+	int fd = open(render_node, O_RDWR);
```
emersion, line 41: Should use `O_CLOEXEC` too to avoid leaking the FD to child processes.

src/screencast/xdpw_gbm.c
```diff
@@ -0,0 +1,56 @@
...
+		return NULL;
+	}
+
+	free(render_node);
+	gbm = gbm_create_device(fd);
+	close(fd);
+	return gbm;
+}
+
+void destroy_gbm_device(struct gbm_device *gbm) {
```
emersion, line 54: Trivial wrapper isn't needed and just adds an indirection for no good reason.

xdg-desktop-portal-wlr.5.scd
```diff
@@ -71,6 +71,12 @@ These options need to be placed under the **[screencast]** section.
 	- simple, dmenu: xdpw will launch the chooser given by **chooser_cmd**. For more details
 	  see **OUTPUT CHOOSER**.
 
+**force_mod_linear** = _bool_
```
emersion, line 74: I feel like this should be a debug option, and nothing more. Modifiers should work correctly out of the box on multi-GPU setups too.

src/screencast/screencast.c
```diff
@@ -109,6 +109,8 @@ bool setup_outputs(struct xdpw_screencast_context *ctx, struct xdpw_session *ses
 		return false;
 	}
 
+	// Disable screencast sharing to avoid sharing between dmabuf and shm capable clients
+	/*
```
emersion, line 113: Hm. Arguably it's better to fallback to a single shm stream, than have two streams (one dmabuf and one shm). Not sure.

src/screencast/pipewire_screencast.c
```diff
@@ -152,8 +235,77 @@ static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
...
 			cast->screencopy_frame.width, cast->screencopy_frame.height, cast->screencopy_frame.stride);
+		buffer->user_data = frame;
+	} else if (d[0].type == SPA_DATA_DmaBuf) {
+		assert(cast->screencopy_type == XDPW_SCREENCOPY_DMABUF);
+
+		struct xdpw_pwr_screencopy_dmabuf_frame *frame = calloc(1,sizeof(struct xdpw_pwr_screencopy_dmabuf_frame));
+
+		uint32_t flags = GBM_BO_USE_RENDERING;
+
+		if (cast->ctx->state->config->screencast_conf.force_mod_linear)
```
emersion, line 249: Style: missing braces

src/screencast/pipewire_screencast.c
```diff
@@ -152,8 +235,77 @@ static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
...
+			flags |= GBM_BO_USE_LINEAR;
+
+		frame->bo = gbm_bo_create(cast->ctx->gbm,
+				cast->screencopy_dmabuf_frame.width, cast->screencopy_dmabuf_frame.height,
+				cast->screencopy_dmabuf_frame.fourcc, flags);
+
+		if (frame->bo == NULL) {
+			logprint(ERROR, "wlroots: failed to create gbm_bo");
```
emersion, line 257: Log prefix is incorrect here

src/screencast/pipewire_screencast.c
```diff
@@ -91,14 +161,23 @@ static void pwr_handle_stream_param_changed(void *data, uint32_t id,
...
-		SPA_PARAM_BUFFERS_size,    SPA_POD_Int(cast->screencopy_frame.size),
-		SPA_PARAM_BUFFERS_stride,  SPA_POD_Int(cast->screencopy_frame.stride),
-		SPA_PARAM_BUFFERS_align,   SPA_POD_Int(XDPW_PWR_ALIGN),
-		SPA_PARAM_BUFFERS_dataType,SPA_POD_CHOICE_FLAGS_Int(1<<SPA_DATA_MemFd));
+	if (spa_pod_find_prop(param, NULL, SPA_FORMAT_VIDEO_modifier) == NULL) {
+		cast->screencopy_type = XDPW_SCREENCOPY_SHM;
+		blocks = 1;
+		size = cast->screencopy_frame.size;
+		stride = cast->screencopy_frame.stride;
+		buffertypes = (1<<SPA_DATA_MemFd);
```
emersion, line 169: Style: unnecessary parens

src/screencast/pipewire_screencast.c
```diff
@@ -91,14 +161,23 @@ static void pwr_handle_stream_param_changed(void *data, uint32_t id,
...
+	} else {
+		if (cast->pwr_format.modifier != DRM_FORMAT_MOD_INVALID)
+			abort();
+		cast->screencopy_type = XDPW_SCREENCOPY_DMABUF;
+		blocks = 1;
+		size = 0;
+		stride = 0;
+		buffertypes = (1<<SPA_DATA_DmaBuf);
```
emersion, line 177: Same here

src/screencast/pipewire_screencast.c
```diff
@@ -91,14 +161,23 @@ static void pwr_handle_stream_param_changed(void *data, uint32_t id,
...
-		SPA_PARAM_BUFFERS_align,   SPA_POD_Int(XDPW_PWR_ALIGN),
-		SPA_PARAM_BUFFERS_dataType,SPA_POD_CHOICE_FLAGS_Int(1<<SPA_DATA_MemFd));
+	if (spa_pod_find_prop(param, NULL, SPA_FORMAT_VIDEO_modifier) == NULL) {
+		cast->screencopy_type = XDPW_SCREENCOPY_SHM;
+		blocks = 1;
+		size = cast->screencopy_frame.size;
+		stride = cast->screencopy_frame.stride;
+		buffertypes = (1<<SPA_DATA_MemFd);
+	} else {
+		if (cast->pwr_format.modifier != DRM_FORMAT_MOD_INVALID)
```
emersion, line 171: Braces are mandatory

src/screencast/pipewire_screencast.c
```diff
@@ -44,6 +86,30 @@ static struct spa_pod *build_format(struct spa_pod_builder *b, enum spa_video_fo
 	return spa_pod_builder_pop(b, &f[0]);
 }
 
+static uint32_t build_formats(struct spa_pod_builder *b, struct xdpw_screencast_instance *cast, const struct spa_pod **params) {
```
emersion, line 89: Use `const struct spa_pod *params[static 2]` to make it clearer this expects exactly 2 params.

src/screencast/pipewire_screencast.c
```diff
@@ -8,27 +9,68 @@
...
 		spa_pod_builder_add(b, SPA_FORMAT_VIDEO_format, SPA_POD_Id(format), 0);
 	} else {
 		spa_pod_builder_add(b, SPA_FORMAT_VIDEO_format,
 				SPA_POD_CHOICE_ENUM_Id(3, format, format, format_without_alpha), 0);
 	}
+	/* modifiers */
+	if (modifier_count == 1 && modifiers[0] == DRM_FORMAT_MOD_INVALID) {
+		// we only support implicit modifiers, use shortpath to skip fixation phase
+		spa_pod_builder_prop(b, SPA_FORMAT_VIDEO_modifier, SPA_POD_PROP_FLAG_MANDATORY);
+		spa_pod_builder_long(b, modifiers[0]);
```
emersion, line 61: I think we need to duplicate this line. The first one is the default, the rest is the full list of possible choices.

emersion:
How to test this? 

(Given the GBM device FD was always closed I wonder if this ever worked…?)

columbarius:
Thanks for the review!

> How to test this?

You can test this with this OBS MR: https://github.com/obsproject/obs-studio/pull/5131

Build obs with cmake (`cmake -DUNIX_STRUCTURE=0 -DBUILD_BROWSER=OFF ..`), or I can trigger the CI again in the flathub repo, which generates a link to install it as flatpak.

> (Given the GBM device FD was always closed I wonder if this ever worked…?)
Proper manipulation of the memory layout randomizer ... idk.

emersion:
Cool, will try. Can you double-check that in your testing DMA-BUFs were used, and it didn't fallback to shm?

src/screencast/screencast.c
columbarius: Last time I checked, the first stream would set the format and the second one would either work or not with that, not triggering renegotiation. Can test this again.

emersion: Ah that's a shame, if that's still the case consider opening a pipewire bug?

xdg-desktop-portal-wlr.5.scd
columbarius: This option is meant of `implicit modifier` only! Should I clarify this in the text? The other option would either be using a env-variable, or checking if multiple GPU's are available and using linear for implicit internally?

columbarius:
> Cool, will try. Can you double-check that in your testing DMA-BUFs were used, and it didn't fallback to shm?

XDPW will print `using buffer type 2/3` (3 for DMA-BUF) and OBS will print it on every import (maybe you have to build as DEBUG).

src/screencast/screencast.c
columbarius: Yeah, probably now is the time. There were some ideas to solve this. Renegotiating a existing stream has to be done in a way to not disturb the existing one, if they are incompatible. Another idea would be to negotiate a second stream, check if formats are compatible and tell PipeWire to reconnect the new stream to another old one, but I don't know the state.

src/screencast/pipewire_screencast.c
columbarius: This is the "fast" path so only one modifier and no SPA_CHOICE_Enum. Analog to announcing the supported format. I can compactify it to 
```
spa_pod_builder_add(b, SPA_FORMAT_VIDEO_modifier, SPA_POD_Long(modifiers[0]), SPA_POD_PROP_FLAG_MANDATORY);
```
.

emersion: Oh right, I see, that makes sense! Current code is fine by me.

src/screencast/xdpw_gbm.c
columbarius: Can we ignore closing the fd on shutdown with this flag?

emersion: This flag means "close on exec", it means the FD will be automatically closed when performing an exec(2) syscall such as we do when spawning the chooser. The flag doesn't change what happens on shutdown.

It's a good practice to close the FDs as necessary, that way FD leaks are less annoying to track down. Also if we ever support features like config hot-reloading, closing at the correct time avoids leaving many FDs open.

src/screencast/xdpw_gbm.c
columbarius: Would this wrapper still classify as trivial?
```
void destroy_gbm_device(struct gbm_device *gbm) {
   int fd = gbm_device_get_fd(gbm);
   gbm_device_destroy(gbm);
   close(fd);
}
```

src/screencast/xdpw_gbm.c
columbarius: Ah read it as "close on exit", which thinking about it doesn't make any sense.

src/screencast/xdpw_gbm.c
emersion: I think that's okay. But can't it just be inlined in the caller?

columbarius: I just like the symmetry of having a create and destroy function for a somewhat "opaque" resource with the compiler probably inlining it (if that is possible with different compile units/files). 
And also if I don't care about the gbm object it's only a `create_` and a `destroy_` function in wlr_screencast.c, which is easy to skip.
But anyway. I'm inlining it.

src/screencast/screencast_common.c
columbarius: Nice!

columbarius:
Gentle ping

columbarius:
Should we add in a future MR a fallback announcing "Linear modifiers" allocated with the old API for Multi-GPU setups with mixed modifier support and vulkan?

columbarius:
Update: OBS just got support for explicit modifiers!

xdg-desktop-portal-wlr.5.scd
columbarius: In case this should be demoted to a debug option, should this be done via build or env variable?

src/screencast/screencast.c
columbarius: Tested this again and even triggering renegotiation after the second stream was connected didn't end in a setting working for both streams. 
Imho doing multiple screencasts at the same time is far less common than only one per output. Even if you are using multiple plugin instances of obs, the xdpw loop will pause, when the obs scene is hidden. 

Will open an issue on the PipeWire Project

xdg-desktop-portal-wlr.5.scd
columbarius: Marked this option as experimental, so we can remove it from the config or replace it with an environment variable later.

PR #161, opened 2021-08-10
screencast: fix pipewire default id_node

columbarius:
The default id node is -1.

### DIFF
--- a/src/screencast/screencast.c
+++ b/src/screencast/screencast.c
@@ -59,6 +59,7 @@ void xdpw_screencast_instance_init(struct xdpw_screencast_context *ctx,
 	cast->framerate = cast->max_framerate;
 	cast->with_cursor = with_cursor;
 	cast->refcount = 1;
+	cast->node_id = SPA_ID_INVALID;
 	logprint(INFO, "xdpw: screencast instance %p has %d references", cast, cast->refcount);
 	wl_list_insert(&ctx->screencast_instances, &cast->link);
 	logprint(INFO, "xdpw: %d active screencast instances",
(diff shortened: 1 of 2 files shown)


src/screencast/screencast.c
```diff
@@ -419,7 +419,7 @@ static int method_screencast_start(sd_bus_message *msg, void *data,
 		start_screencast(cast);
 	}
 
-	while (cast->node_id == 0) {
+	while ((int)cast->node_id == -1) {
```
emersion, line 422: It seems like the node ID is initialized to `SPA_ID_INVALID` in PipeWire. Can we use this value instead of the cast and hardcoded int?

columbarius: Makes sense!

src/screencast/screencast.c
```diff
@@ -419,7 +419,7 @@ static int method_screencast_start(sd_bus_message *msg, void *data,
 		start_screencast(cast);
 	}
 
-	while (cast->node_id == 0) {
+	while (cast->node_id == SPA_ID_INVALID) {
```
emersion, line 422: I think this was working fine before because `cast` gets zero-initialized. But `SPA_ID_INVALID != 0`, so we now need to explicitly initialize `node_id` when we create the stream. Or else we'll never enter this loop and end up with a zero node ID.

Can youa dd this in e.g. `xdpw_pwr_stream_create`?

columbarius: Added it to `xdpw_screencast_instance_init` since the other initializations take place there.

emersion (approved):
LGTM, thanks!

columbarius:
Thanks for merging and review!

PR #162, opened 2021-08-26
Fix exec_before/after leaving a zombie process behind

junglerobba:


src/screencast/screencast.c
```diff
@@ -19,20 +19,30 @@ static const char object_path[] = "/org/freedesktop/portal/desktop";
...
-	pid_t pid = fork();
-	if (pid < 0) {
+	pid_t pid1, pid2;
+	int status;
+
+	if ((pid1 = fork())) {
+		waitpid(pid1, &status, 0);
+	} else if (pid1 == 0) {
+		if ((pid2 = fork())) {
+			exit(0);
```
emersion, line 29: We should use `_exit` here instead, or else will end up hitting issues with finalizers. Might be a good idea to replace the one in the `execvp` error path below as well.

src/screencast/screencast.c
```diff
@@ -19,20 +19,30 @@ static const char object_path[] = "/org/freedesktop/portal/desktop";
 static const char interface_name[] = "org.freedesktop.impl.portal.ScreenCast";
 
 void exec_with_shell(char *command) {
-	pid_t pid = fork();
-	if (pid < 0) {
+	pid_t pid1, pid2;
+	int status;
+
+	if ((pid1 = fork())) {
+		waitpid(pid1, &status, 0);
```
emersion, line 26: Would be a good idea to check for `waitpid` errors

src/screencast/screencast.c
```diff
@@ -19,20 +19,30 @@ static const char object_path[] = "/org/freedesktop/portal/desktop";
 static const char interface_name[] = "org.freedesktop.impl.portal.ScreenCast";
 
 void exec_with_shell(char *command) {
-	pid_t pid = fork();
-	if (pid < 0) {
+	pid_t pid1, pid2;
+	int status;
```
emersion, line 23: These can be declared below as needed

emersion:
Style nit: I personally prefer handling `fork` failures early and omitting the no-error `pid > 0` branch, as done in the original code. I don't care too much though, feel free to disregard.

emersion (approved):
LGTM, thanks!

PR #171, opened 2021-11-29
Add support for 10-bit colour formats

quantum5:
This makes PipeWire 0.3.41 the minimum required version, as the colour formats were introduced in that version.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -23,7 +23,7 @@ add_project_arguments('-DSYSCONFDIR="@0@"'.format(join_paths(prefix, sysconfdir)
 inc = include_directories('include')
 
 rt = cc.find_library('rt')
-pipewire = dependency('libpipewire-0.3', version: '>= 0.3.34')
+pipewire = dependency('libpipewire-0.3', version: '>= 0.3.41')
 wayland_client = dependency('wayland-client')
 wayland_protos = dependency('wayland-protocols', version: '>=1.14')
 iniparser = dependency('inih')
--- a/src/screencast/screencast_common.c
+++ b/src/screencast/screencast_common.c
@@ -6,6 +6,8 @@
 #include <sys/stat.h>
 #include <unistd.h>
 
+#include "logger.h"
+
 void randname(char *buf) {
 	struct timespec ts;
 	clock_gettime(CLOCK_REALTIME, &ts);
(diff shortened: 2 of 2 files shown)


emersion:
> r210 is really only WL_SHM_FORMAT_XRGB2101010, but it's better to have swapped red and blue than for the portal to crash.

Hm, no, it's better to gracefully error out in this case.

quantum5:
Changed, although it makes me sad because on my computer it's actually using `WL_SHM_FORMAT_XBGR2101010` and the portal does not work.

emersion:
There is also `GBR_10LE`, but not sure it matches…

quantum5:
We may have to get pipewire to add the format if it does not exist.

I reduced the scope of this PR so you can merge it and at least help users who have the xRGB2101010 format.

emersion (dismissed):
LGTM

> We may have to get pipewire to add the format if it does not exist.

Yup, would be nice

columbarius:
Note: Currently neither the browsers nor OBS support anything other than variants of 8bit (x)RGB.

quantum5:
Reading the conversation on https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1470, it seems that all the formats are big endian, including `SPA_VIDEO_FORMAT_r210`. Since the Wayland formats are explicit little endian, that should actually correspond to `WL_SHM_FORMAT_BGRX1010102`. Is this correct?

quantum5:
I created https://gitlab.freedesktop.org/pipewire/pipewire/-/merge_requests/1077 on the PipeWire side to add the missing colour formats.

emersion:
> it seems that all the formats are big endian, including SPA_VIDEO_FORMAT_r210. Since the Wayland formats are explicit little endian, that should actually correspond to WL_SHM_FORMAT_BGRX1010102. Is this correct?

Hm, actually, it would seem like `SPA_VIDEO_FORMAT_r210` doesn't map to any Wayland format. The BE/LE shuffling only works if each component uses exactly one byte. Using https://github.com/afrantzis/pixel-format-guide:

```
> python -m pfg describe DRM_FORMAT_BGRX1010102
Format:               DRM_FORMAT_BGRX1010102
Component data type:  UNORM
Described as:         Bytes in memory
Memory little-endian: 0                1                2                3
                      M              L M              L M              L M              L
                      R₅R₄R₃R₂R₁R₀X₁X₀ G₃G₂G₁G₀R₉R₈R₇R₆ B₁B₀G₉G₈G₇G₆G₅G₄ B₉B₈B₇B₆B₅B₄B₃B₂
Memory big-endian:    0                1                2                3
                      M              L M              L M              L M              L
                      R₅R₄R₃R₂R₁R₀X₁X₀ G₃G₂G₁G₀R₉R₈R₇R₆ B₁B₀G₉G₈G₇G₆G₅G₄ B₉B₈B₇B₆B₅B₄B₃B₂
```

I think we need to introduce explicit LE variants in PipeWire.

quantum5:
> I think we need to introduce explicit LE variants in PipeWire.

Yes, that does seem to be the case. Good thing we caught this now before we merged it. I updated the PR on the PipeWire side to reflect this.

quantum5:
I updated this PR with the formats that were just merged into PipeWire today, but since there have been no release, they builds are currently failing. I'll update again when the formats are released.

emersion:
We'll need to bump the PipeWire dep version when this happens.

quantum5:
Bumped version requirement, not sure how to FreeBSD the updated version though.

jbeich:
> Bumped version requirement, not sure how to FreeBSD the updated version though.
[...]
> meson.build:26:0: ERROR: Invalid version of dependency, need 'libpipewire-0.3' ['>= 0.3.41'] found '0.3.40'.

@arrowd, can you update pipewire to 0.3.41 on FreeBSD?

arrowd:
Aye, done.

quantum5:
Still doesn't work, do we have to wait for builds.sr.ht to come up with a new image?

emersion:
Thanks @arrowd and @jbeich!

@quantum5 Usually need to wait a few days for the mirrors to be updated etc. We aren't in a hurry anyways :)

jbeich:
> @quantum5 Usually need to wait a few days for the mirrors to be updated etc. We aren't in a hurry anyways :)

The binary package hasn't been built yet, only the port (distro recipe or source package) was updated. Package builds are completely automatic, always try to build everything (for consistency) and scheduled to start on Tue/Thu/Sat/Sun at 01:00 UTC but if the previous one didn't finish (e.g., too many queued packages) the next one is skipped. Once a build has finished syncing to mirrors takes a few hours (not days) depending on how many packages were rebuilt.

How many packages are rebuilt each time may be drastically reduced by https://github.com/freebsd/poudriere/issues/822. After that one can ask portmgr@ to schedule builds more frequently e.g., every day or twice a day.

jbeich:
Can you re-trigger CI build? FreeBSD [package build](http://beefy16.nyi.freebsd.org/build.html?mastername=130amd64-default&build=277903d23660) has finished, so [pipewire-0.3.42_1.pkg](https://pkg.freebsd.org/FreeBSD:13:amd64/latest/All/pipewire-0.3.42_1.pkg) is available.

quantum5:
Yes, it works now. Thank you!

emersion (approved):
Thanks!

PR #175, opened 2021-12-28
Implement xdpw_format_pw_strip_alpha for 10-bit colour

quantum5:
This was missed in the original PR adding support for 10-bit colour.


emersion (approved):
Thanks!

PR #176, opened 2021-12-28
screencast: destroy session if output is removed

columbarius:


src/screencast/wlr_screencast.c
```diff
@@ -586,6 +596,7 @@ static void wlr_registry_handle_remove(void *data, struct wl_registry *reg,
 	struct xdpw_screencast_context *ctx = data;
 	struct xdpw_wlr_output *output = xdpw_wlr_output_find(ctx, NULL, id);
 	if (output) {
+		wlr_close_output_attached_instances(ctx, output);
```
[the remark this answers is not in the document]
columbarius: We can, just my habbit to introduce unnecessary functions when mapping what has to be done.

columbarius:
Added logging when output is removed. This helps debugging, since it adds a reason why the instance/session was closed.

src/screencast/wlr_screencast.c
```diff
@@ -580,7 +584,19 @@ static void wlr_registry_handle_remove(void *data, struct wl_registry *reg,
...
 	struct xdpw_wlr_output *output = xdpw_wlr_output_find(ctx, NULL, id);
 	if (output) {
-		wlr_remove_output(output);
+		logprint(DEBUG, "wlroots: output removed (%s)", output->name);
+		output->removed = true;
+		bool used = false;
+		struct xdpw_screencast_instance *cast, *tmp;
+		wl_list_for_each_safe(cast, tmp, &ctx->screencast_instances, link) {
+			if (cast->target_output == output) {
+				xdpw_screencast_instance_teardown(cast);
```
any1, line 608: I think that if you just set `cast->target_output = NULL` if it matches the removed output, and then check inside the screencast instance if it has an output before creating a frame, you should be fine. But for good measure, you could also call `zwlr_screencopy_v1_frame_destroy` here on any frames that have already been created.

columbarius: > you could also call `zwlr_screencopy_v1_frame_destroy` here on any frames that have already been created.

That was to easy.... thanks a lot!

Issue #170, opened 2021-11-21
Implement screencast session restore


GeorgesStavracas:
Recently xdg-desktop-portal gained the screencast session restore feature, but it requires frontends to implement that. So xdg-desktop-portal-wlr would need to send `restore_data` back to xdg-desktop-portal when a permission mode is passed.

There is a reference implementation in xdg-desktop-portal-gnome that is merged already: https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/14

You can use libportal's mainline branch to test this feature, since it's landed support for it too.

columbarius:
Thanks for the info. Interested to implement it but it is blocked by #169.

Note:
If screencast without user interaction is required this can be achieved by using the following options.
```
chooser_type=none
output_name=<Name of output>
```

Issue #169, opened 2021-11-11
Create a more extendable chooser protocol


columbarius:
#107 #154(resp. #156) and https://github.com/flatpak/xdg-desktop-portal/pull/638 need more communication between the chooser and xdg-desktop-portal-wlr. As such we should use sth. with more structure. Maybe json based?

PR #179, opened 2022-01-16
[WIP]: Restore screencast session

columbarius:
First shot at implementing restoring screencast sessions.

Implements #170 
Blocked-by #169


columbarius:
Got the dbus stuff working!

columbarius:
Maybe this could be useful with an option to allow at least the transient mode to get rid of the annoying second query from chromium before making this available via the chooser protocol?

Algram:
@columbarius What would be needed to get this in a mergeable state? Are there any problems left or just cleanup?

columbarius:
@Algram This is a PoC for the dbus side. If you want a lite version of it so chromium won't show the dialog twice, this could be easily completed. To support restoring sessions after a restart of the application all the user facing dialogs needs to be done. This also requires a more complex chooser.

Algram:
@columbarius My problem is that I am using the OBS Pipewire capturing which currently re prompts me every time I start OBS right now.

I already tried your comment here: https://github.com/emersion/xdg-desktop-portal-wlr/issues/170#issuecomment-974891571 but this does not seem to remove the prompt.

columbarius:
> I already tried your comment here: [#170 (comment)](https://github.com/emersion/xdg-desktop-portal-wlr/issues/170#issuecomment-974891571) but this does not seem to remove the prompt.

Try restarting xdpw, or just restart your pc after changing the config. If you see still the prompt, please open a separate issue.

PR #183, opened 2022-01-20
screencast: remove check on loop_iterate

columbarius:
According to the documentation [1] pw_loop_iterate (a macro for
spa_loop_control_iterate) will return the number of dispatched fd's
which don't have any value for error handling.

[1] https://docs.pipewire.org/group__spa__loop.html#ga3bae0b32100f5752b3372a505c8e04f6

### DIFF
--- a/src/screencast/screencast.c
+++ b/src/screencast/screencast.c
@@ -432,8 +432,9 @@ static int method_screencast_start(sd_bus_message *msg, void *data,
 
 	while (cast->node_id == SPA_ID_INVALID) {
 		int ret = pw_loop_iterate(state->pw_loop, 0);
-		if (ret != 0) {
+		if (ret < 0) {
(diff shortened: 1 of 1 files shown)


emersion:
We should still check for errors with `if (pw_loop_iterate(…) < 0)` and ideally break out of the loop on error.

src/screencast/screencast.c
```diff
@@ -432,7 +432,7 @@ static int method_screencast_start(sd_bus_message *msg, void *data,
 
 	while (cast->node_id == SPA_ID_INVALID) {
 		int ret = pw_loop_iterate(state->pw_loop, 0);
-		if (ret != 0) {
+		if (ret < 0) {
 			logprint(ERROR, "pipewire_loop_iterate failed: %s", spa_strerror(ret));
```
emersion, line 436: This will loop forever on error. Can we also `return ret` to prevent this?

emersion (approved):
Thanks for the fix!

Issue #182, opened 2022-01-20
screencast: refactor to remove the wlr/pipewire split


emersion:
This split causes a weird achitecture, with these two files appearing as completely separate black-box "modules". This causes an extra complexity which could be avoided by unifying the data structures and better organizing the code.

columbarius:
Analyzing our current situation:
With #184 we are most of the time saved by the framerate limiter. If we would go directly from `_finish` to `_start`, we often reach the `buffer_done` event before the PipeWire loop (the actual pipewire loop not our events cicle) has not enough time to provide a new pw_buffer. The most time is spend between `buffer_done` requesting a buffer copy and the `frame_ready/failed` events.

So if we want the option to run without a frame limiter we have to either spend more time between enqueuing the buffer in `_finish` and dequeuing a new one while the `buffer_done` event (the dequeueing in `_start` will always fail since we didn't idle inbetween).
Another solution could be to use N+1 buffers with N buffers in the PipeWire queue and one for wlroots to render in and at an event we just exchange the fd of the pw_buffer. This has the drawback that clients can't optimize the importing/mapping which comes with fixed fd tied to the pw_buffer (if it is even possible to import a image and convert it into a texture (referencing the underlying memory), which a client can reuse everytime it recieves the same fd from us.)

PR #184, opened 2022-01-20
screencast: unmess the pipewire/wlr split

columbarius:
Using the on_process event of the PipeWire can create a race condition
when a previous started screencast hasn't finished before the next
event.

This MR is composed of following parts:
* nuke the racy on_process part
* cleanup some stuff
* intruduce xdpw_buffers 
* decouple the actual buffer (xdpw_buffer) from the pw_buffer carrier and swap the buffer from the PipeWire queue after each copy.
* Relax the swap incase the PipeWire queue is lagging behind.

Resolves: #182

### DIFF
--- a/include/pipewire_screencast.h
+++ b/include/pipewire_screencast.h
@@ -3,11 +3,13 @@
 
 #include "screencast_common.h"
 
-#define XDPW_PWR_BUFFERS 4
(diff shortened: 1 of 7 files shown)


columbarius:
This solves the current race condition quite nicely. Additionally this design won't hinder the adoption of dmabufs or ext-screencopy.

PR #192, opened 2022-03-02
Add phosh to portal's UseIn list

agx:
The compositor is phoc but XDG_CURRENT_DESKTOP has "Phosh".


columbarius:
> has "Phosh"

Your path has that lower case. Can you please confirm which is the right one?

agx:
@columbarius the `UseIn` comparison in xdg-desktop-portal is case insensitive and GNOME uses lowercase `gnome` everywhere so I used all lowercase as well. The actual `XDG_CURRENT_DESKTOP` is `Phosh`.

columbarius:
All right thanks for the info. Then LGTM

columbarius:
thanks

PR #199, opened 2022-04-18
chore(logging): change config file error to info

Algram:
This does not seem to be an error and cost me some time to realize, so I fixed it for everyone else.


emersion (approved):
LGTM, thanks!

PR #200, opened 2022-04-24
Improve Interoperability between APIs using explicit and implicit modifiers

columbarius:
When having to share buffers between modifier aware APIs and modifier unaware APIs the common denominator is to create a buffer with linear layout. This MR announces support for a linear modifier when the compositor only supports implicit modifiers using the `GBM_BO_USE_LINEAR` flag to create a buffer with linear layout.


columbarius:
wlroots now announces both linear and invalid if the GPU only supports invalid modifiers, so we just need the fallback.

PR #202, opened 2022-04-28
Add `jay` to wlr.portal UseIn list

mahkoh:
Jay [1] is a new wayland compositor that is not based on wlroots and
does not support all wlroots extensions. Therefore it would not be
appropriate to add wlroots to its XDG_CURRENT_DESKTOP list. It does
however implement all wlroots extensions necessary to run
xdg-desktop-portal-wlr.

[1] https://github.com/mahkoh/jay


mahkoh:
I'm planning to write my own portal in the near future so this PR would probably do more harm than good. For now I'll tell users to manually add jay to the list.

PR #207, opened 2022-05-05
Main device

columbarius:
Use linux_dmabuf_feedback to get the main device from the compositor and use that.

### DIFF
--- a/include/wlr_screencast.h
+++ b/include/wlr_screencast.h
@@ -12,7 +12,8 @@
 
 #define XDG_OUTPUT_MANAGER_VERSION 3
 
-#define LINUX_DMABUF_VERSION 3
+#define LINUX_DMABUF_VERSION 4
+#define LINUX_DMABUF_VERSION_MIN 3
 
 struct xdpw_state;
 
--- a/src/screencast/screencast_common.c
+++ b/src/screencast/screencast_common.c
@@ -7,7 +7,6 @@
 #include <sys/stat.h>
 #include <unistd.h>
 #include <libdrm/drm_fourcc.h>
-#include <xf86drm.h>
 #include "linux-dmabuf-unstable-v1-client-protocol.h"
 
 #include "logger.h"
@@ -23,13 +22,16 @@ void randname(char *buf) {
 	}
 }
 
-static char *gbm_find_render_node() {
+static char *gbm_find_render_node(drmDevice *device) {
 	drmDevice *devices[64];
 	char *render_node = NULL;
 
 	int n = drmGetDevices2(0, devices, sizeof(devices) / sizeof(devices[0]));
 	for (int i = 0; i < n; ++i) {
 		drmDevice *dev = devices[i];
+		if (device && !drmDevicesEqual(device, dev)) {
+				continue;
+		}
 		if (!(dev->available_nodes & (1 << DRM_NODE_RENDER)))
 			continue;
 
@@ -41,11 +43,11 @@ static char *gbm_find_render_node() {
 	return render_node;
 }
 
-struct gbm_device *xdpw_gbm_device_create(void) {
+struct gbm_device *xdpw_gbm_device_create(drmDevice *device) {
 	struct gbm_device *gbm;
 	char *render_node = NULL;
 
-	render_node = gbm_find_render_node();
+	render_node = gbm_find_render_node(device);
 	if (render_node == NULL) {
 		logprint(ERROR, "xdpw: Could not find render node");
 		return NULL;
--- a/src/screencast/wlr_screencast.c
+++ b/src/screencast/wlr_screencast.c
@@ -15,6 +15,7 @@
 #include <unistd.h>
 #include <assert.h>
 #include <wayland-client-protocol.h>
+#include <xf86drm.h>
 
 #include "screencast.h"
 #include "pipewire_screencast.h"
@@ -545,6 +546,26 @@ static void wlr_remove_output(struct xdpw_wlr_output *out) {
 	free(out);
 }
 
+static void wlr_format_modifier_pair_add(struct xdpw_screencast_context *ctx,
+		uint32_t format, uint64_t modifier) {
+	struct xdpw_format_modifier_pair *fm_pair = calloc(1, sizeof(struct xdpw_format_modifier_pair));
+
+	fm_pair->fourcc = format;
+	fm_pair->modifier = modifier;
+
+	logprint(TRACE, "wlroots: format %u (%lu)", fm_pair->fourcc, fm_pair->modifier);
+
+	wl_list_insert(&ctx->format_modifier_pairs, &fm_pair->link);
+}
+
+static void wlr_format_modifier_pair_emtpy_list(struct xdpw_screencast_context *ctx) {
+	struct xdpw_format_modifier_pair *fm_pair, *tmp;
+	wl_list_for_each_safe(fm_pair, tmp, &ctx->format_modifier_pairs, link) {
+		wl_list_remove(&fm_pair->link);
+		free(fm_pair);
+	}
+}
+
 static void linux_dmabuf_handle_modifier(void *data,
 		struct zwp_linux_dmabuf_v1 *zwp_linux_dmabuf_v1,
 		uint32_t format, uint32_t modifier_hi, uint32_t modifier_lo) {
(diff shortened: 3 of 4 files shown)


columbarius:
@emersion: Could you please take a look at the `dev_t, drmDevicePtr` handling in the [second commit in screencast_common.c](https://github.com/emersion/xdg-desktop-portal-wlr/pull/207/commits/bfaec1611de084c62a7179729f718b2e14c44d54#diff-7004fde2b3e368e364c57aa5dc98b5ce8e117c8a90022233b6422b978a31b3baR26). It works, but I couldn't find good documentation on those functions/structs.

columbarius:
Right, linux_dmabuf_v1 version 4 deprecates the modifier event. Added a parser for the format_table.

src/screencast/screencast_common.c
```diff
@@ -23,13 +23,18 @@ void randname(char *buf) {
...
-static char *gbm_find_render_node() {
+static char *gbm_find_render_node(drmDevicePtr device) {
 	drmDevice *devices[64];
 	char *render_node = NULL;
 
 	int n = drmGetDevices2(0, devices, sizeof(devices) / sizeof(devices[0]));
 	for (int i = 0; i < n; ++i) {
 		drmDevice *dev = devices[i];
+		if (device) {
+			if(!drmDevicesEqual(device, dev)) {
```
emersion, line 34: Nit: can be folded into one condition:

```c
if (device && !drmDevicesEqual(device, dev)) {
```

src/screencast/screencast_common.c
```diff
@@ -41,11 +46,18 @@ static char *gbm_find_render_node() {
...
 }
 
-struct gbm_device *xdpw_gbm_device_create(void) {
+struct gbm_device *xdpw_gbm_device_create(struct xdpw_screencast_context *ctx) {
 	struct gbm_device *gbm;
+	drmDevicePtr main_dev;
 	char *render_node = NULL;
 
-	render_node = gbm_find_render_node();
+	drmGetDeviceFromDevId(ctx->main_device, /* flags */ 0, &main_dev);
```
emersion, line 54: If this fails (because the device disappeared, or the `dev_t` value got mangled, etc), `main_dev` is undefined memory and will cause a crash in `gbm_find_render_node`.

src/screencast/screencast_common.c
```diff
@@ -23,13 +23,18 @@ void randname(char *buf) {
 	}
 }
 
-static char *gbm_find_render_node() {
+static char *gbm_find_render_node(drmDevicePtr device) {
```
emersion, line 26: Nit: in general I advise against using libdrm's `Ptr` types as they are confusing for no good reason. `drmDevice *device` is less confusing.

src/screencast/wlr_screencast.c
```diff
@@ -545,28 +545,90 @@ static void wlr_remove_output(struct xdpw_wlr_output *out) {
...
 
-	struct xdpw_format_modifier_pair *fm_pair = calloc(1, sizeof(struct xdpw_format_modifier_pair));
-
-	fm_pair->fourcc = format;
-	fm_pair->modifier = ((((uint64_t)modifier_hi) << 32) | modifier_lo);
-
-	logprint(TRACE, "wlroots: format %u (%u)", fm_pair->fourcc, fm_pair->modifier);
-
-	wl_list_insert(&ctx->format_modifier_pairs, &fm_pair->link);
+	wlr_format_modifier_pair_add(ctx, format, ((((uint64_t)modifier_hi) << 32) | modifier_lo));
```
emersion, line 576: Nit: there is an unnecessary parenthesis here. Maybe just use an intermediate `uint64_t modifier` variable to avoid the parenthesis soup?

include/wlr_screencast.h
```diff
@@ -12,7 +12,8 @@
 
 #define XDG_OUTPUT_MANAGER_VERSION 3
 
-#define LINUX_DMABUF_VERSION 3
+#define LINUX_DMABUF_VERSION 4
```
emersion, line 15: Nit: maybe rename to `_MIN`?

src/screencast/wlr_screencast.c
```diff
@@ -545,28 +545,90 @@ static void wlr_remove_output(struct xdpw_wlr_output *out) {
...
+		uint32_t format;
+		uint8_t padding;
+		uint64_t modifier;
+	};
+	uint32_t n_modifiers = size / 16;
+
+	void *fm_data = mmap(NULL , size, PROT_READ, MAP_PRIVATE, fd, 0);
+
+	struct fm_entry *fm_entry = fm_data;
+	for (uint32_t i = 0; i < n_modifiers; i++) {
```
emersion, line 611: This won't work -- the table might contain unsupported modifiers (e.g. some intended for a different device, or for a different client, or just arbitrary holes/undefined entries). We need to listen to `tranche_formats` to fill the list.

src/screencast/wlr_screencast.c
```diff
@@ -545,28 +545,90 @@ static void wlr_remove_output(struct xdpw_wlr_output *out) {
...
+	wlr_format_modifier_pair_emtpy_list(ctx);
+
+	struct fm_entry {
+		uint32_t format;
+		uint8_t padding;
+		uint64_t modifier;
+	};
+	uint32_t n_modifiers = size / 16;
+
+	void *fm_data = mmap(NULL , size, PROT_READ, MAP_PRIVATE, fd, 0);
```
emersion, line 608: This might fail, in case `MAP_FAILED` is returned.

src/screencast/wlr_screencast.c
```diff
@@ -545,28 +545,90 @@ static void wlr_remove_output(struct xdpw_wlr_output *out) {
...
+		struct zwp_linux_dmabuf_feedback_v1 *zwp_linux_dmabuf_feedback_v1, int fd, uint32_t size) {
+	struct xdpw_screencast_context *ctx = data;
+
+	logprint(DEBUG, "wlroots: linux_dmabuf_feedback_format_table called");
+
+	wlr_format_modifier_pair_emtpy_list(ctx);
+
+	struct fm_entry {
+		uint32_t format;
+		uint8_t padding;
```
emersion, line 603: This needs to be `uint32_t`. It might work "by chance" thanks to struct packing, but it's not something we should rely on.

src/screencast/wlr_screencast.c
```diff
@@ -545,28 +545,90 @@ static void wlr_remove_output(struct xdpw_wlr_output *out) {
...
+		uint32_t format, uint64_t modifier) {
+	struct xdpw_format_modifier_pair *fm_pair = calloc(1, sizeof(struct xdpw_format_modifier_pair));
+
+	fm_pair->fourcc = format;
+	fm_pair->modifier = modifier;
+
+	logprint(TRACE, "wlroots: format %u (%lu)", fm_pair->fourcc, fm_pair->modifier);
+
+	wl_list_insert(&ctx->format_modifier_pairs, &fm_pair->link);
+
```
emersion, line 558: Style nit: stray newline

src/screencast/wlr_screencast.c
```diff
@@ -545,28 +545,90 @@ static void wlr_remove_output(struct xdpw_wlr_output *out) {
...
+static void wlr_format_modifier_pair_add(struct xdpw_screencast_context *ctx,
+		uint32_t format, uint64_t modifier) {
+	struct xdpw_format_modifier_pair *fm_pair = calloc(1, sizeof(struct xdpw_format_modifier_pair));
+
+	fm_pair->fourcc = format;
+	fm_pair->modifier = modifier;
+
+	logprint(TRACE, "wlroots: format %u (%lu)", fm_pair->fourcc, fm_pair->modifier);
+
+	wl_list_insert(&ctx->format_modifier_pairs, &fm_pair->link);
```
emersion, line 557: It might be easier to use a `wl_array` for the format+modifier list, instead of using a `wl_list`.

src/screencast/wlr_screencast.c
```diff
@@ -545,28 +545,90 @@ static void wlr_remove_output(struct xdpw_wlr_output *out) {
...
+	logprint(DEBUG, "wlroots: linux_dmabuf_feedback_format_table called");
+
+	wlr_format_modifier_pair_emtpy_list(ctx);
+
+	struct fm_entry {
+		uint32_t format;
+		uint8_t padding;
+		uint64_t modifier;
+	};
+	uint32_t n_modifiers = size / 16;
```
emersion, line 606: Better to use `sizeof(struct fm_entry)` to avoid hardcoding this magic number.

src/screencast/screencast_common.c
columbarius: I didn't find drmGetDiveFromDevId documented in the local installed header file, but the one in [your MR to mesa](https://gitlab.freedesktop.org/mesa/drm/-/merge_requests/129) suggest, that we can check for a non negative return value for success.

src/screencast/wlr_screencast.c
columbarius: `wl_list` currently works. I would put this onto the to refactor list

src/screencast/wlr_screencast.c
columbarius: Right ... confused myself. Thx.

src/screencast/wlr_screencast.c
columbarius: done

include/wlr_screencast.h
columbarius: How should wayland protocol versioning work? My guess would be to aim at the latest available version, which has usefull features for us and then add a fallback if possible for lower versions. With this semantics we have
- (Aimed) Version
- Minimum version

src/screencast/wlr_screencast.c
columbarius: done

src/screencast/screencast_common.c
columbarius: done

src/screencast/screencast_common.c
columbarius: I'd rather have the additional check to log if we are using the main_device or not.

src/screencast/wlr_screencast.c
columbarius: Changed this to an assert. In this case I'm for a documented magic number, since this is specified by the protocol.

src/screencast/wlr_screencast.c
```diff
@@ -545,28 +546,146 @@ static void wlr_remove_output(struct xdpw_wlr_output *out) {
...
+	assert(device_arr->size == sizeof(device));
+	memcpy(&device, device_arr->data, sizeof(device));
+
+	drmDevice *drmDev;
+	if (drmGetDeviceFromDevId(device, /* flags */ 0, &drmDev) < 0) {
+		return;
+	}
+	drmDevice *drmDevRenderer = NULL;
+	drmGetDevice2(gbm_device_get_fd(ctx->gbm), /* flags */ 0, &drmDevRenderer);
+	ctx->feedback_data.used_device = drmDevicesEqual(drmDevRenderer, drmDev) > 0;
```
columbarius, line 676: @emersion  Is this the right way to compare them?

src/screencast/wlr_screencast.c
```diff
@@ -545,28 +546,146 @@ static void wlr_remove_output(struct xdpw_wlr_output *out) {
...
+
+	logprint(DEBUG, "wlroots: linux_dmabuf_feedback_handle_done called");
+	munmap(ctx->feedback_data.format_table_data, ctx->feedback_data.format_table_size);
+	ctx->feedback_data.format_table_data = NULL;
+	ctx->feedback_data.format_table_size = 0;
+}
+
+static void linux_dmabuf_feedback_tranche_done(void *data,
+		struct zwp_linux_dmabuf_feedback_v1 *zwp_linux_dmabuf_feedback_v1) {
+	logprint(DEBUG, "wlroots: linux_dmabuf_feedback_tranche_done called");
```
columbarius, line 625: Note for me: Set used_device to `false`

columbarius: done

src/screencast/screencast_common.c
columbarius: done

src/screencast/screencast_common.c
columbarius: I reworked this to try the main_device, then try a fallback on each tranche_target_device and if nothing succeeds here, we use the first render device (which we also need for the pre feedback case).

src/screencast/wlr_screencast.c
columbarius: Thanks. Implemented to go through the tranches. Just needed sth. yesterday to get dmabufs available again.

src/screencast/wlr_screencast.c
columbarius: done

emersion:
I won't have time to go through this PR again.

columbarius:
Ok thanks for your review. Will finish that up myself.

src/screencast/wlr_screencast.c
columbarius: The return of [drmDevicesEqual](https://gitlab.freedesktop.org/mesa/drm/-/blob/main/xf86drm.c#L3719) are compatible with a bool, so we can just assign the return value directly.

src/screencast/wlr_screencast.c
columbarius: Addressed in #209

PR #208, opened 2022-05-05
Nuke wp protocols

columbarius:


protocols/meson.build
```diff
@@ -1,22 +1,25 @@
...
 	)
 else
 	wayland_scanner = find_program('wayland-scanner', native: true)
 endif
 
 client_protocols = [
-	'linux-dmabuf-unstable-v1.xml',
-	'wlr-screencopy-unstable-v1.xml',
-	'xdg-output-unstable-v1.xml',
+	[wl_protocol_dir, 'unstable/linux-dmabuf/linux-dmabuf-unstable-v1.xml'],
```
emersion, line 14: Nit: can be simplified with `/`: `wl_protocol_dir / 'unstable/linux-dmabuf/…'`

emersion:
LGTM

protocols/meson.build
columbarius: done

Issue #214, opened 2022-06-04
Crash when screen casting


zsolt-donca:
I have trouble setting up xdpw on my new Arch Linux system using Sway, with Intel Alder Lake CPU and integrated GPU.

xdpw crashes when trying to do screen casting; I've included debug logs, coredump, and the sway logs:  https://gist.github.com/zsolt-donca/7eda442053e0cfd3a848088fb0bdc361

Something that stood out for me in the logs was:
```
2022/06/04 07:28:56 [INFO] - wlroots: no modifiers available for format 0
2022/06/04 07:28:56 [ERROR] - xdg-desktop-portal-wlr: failed to convert drm format 0x00000000 to spa_video_format
```

Reproduced using a local xdpw build, and using xdp-screen-cast.py. 

I am on an up-to-date Arch Linux, using the latest git versions of wlroots, sway.

The environment variables XDG_CURRENT_DESKTOP, WAYLAND_DISPLAY should be properly set.

columbarius:
Please provide a log of xdpw with logleven TRACE

zsolt-donca:
Here are the trace logs: https://gist.github.com/zsolt-donca/74c6e441ece55b7418b25c78829da98b

zsolt-donca:
Did some further testing. Turns out that xdpw works just fine with the latest released versions of wlroots+sway, it the latest master builds that it crashes with. Any idea how to investigate this? Should I raise an issue with sway as well?

columbarius:
You could create a log with the env var `WAYLAND_DEBUG=1`. This should print every wayland interaction.
If you could bisect sway/wlroots, that would be great to find the faulty commit.

The issue is that the screencopy request just fails and as such we don't get any valid format.
```
2022/06/04 08:18:49 [TRACE] - wlroots: callbacks registered
2022/06/04 08:18:49 [TRACE] - wlroots: failed event handler
```

@emersion wasn't there a change recently how outputs are referenced? Could that be not applied to screencopy?

columbarius:
> it the latest master builds that it crashes with

I build it with the current git head. Did you update the wlroots subproject? Got one sway commit since the last time I build sway, but a lot of wlroot commits.

zsolt-donca:
> Did you update the wlroots subproject?

[...]

PR #215, opened 2022-06-06
screencast: check if a valid format was received from wlr_screencopy

columbarius:
When a screencast is started we do a roundtrip on get the offered
formats from wlr_screencopy. This roundtrip can fail [1]. In this case
the formats won't be initialized and we will just error out of the
screencast session.

[1] https://github.com/emersion/xdg-desktop-portal-wlr/issues/214

### DIFF
--- a/src/screencast/screencast.c
+++ b/src/screencast/screencast.c
@@ -6,6 +6,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <assert.h>
+#include <drm_fourcc.h>
 #include <sys/wait.h>
 #include <sys/mman.h>
 #include <spa/utils/result.h>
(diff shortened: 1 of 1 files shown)


src/screencast/screencast.c
```diff
@@ -167,6 +168,13 @@ static int start_screencast(struct xdpw_screencast_instance *cast) {
 	wl_display_dispatch(cast->ctx->state->wl_display);
 	wl_display_roundtrip(cast->ctx->state->wl_display);
 
+	if (cast->screencopy_frame_info[WL_SHM].format == DRM_FORMAT_INVALID ||
```
emersion, line 171: The SHM format should not be a DRM format? See `enum wl_shm_format`. 0 there means `WL_SHM_FORMAT_XRGB8888`, and `DRM_FORMAT_INVALID` also happens to be 0.

emersion:
Maybe it would be easier to set a bool field when receiving the `buffer_done` event, and check that field after the roundtrip?

src/screencast/screencast.c
columbarius: We are using drm_format for everything internally, so this shouldn't be an issue. (see https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/src/screencast/wlr_screencast.c#L106)

emersion: Oh okay that makes sense then

columbarius:
I'd rather not use a flag for each state of setup we reached. Since we use drm_format internally for shm, the 0 shouldn't be an issue. I thought about making it possible for compositors to drop shm buffer screencopy in the future (signaled with an invalid shm_format, if there is one) and this could work with checking the format, but I saw no reason to make a more detailed check here. I assume that failing screencopy, because the compositor can't find a usable format (while we are fine with any at that point) should not be common on released swas/wlroot versions.

PR #219, opened 2022-06-19
[WIP] screencast: Use PipeWire to drive screencast

columbarius:
Without PW_STREAM_FLAG_DRIVER we can use on_process callbacks to drive
the stream.


emersion:
Merged as part of #305

PR #228, opened 2022-08-05
Draft: Json chooser v2: A complex chooser to handle options

columbarius:
If we want to query the user for options like `persist_mode` we need to send more data than just the name of the output. We also need sth. more complex when window selection is possible. 

In any case this is a PoC sending the data encoded via json between chooser and xdpw in both directions. The scheme is versioned with `version (breaking)` and `revision (backwards compatible on the xdpw side)`. 

Note: man-page has still the scheme from #105 . The new scheme is demonstrated in `chooser-protocol.json`.

Windows got an additional arbitrary id, since two toplevels can have the same `app_id` and `title`.

Supersedes: #105 

CC: @emersion

### DIFF
--- a/.builds/alpine.yml
+++ b/.builds/alpine.yml
@@ -10,6 +10,7 @@ packages:
   - scdoc
   - libdrm
   - mesa-dev
+  - json-c-dev
 sources:
   - https://github.com/emersion/xdg-desktop-portal-wlr
 tasks:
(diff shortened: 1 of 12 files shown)


columbarius:
Updated this with an idea to separate the chooser by using choser, chooser_opts and a target struct. 
There are still open todos like collecting the persist_mode.

YellowOnion:
being able to choose the fps would be nice as well, max_fps could be a default.

PR #234, opened 2022-09-02
Restore session transient

columbarius:
This is a first step to restore the session.

While we don't have a choose capable of communicating the option to allow a program to restore the session, we can at least allow transient sessions to be restored via a config option.

This would let users select the output once when using a chromium based browser.


columbarius:
Superseeded by #179

PR #238, opened 2022-09-17
contrib/systemd: remove unnecessary `After=` directive

ngkz:
`After=graphical-session.target` causes a dependency loop.

If there are user services that depend on the portal and start before `graphical-session.target`, `graphical-session.target` waits for the service, the service waits for `xdg-desktop-portal`, `xdg-desktop-portal` waits for `xdg-desktop-portal-wlr`, and then `xdg-desktop-portal-wlr` waits for `graphical-session.target`.

The service and the portal stuck until `xdg-desktop-portal-wlr` activation timeout (25s).

The compositor must be initialized before GUI apps therefore this directive is not needed.

### DIFF
--- a/contrib/systemd/xdg-desktop-portal-wlr.service.in
+++ b/contrib/systemd/xdg-desktop-portal-wlr.service.in
@@ -1,7 +1,6 @@
 [Unit]
 Description=Portal service (wlroots implementation)
 PartOf=graphical-session.target
-After=graphical-session.target
 ConditionEnvironment=WAYLAND_DISPLAY
 
 [Service]


ngkz:
I mean, `graphical-session.target` gets started after the compositor therefore  `xdg-desktop-portal-wlr` does not need to wait for anything

emersion:
Maybe we should replace that with `graphical-session-pre.target`?

We use `ConditionEnvironment`, which is known to not work correctly. I think with that patch systemd would only check the `ConditionEnvironment` once, never trying to start xdpw again once the Wayland compositor is up. More details here: https://github.com/emersion/mako/commit/6c955b74a74a3eb7eed6566ddcec125bc3a55c8c

BTW, mako has this `After` key and seems to work correctly, so not sure what's up here.

emersion:
Closing, please re-open if you believe this should be discussed further.

PR #258, opened 2023-02-19
screencast: calculate spa buffer sizes

columbarius:
Modern GPUs support many modifiers which increases the size required for the param_buffer. Instead of randomly bumping the hardcoded value we should calculate the required space dynamically.

### DIFF
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -74,6 +74,53 @@ static struct spa_pod *fixate_format(struct spa_pod_builder *b, enum spa_video_f
 	return spa_pod_builder_pop(b, &f[0]);
 }
 
+static uint32_t calc_param_size(struct xdpw_screencast_instance *cast, bool fixate, uint32_t min_size) {
+	// Read from spa/pod/pod.h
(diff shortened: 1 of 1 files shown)


emersion:
The [docs](https://docs.pipewire.org/page_spa_pod.html) say:

> It is also possible to dynamically grow the memory as needed.

Can we use that instead?

columbarius:
Then we would have to use a separate buffer for each param since a realloc would invalidate previous pointers to that memory or we could try to create the params and when that fails increase the buffersize and create them again.

emersion:
No, I mean that this doc sentence hints that PipeWire has an API to dynamically allocate the buffers.

columbarius:
Ahh I see there is a new builder api (`spa_pod_dynamic_builder`) in PipeWire. Will take a look at it.

columbarius:
Implemented #260 just to realize, while trying to figure out the minimum PipeWire version, that after the second reallocation we have dangling pointers. See: https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/d3a1b9fe31ba5978fa50299921fd4481f125dba2

columbarius:
Yes, updated #260

columbarius:
Superseeded by #260

PR #260, opened 2023-02-20
screencast: Use spa_pod_dynamic_builder

columbarius:
Modern GPUs support many modifiers which increases the size required for the param_buffer. Instead of randomly bumping the hardcoded value we can use spa_pod_dynamic_builder, which increases the buffer automatically.

Superseedes: #258


emersion:
Looks good to me from a quick look.

columbarius:
Thanks!

PR #264, opened 2023-03-21
Screencast: fix buffer type only being set on first plane

Ferdi265:
xdg-desktop-portal-wlr previously only set the buffer type of the first plane of the buffer. This, among possibly other issues, causes the file descriptor of DMA-BUF buffers to get lost somewhere between the pipewire daemon and the consuming application (e.g. obs).

The client libpipewire will see a buffer with a type of -1, and will proceed to drop the file descriptor, leaving the user with an invalid DMA-BUF if the format has more than one plane. This fixes that issue.

I initially encountered this issue trying to add xdg-desktop-portal support to [wl-mirror](https://github.com/Ferdi265/wl-mirror), but the same issue also happens with [obs-studio](https://github.com/obsproject/obs-studio).

With obs, this manifests as obs first negotiating a tiled format with 2 planes, failing to import the broken DMA-BUF, and then renegotiating to a different format with only one plane. After this patch, obs successfully imports the initial DMA-BUF with 2 planes.

This can be tested easily by simply adding a debug print into [obs-studio/plugins/linux-pipewire/pipewire.c](https://github.com/obsproject/obs-studio/blob/master/plugins/linux-pipewire/pipewire.c#L614) printing out the fd and type of the buffer.

### DIFF
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -339,6 +339,7 @@ fixate_format:
 static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
 	struct xdpw_screencast_instance *cast = data;
 	struct spa_data *d;
+	enum spa_data_type t;
 
 	logprint(DEBUG, "pipewire: add buffer event handle");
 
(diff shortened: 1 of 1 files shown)


emersion (approved):
Good catch! LGTM.

columbarius:
Thanks! LGTM

Issue #268, opened 2023-05-09
UseIn= superseded by per-DE portal configurations


kennylevinsen:
As of https://github.com/flatpak/xdg-desktop-portal/commit/939f0b0fcec6a42c1acdc397986547e9805359ff, UseIn= has been superseded by a portal configuration file.

This file is means to exist per "Desktop environment" (i.e., the `XDG_CURRENT_DESKTOP` value), and can specify providers per portal, as well as a default. This allows us to circumvent the major nuisance of the GNOME portal being used as fallback.

Do we want to carry an example here, have compositors carry their own, or let package managers define one suitable for their distro and dependencies?

columbarius:
I've tested the new configuration file and it still takes arround 20-30 seconds for gtk applications to start, but after that they are no longer unresponsive. 

I can upload my configuration file as an example, but I'd suggest compositors shipping their own configuration file, which distro maintainers can adapt to their choice of components.

I'll ask xdg-desktop-portal for a new release.

PR #269, opened 2023-05-09
contrib: Add portals config example

columbarius:
This is an example configuration for choosing the portal implementations which should be used [1]. Compositors and distributions are expected to ship their modified version according to their choice of components.

[1] https://github.com/flatpak/xdg-desktop-portal/pull/955

Resolves: #268

### DIFF
--- a/contrib/wlroots-portals.conf
+++ b/contrib/wlroots-portals.conf
@@ -1,0 +1,7 @@
+[preferred]
+# Use xdg-desktop-portal-gtk for every portal interface...
+default=gtk
+# ... except for the Screencast, Screenshot and Settings (dark/light mode) interface
+org.freedesktop.impl.portal.Screencast=wlr
+org.freedesktop.impl.portal.Screenshot=wlr
+org.freedesktop.impl.portal.Settings=darkman


emersion:
LGTM

contrib/wlroots-portals.conf
```diff
@@ -0,0 +1,7 @@
+[preferred]
+# Use xdg-desktop-portal-gtk for every portal interface...
+default=gtk
+# ... except for the Screencast, Screenshot and Settings (dark/light mode) interface
+org.freedesktop.impl.portal.Screencast=wlr
```
alebastr, line 5: `ScreenCast`.

The case mismatch causes x-d-p to fallback to UseIn, which inconveniently hides the error. But at least there's a log:
```
xdg-desktop-por[24921]: Choosing wlr.portal for org.freedesktop.impl.portal.ScreenCast via the deprecated UseIn key
xdg-desktop-por[24921]: The preferred method to match portal implementations to desktop environments is to use the portals.conf(5) configuration file
```

emersion: Seems like this might also be an issue upstream?

https://github.com/flatpak/xdg-desktop-portal/pull/955/files#diff-0958e51ba7fb0f74e91c3e294229b576a1e7d022c734f45aaa10c5a56b72c268R72

Consolatis:
I have no experience with portals but I am a bit puzzled to see that the example config
- adds some -wlr external interface (`org.freedesktop.impl.portal.Settings=darkman`)
- hardcodes the default to `gtk`

Would this work as well, without making decisions for the user what kind of portals to use for interfaces that the -wlr implementation does not support? Or should it just be `default=wlr` without the `;*` at the end?

```ini
[preferred]
default=wlr;*
```

emersion:
The intent seems to provide an _example_ configuration though?

Consolatis:
Alright thanks for the replies.

PR #276, opened 2023-07-25
Handle pipewire restarts

### LINKED ISSUE #178: Needs restart after pipewire restart
On Arch Linux, xdg-desktop-portal-wlr 0.5.0.

I was playing around with pipewire and noticed that after restarting that OBS would not longer get a picture.
It looks to me as if after issuing a `systemctl --user restart pipewire.socket pipewire-pulse.socket` I also need to issue a `systemctl --user restart xdg-desktop-portal-wlr.service`.

Would it make sense to do this automatically via xdg-desktop-portal-wlr.service?
```
[Unit]
...
BindsTo=pipewire.service
...
```

### LINKED ISSUE #242: Screensharing disabled after restarting pipewire service
I ran into this issue after having to restart pipewire service.
To reproduce:

share screen (ok)
stop sharing (ok)
restart pipewire (systemctl --user restart pipewire pipewire-pulse wireplumber)
share screen (unable to share screen, only black screen visible to participants)

If I restart pipewire without sharing screen before the restart, then I'm able to share screen even though pipewire has been restarted. Log attached for the procedure outlined above.

[sway.log](https://github.com/emersion/xdg-desktop-portal-wlr/files/9756718/sway.log)

cillian64:
If the pipewire daemon restarted while xdpw was running then our pipewire connection would drop and attempting to start a screencast wouldn't work until user manually restarted xdpw.

This commit adds a listener for fatal core connection errors from pipewire.  In the handler it cleans up the connection and marks it as dead.  Then, when the user attempts to start a screencast, we reconnect the pipewire core if needed.

To reproduce the problem:
- Start off with pipewire, xdg-desktop-portal, and xdg-desktop-portal-wlr running as normal in a working state
- `systemctl --user restart pipewire`
- Go to https://mozilla.github.io/webrtc-landing/gum_test.html in a recent release of Firefox and click "Screen capture"
- The screen capture does not start and xdg-desktop-portal-wlr hangs, consuming a full CPU core in a hard loop

With this change applied, the screen capture works after restarting pipewire (after testing the bad case, make sure to restart Firefox before testing with the fix applied).

Fixes #242, fixes #178

### DIFF
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -594,6 +594,21 @@ void xdpw_pwr_stream_destroy(struct xdpw_screencast_instance *cast) {
 	cast->stream = NULL;
 }
 
+static void on_core_error(void *data, uint32_t id, int seq, int res, const char* message) {
+	// If our pipewire connection drops then we won't be able to actually
+	// do a screencast.  Exit the process so someone restarts us and the
+	// new xdpw can reconnect to pipewire.
+	logprint(ERROR, "pipewire: fatal error event from core");
+	exit(1);
+}
+
+static const struct pw_core_events core_events = {
+	PW_VERSION_CORE_EVENTS,
+	.error = on_core_error,
+};
+
+static struct spa_hook core_listener;
+
 int xdpw_pwr_context_create(struct xdpw_state *state) {
 	struct xdpw_screencast_context *ctx = &state->screencast;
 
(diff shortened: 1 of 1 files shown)


emersion:
Yeah, I'd prefer to `exit(1)` when we get disconnected from PipeWire or Wayland. Trying to recreate our state doesn't seem worth the trouble.

emersion (approved):
LGTM, thanks!

PR #282, opened 2023-09-07
Fix bug in sd-bus polling

### LINKED ISSUE #281: Bug in sd-bus polling code causes hang
I've been hitting a bug where, on session start, xdg-desktop-portal (xdp) occasionally hangs in `xdp_dbus_impl_screenshot_proxy_new_sync`.  This blocking call, amongst other things, starts xdg-desktop-portal-wlr (xdpw) and calls `GetAll` on `org.freedesktop.DBus.Properties` on xdpw.  Based on dbus-monitor logs it looks like xdpw doesn't respond to the `GetAll` call for several minutes.  Eventually xdpw receives another dbus message and then responds to both immediately.

I believe this is because of a bug in how xdpw uses `poll()` with sd-bus.  [The docs for sd_bus_get_fd](https://www.freedesktop.org/software/systemd/man/sd_bus_get_fd.html) require that all three of `sd_bus_get_fd`, `sd_bus_get_events`, and `sd_bus_get_timeout` are called for every invocation of `poll`.  xdpw only calls `sd_bus_get_fd` once before its main loop and never calls `sd_bus_get_events` or `sd_bus_get_timeout`.

What's happening with my hang is that sd-bus has a message in its receive queue before xdpw polls. `sd_bus_get_events` and `sd_bus_get_timeout` would both return 0 to indicate that poll should return immediately so we can handle the message.  Instead, xdpw goes into poll with no timeout.  Because the received message has already been read from the FD to the sd-bus receive queue, poll will not see it and we don't handle the message until another arrives.

I've got a provisional fix for this but I'd appreciate some comments on how it could be refined. I'll submit a PR shortly.

System details:
- Raspberry Pi 4, arm64
- Debian 12 bookworm
- xdg-desktpo-portal 1.16.0-2
- xdg-desktop-portal-wlr 0.7.0-1+rpt1
- dbus 1.14.8-2~deb12u1
- systemd 252.12-1~deb12u1

cillian64:
sd-bus requires that `sd_bus_get_fd`, `sd_bus_get_events`, and `sd_bus_get_timeout` are called for every invocation of `poll`, and the returned FD, events, and timeout used.  In particular, sd-bus may return `events=0` and `timeout=0` if it already has dbus messages to process in its receive queue.

xdpw currently doesn't call `sd_bus_get_events` or `sd_bus_get_timeout` at all.  This can result in the situation where xdpw goes into a `poll` with no timeout and a message in sd-bus' receive queue is never handled until another dbus message arrives.  I have a case where this ignored messages causes xdg-desktop-portal to hang.

This commit changes the polling code to match sd-bus' requirements.

Fixes #281

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -175,10 +175,7 @@ int main(int argc, char *argv[]) {
 	wl_list_init(&state.timers);
 
 	struct pollfd pollfds[] = {
-		[EVENT_LOOP_DBUS] = {
-			.fd = sd_bus_get_fd(state.bus),
-			.events = POLLIN,
-		},
+		[EVENT_LOOP_DBUS] = {0}, // Filled in later
 		[EVENT_LOOP_WAYLAND] = {
 			.fd = wl_display_get_fd(state.wl_display),
 			.events = POLLIN,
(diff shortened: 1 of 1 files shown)


name-snrl:
As I said in #281, this PR completely solves the random startup delay for me. I hope it will be merged in the near feature

src/core/main.c
```diff
@@ -196,7 +193,27 @@ int main(int argc, char *argv[]) {
 	state.timer_poll_fd = pollfds[EVENT_LOOP_TIMER].fd;
 
 	while (1) {
-		ret = poll(pollfds, sizeof(pollfds) / sizeof(pollfds[0]), -1);
+		// sd-bus requires that we update FD/events/timeout every time we poll
+		pollfds[EVENT_LOOP_DBUS].fd = sd_bus_get_fd(state.bus);
+		if (pollfds[EVENT_LOOP_DBUS].fd < 0) {
+			logprint(ERROR, "sd_bus_get_fd failed: %s", strerror(errno));
```
emersion, line 199: Nit: it doesn't seem like sd-bus populates errno. It seems like the error number is returned instead.

emersion:
LGTM apart from this nit.

emersion (approved):
Thanks!

cillian64:
Thanks for the review!

PR #283, opened 2023-09-07
Remove unused calls to xdpw_request_create

r3dey3:
The req object was not used and these calls were resulting with an error when attempting the pick-color from kdenlive on ubuntu 22.04 and 23.04

These objects were also not freed which is a memory leak.

### DIFF
--- a/src/screencast/screencast.c
+++ b/src/screencast/screencast.c
@@ -263,12 +263,6 @@ static int method_screencast_create_session(sd_bus_message *msg, void *data,
 		return ret;
 	}
 
-	struct xdpw_request *req =
-		xdpw_request_create(sd_bus_message_get_bus(msg), request_handle);
-	if (req == NULL) {
-		return -ENOMEM;
-	}
-
 	struct xdpw_session *sess =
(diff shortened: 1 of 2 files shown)


emersion:
Hm, I was under the impression that allocating D-Bus objects for these handles was important...

r3dey3:
So I went and read through a bit of the docs (https://docs.flatpak.org/en/latest/portal-api-reference.html#gdbus-org.freedesktop.portal.Request)  and other code - it looks like the code is supposed to respond with the Request object, then do the action (and if the Close method is called, cancel); then finally free up the request object. 

I think the main issue why things aren't working is multi faceted -
1. The request object isn't actually sent in a reply - "the reply includes a handle (i.e. object path) for a Request object"
2. "handle... which will stay alive for the duration of the user interaction" - the request object needs to be released at the end of the user interaction
3. The code doesn't process dbus messages while waiting for user input - so the close method wouldn't be processed anyways.

Feel free to reject this pull request :-)

emersion:
Superseded by #307

PR #291, opened 2023-11-23
Add Qtile to portal list

Sunderland93:


emersion:
I wonder if we should add more desktops to the mix now that xdg-desktop-portal has [switched to a different approach](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/contrib/wlroots-portals.conf)?

emersion:
Closing, please re-open if you feel differently.

PR #263, opened 2023-03-09
Add org.freedesktop.portal.RemoteDesktop support

andreafeletto:
Implements https://github.com/emersion/xdg-desktop-portal-wlr/issues/2

ToDo:
- [ ] devices selection
- [ ] pointer support
  - [x] relative motion
  - [ ] absolute motion
  - [ ] buttons
  - [ ] scroll
- [ ] keyboard support
- [ ] touch support

PR #292, opened 2024-01-06
Screencast response fixes

cptpcrd:
This PR fixes 2 errors in the response returned by the screencast interface:

- The source type was wrong (`1 << MONITOR` -> `MONITOR`).
- The position was hardcoded as (0, 0). If the compositor supports the xdg-output protocol, use that protocol to get the correct position and report it. Otherwise, omit the position entirely.

### DIFF
--- a/src/screencast/screencast.c
+++ b/src/screencast/screencast.c
@@ -585,12 +585,73 @@ static int method_screencast_start(sd_bus_message *msg, void *data,
 	if (ret < 0) {
 		return ret;
 	}
-	ret = sd_bus_message_append(reply, "{sv}",
-		"streams", "a(ua{sv})", 1,
-		cast->node_id, 3,
-		"position", "(ii)", 0, 0,
-		"size", "(ii)", cast->screencopy_frame_info[WL_SHM].width, cast->screencopy_frame_info[WL_SHM].height,
-		"source_type", "u", MONITOR);
+	ret = sd_bus_message_open_container(reply, 'e', "sv");
+	if (ret < 0) {
+		return ret;
+	}
+	ret = sd_bus_message_append(reply, "s", "streams");
+	if (ret < 0) {
+		return ret;
+	}
+	ret = sd_bus_message_open_container(reply, 'v', "a(ua{sv})");
+	if (ret < 0) {
+		return ret;
+	}
+	ret = sd_bus_message_open_container(reply, 'a', "(ua{sv})");
+	if (ret < 0) {
+		return ret;
+	}
+	ret = sd_bus_message_open_container(reply, 'r', "ua{sv}");
+	if (ret < 0) {
+		return ret;
+	}
+	ret = sd_bus_message_append(reply, "u", cast->node_id);
+	if (ret < 0) {
+		return ret;
+	}
+	ret = sd_bus_message_open_container(reply, 'a', "{sv}");
+	if (ret < 0) {
+		return ret;
+	}
+	if (cast->target->output->xdg_output) {
+		ret = sd_bus_message_append(reply, "{sv}",
+			"position", "(ii)", cast->target->output->x, cast->target->output->y);
+		if (ret < 0) {
+			return ret;
+		}
+		ret = sd_bus_message_append(reply, "{sv}",
+			"size", "(ii)", cast->target->output->width, cast->target->output->height);
+		if (ret < 0) {
+			return ret;
+		}
+	} else {
+		ret = sd_bus_message_append(reply, "{sv}",
+			"size", "(ii)", cast->screencopy_frame_info[WL_SHM].width, cast->screencopy_frame_info[WL_SHM].height);
+		if (ret < 0) {
+			return ret;
+		}
+	}
+	ret = sd_bus_message_append(reply, "{sv}", "source_type", "u", MONITOR);
+	if (ret < 0) {
+		return ret;
+	}
+	ret = sd_bus_message_close_container(reply);
+	if (ret < 0) {
+		return ret;
+	}
+	ret = sd_bus_message_close_container(reply);
+	if (ret < 0) {
+		return ret;
+	}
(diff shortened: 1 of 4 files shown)


emersion:
> The position was hardcoded as (0, 0). If the compositor supports the xdg-output protocol, use that protocol to get the correct position and report it. Otherwise, omit the position entirely.

I'm not sure that's correct. The position is in logical coordinates, while the size is in buffer coordinates.

emersion:
The first commit LGTM, I've pushed it to master.

cptpcrd:
> I'm not sure that's correct. The position is in logical coordinates, while the size is in buffer coordinates.

Hmm, I hadn't considered that. This is interesting though:

https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.impl.portal.ScreenCast.html#org-freedesktop-impl-portal-screencast-start

> - `position` (`(ii)`)
> 
>   A tuple consisting of the position (x, y) in the compositor coordinate space. Note that the position may not be equivalent to a position in a pixel coordinate space. Only available for monitor streams.
> 
> - `size` (`(ii)`)
> 
>   A tuple consisting of (width, height). The size represents the size of the stream as it is displayed in the compositor coordinate space. Note that this size may not be equivalent to a size in a pixel coordinate space. The size may differ from the size of the stream.

https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/unstable/xdg-output/xdg-output-unstable-v1.xml

> 	The position event describes the location of the wl_output within
> 	the global compositor space.
> 
> 	The logical_position event is sent after creating an xdg_output
> 	(see xdg_output_manager.get_xdg_output) and whenever the location
> 	of the output changes within the global compositor space.
> 
> 	...
> 
> 	The logical_size event describes the size of the output in the
> 	global compositor space.
> 
> 	...

Unless I'm misreading something here (which is possible, I've only done a llittle Wayland programming), it seems that the logical coordinates from xdg-output should really be used for both the position and size?

emersion:
Yeah, sounds like it.

Note that ideally we shouldn't make xdg-output mandatory (fallback to the current behavior in that case).

cptpcrd:
Updated to use the logical position and size from xdg-output if available, otherwise report the buffer size and omit the position. ([This page](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.ScreenCast.html#org-freedesktop-portal-screencast-start) seems to indicate the position is optional, and IMO it's better to leave it out than have 
 it be wrong. If you feel differently I can change it.)

src/screencast/screencast.c
```diff
@@ -585,12 +585,73 @@ static int method_screencast_start(sd_bus_message *msg, void *data,
...
+			"position", "(ii)", cast->target->output->x, cast->target->output->y);
+		if (ret < 0) {
+			return ret;
+		}
+		ret = sd_bus_message_append(reply, "{sv}",
+			"size", "(ii)", cast->target->output->width, cast->target->output->height);
+		if (ret < 0) {
+			return ret;
+		}
+	} else {
```
emersion, line 627: We should probably still send "position" when xdg-output is unavailable? The D-Bus API docs suggest this is always set for MONITOR?

src/screencast/screencast.c
```diff
@@ -585,12 +585,73 @@ static int method_screencast_start(sd_bus_message *msg, void *data,
 	if (ret < 0) {
 		return ret;
 	}
-	ret = sd_bus_message_append(reply, "{sv}",
```
emersion, line 588: Did you have any specific reason why you replaced this with multiple calls? Was it for conditional properties, or did you like the explicit version better, or something else?

src/screencast/screencast.c
emersion: Oh, you replied above. And indeed, it's marked optional here. I was looking at the backend API docs where the "optional" word is missing.

Yeah, I prefer to leave it out if we don't have the info, rather than send zero.

Can we also skip sending the size if we don't have xdg-output?

cptpcrd: > Can we also skip sending the size if we don't have xdg-output?

Sure, if you think that's best.

For the record, `xdg-desktop-portal` does try to pull out [the size](https://github.com/flatpak/xdg-desktop-portal/blob/main/src/screen-cast.c#L721), which then appears to [be used](https://github.com/flatpak/xdg-desktop-portal/blob/main/src/screen-cast.c#L693C18-L693C18) by the [remote desktop](https://github.com/flatpak/xdg-desktop-portal/blob/main/src/remote-desktop.c#L821) code to determine if pointer/touch events are in bounds. So if #263 or similar is merged it would have ramifications on compositors that don't support xdg-output.

src/screencast/screencast.c
cptpcrd: Just for conditional properties. If you know a better way to do this I'm happy to change it.

src/screencast/screencast.c
emersion: If the docs say it's optional but xdp itself requires it, I'd argue that it's an xdg bug :P

emersion:
Thanks!

columbarius:
> Updated to use the logical position and size from xdg-output if available, otherwise report the buffer size and omit the position. ([This page](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.ScreenCast.html#org-freedesktop-portal-screencast-start) seems to indicate the position is optional, and IMO it's better to leave it out than have it be wrong. If you feel differently I can change it.)

No client should rely on this value for the position of the output or the buffer size. The screencast will continue if the output changes the resolution, the orientation (both changing the buffer size) or position (absolut irrelevant for xdpw) and we will update the buffer size via pipewire accordingly. This was the reason, why I omitted them.

columbarius:
It would probably be better to attach the output position as a parameter to the PipeWire stream. Would be interesting if xdp can read them.

PR #295, opened 2024-01-29
Drop wlr.portal

emersion:
xdg-desktop-portal has a new way to find portal implementations now.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -105,11 +105,6 @@ configure_file(
 	install_dir: get_option('datadir') / 'dbus-1' / 'services',
 )
 
-install_data(
-	'wlr.portal',
-	install_dir: get_option('datadir') / 'xdg-desktop-portal' / 'portals',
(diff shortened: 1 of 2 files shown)


emersion:
Hm, is the whole portal file the right thing to do? Is it now unused? Should we just drop the `UseIn` part instead?

alebastr:
`wlr.portal` is still used to resolve `wlr` in portals conf and determine supported implementations for
```ini
[preferred]
default=wlr;gtk
```

`UseIn=` will be used as a fallback when portals.conf does not define an implementation for a certain portal.
E.g. contrib/wlroots-portals.conf in this repo does not have an entry for `ScreenCast` and `default=gtk` doesn't implement it, but things work because `UseIn` fallback lookup returns `wlr`.

emersion:
Alright, let's keep this for later then.

PR #301, opened 2024-04-19
screencast: simplify wl_drm format code conversions

emersion:
All formats except ARGB8888 and XRGB8888 are guaranteed to always match the DRM FourCC.


columbarius:
Should we also update the mapping between drm and PipeWire formats for all possible values?

emersion:
I don't think there is a straightforward generic mapping like for these, like there is for `wl_shm` and DRM FourCC?

columbarius:
My idea was to manually update the conversion to formats available in PipeWire, but this can be done when needed.

columbarius:
LGTM

columbarius (approved):
LGTM

emersion:
Right, this patch is mostly intended as a cleanup to avoid unnecessary repetition and remove the need to maintain format lists for `wl_shm`.

PR #302, opened 2024-04-19
screencast: simplify xdpw_gbm_device_create()

emersion:
No need to list all DRM devices, we already have a drmDevice pointer passed in.


columbarius:
Was there a reason for the enumeration?

PR #303, opened 2024-04-19
build: don't use cc.get_supported_arguments() for -D flag

emersion:
If somehow the compiler doesn't support -D, we should fail hard instead of omitting the define. Should also speed up Meson setup time a bit.


columbarius (approved):
LGTM

PR #306, opened 2024-05-18
colorchooser: force grim to only return a single pixel

t-8ch:
On scaled outputs grim returns scaled images.
This breaks the enforced invariant in exec_color_picker() to expect single pixel images.

Force grim to use a scale of "1" to avoid this.


emersion (approved):
Thanks!

PR #309, opened 2024-05-22
screencast: fix SelectSources type bitmask check

lilyinstarlight:
In the same vein as aff78aeecd324034c6e7e976f0156c04e83f6459, `WINDOW` is the value and not the index in the bitmask. But additionally, as I understand it, a "non-monitor" source type would be one that lacks `MONITOR` rather than one that includes `WINDOW` (or `VIRTUAL`) at all.

Technically this is a change in behavior since previously `WINDOW`-only source types were accepted but treated as `MONITOR` source types, and those will now be rejected (as only `VIRTUAL` source types were technically checked by mistake due to the bitmask value/index checking bug).


emersion (approved):
Makes sense to me. Thanks!

PR #315, opened 2024-08-23
contrib/wlroots-portals: add Inhibit=none

llyyr:
Works around idle inhibit not working with Firefox.

### DIFF
--- a/contrib/wlroots-portals.conf
+++ b/contrib/wlroots-portals.conf
@@ -1,7 +1,13 @@
 [preferred]
 # Use xdg-desktop-portal-gtk for every portal interface...
 default=gtk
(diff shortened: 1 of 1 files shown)


emersion:
Sounds like bugs elsewhere:

- w-d-p-gtk shouldn't expose the idle inhibit interface if GNOME backend interfaces aren't available.
- Firefox should prefer the Wayland protocol over the portal.

llyyr:
> w-d-p-gtk shouldn't expose the idle inhibit interface if GNOME backend interfaces aren't available.

Tracked here https://github.com/flatpak/xdg-desktop-portal-gtk/issues/465

> Firefox should prefer the Wayland protocol over the portal.

Doesn't seem to be tracked anywhere

PR #326, opened 2025-03-31
ext_image_copy_capture_v1: Initial implementation

kennylevinsen:
Implement ext_image_copy_capture_v1.

Notes:
- I don't have a test for the wl_shm bits - any suggestions?
- Support for capturing 10-bit output with 8-bit clients requires https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5019

### DIFF
--- a/src/screencast/ext_image_copy.c
+++ b/src/screencast/ext_image_copy.c
@@ -1,0 +1,330 @@
+#include "wlr_screencast.h"
+
+#include "linux-dmabuf-unstable-v1-client-protocol.h"
+#include "ext-image-capture-source-v1-client-protocol.h"
+#include "ext-image-copy-capture-v1-client-protocol.h"
+#include <drm_fourcc.h>
+#include <fcntl.h>
+#include <limits.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <sys/param.h>
+#include <sys/stat.h>
+#include <sys/wait.h>
+#include <unistd.h>
+#include <assert.h>
+#include <wayland-client-protocol.h>
+#include <xf86drm.h>
+#include <sys/types.h>
+
+#include "ext_image_copy.h"
+#include "screencast.h"
+#include "pipewire_screencast.h"
+#include "xdpw.h"
+#include "logger.h"
+#include "fps_limit.h"
+
+static void ext_session_buffer_size(void *data,
+		struct ext_image_copy_capture_session_v1 *ext_image_copy_capture_session_v1,
+		uint32_t width, uint32_t height) {
+	struct xdpw_screencast_instance *cast = data;
+
+	cast->pending_constraints.width = width;
+	cast->pending_constraints.height = height;
+	cast->pending_constraints.dirty = true;
+}
+
+static void ext_session_shm_format(void *data,
+		struct ext_image_copy_capture_session_v1 *ext_image_copy_capture_session_v1,
+		uint32_t format) {
+	struct xdpw_screencast_instance *cast = data;
+
+	uint32_t fourcc = xdpw_format_drm_fourcc_from_wl_shm(format);
+
+	char *fmt_name = drmGetFormatName(fourcc);
+	struct xdpw_shm_format *fmt;
+	wl_array_for_each(fmt, &cast->pending_constraints.shm_formats) {
+		if (fmt->fourcc == fourcc) {
+			logprint(TRACE, "ext: skipping duplicated format: %s (%X)", fmt_name, fourcc);
+			goto done;
+		}
+	}
+	if (xdpw_bpp_from_drm_fourcc(fourcc) <= 0) {
+		logprint(WARN, "ext: unsupported shm format: %s (%X)", fmt_name, fourcc);
+		goto done;
+	}
+
+	fmt = wl_array_add(&cast->pending_constraints.shm_formats, sizeof(*fmt));
+	if (fmt == NULL) {
+		logprint(WARN, "ext: allocation for shm format %s (%X) failed", fmt_name, fourcc);
+		goto done;
+	}
+	fmt->fourcc = fourcc;
+	// Stride will be calculated when session_done is received
+	fmt->stride = 0;
+	cast->pending_constraints.dirty = true;
+	logprint(TRACE, "ext: shm_format: %s (%X)", fmt_name, fourcc);
+
+done:
+	free(fmt_name);
+}
+
+static void ext_session_dmabuf_device(void *data,
+		struct ext_image_copy_capture_session_v1 *ext_image_copy_capture_session_v1,
+		struct wl_array *device_arr) {
+	struct xdpw_screencast_instance *cast = data;
+
+	dev_t device;
+	assert(device_arr->size == sizeof(device));
+	memcpy(&device, device_arr->data, sizeof(device));
+
+	drmDevice *drmDev;
+	if (drmGetDeviceFromDevId(device, /* flags */ 0, &drmDev) != 0) {
+		return;
+	}
+
+	cast->pending_constraints.gbm = xdpw_gbm_device_create(drmDev);
+	cast->pending_constraints.dirty = true;
+	logprint(TRACE, "ext: dmabuf_device handler");
+}
+
+static void ext_session_dmabuf_format(void *data,
+		struct ext_image_copy_capture_session_v1 *ext_image_copy_capture_session_v1,
+		uint32_t format, struct wl_array *modifiers) {
+	struct xdpw_screencast_instance *cast = data;
+
+	char *fmt_name = drmGetFormatName(format);
+	uint64_t *modifier;
+	wl_array_for_each(modifier, modifiers) {
+		struct xdpw_format_modifier_pair *fm_pair;
+		bool new = true;
+		wl_array_for_each(fm_pair, &cast->pending_constraints.dmabuf_format_modifier_pairs) {
+			if (fm_pair->fourcc == format && fm_pair->modifier == *modifier) {
+				new = false;
+				break;
+			}
+		}
+		if (!new) {
+			logprint(TRACE, "ext: skipping duplicated format %s (%X, %lu)", fmt_name, format, *modifier);
+			continue;
+		}
+
+		fm_pair = wl_array_add(&cast->pending_constraints.dmabuf_format_modifier_pairs, sizeof(*fm_pair));
+		fm_pair->fourcc = format;
+		fm_pair->modifier = *modifier;
+
+		char *modifier_name = drmGetFormatModifierName(*modifier);
+		logprint(TRACE, "ext: dmabuf_format handler: %s (%X), modifier: %s (%X)", fmt_name, format, modifier_name, *modifier);
+		free(modifier_name);
+	}
+
+	cast->pending_constraints.dirty = true;
+	free(fmt_name);
+}
+
+static void ext_session_done(void *data,
+		struct ext_image_copy_capture_session_v1 *ext_image_copy_capture_session_v1) {
+	struct xdpw_screencast_instance *cast = data;
+
+	logprint(TRACE, "ext: done handler");
+
+	// We can only calculate the stride now we have both formats and width
+	struct xdpw_shm_format *fmt;
+	wl_array_for_each(fmt, &cast->pending_constraints.shm_formats) {
+		int bpp = xdpw_bpp_from_drm_fourcc(fmt->fourcc);
+		assert(bpp > 0);
+		fmt->stride = bpp * cast->pending_constraints.width;
+	}
+
+	if (xdpw_buffer_constraints_move(&cast->current_constraints, &cast->pending_constraints)) {
+		logprint(DEBUG, "ext: buffer constraints changed");
+		pwr_update_stream_param(cast);
+		return;
+	}
+
+	if (!cast->initialized) {
+		return;
+	}
+}
+
+static void ext_session_stopped(void *data,
+		struct ext_image_copy_capture_session_v1 *ext_image_copy_capture_session_v1) {
+	struct xdpw_screencast_instance *cast = data;
+
+	xdpw_screencast_instance_destroy(cast);
+	logprint(TRACE, "ext: session_stopped handler");
+}
+
+static const struct ext_image_copy_capture_session_v1_listener ext_session_listener = {
+	.buffer_size = ext_session_buffer_size,
+	.shm_format = ext_session_shm_format,
+	.dmabuf_device = ext_session_dmabuf_device,
+	.dmabuf_format = ext_session_dmabuf_format,
+	.done = ext_session_done,
+	.stopped = ext_session_stopped,
+};
+
+static void ext_frame_transform(void *data,
+		struct ext_image_copy_capture_frame_v1 *ext_image_copy_capture_frame_v1,
+		uint32_t transform) {
+	struct xdpw_screencast_instance *cast = data;
+	logprint(TRACE, "ext: transform handler %u", transform);
+	cast->current_frame.transformation = transform;
+}
+
+static void ext_frame_damage(void *data,
+		struct ext_image_copy_capture_frame_v1 *ext_image_copy_capture_frame_v1,
+		int32_t x, int32_t y, int32_t width, int32_t height) {
+	struct xdpw_screencast_instance *cast = data;
+
+	logprint(TRACE, "ext: damage: %"PRId32",%"PRId32"x%"PRId32",%"PRId32, x, y, width, height);
+
+	// Our damage tracking
+	struct xdpw_buffer *buffer;
+	wl_list_for_each(buffer, &cast->buffer_list, link) {
+		struct xdpw_frame_damage *damage = wl_array_add(&buffer->damage, sizeof(*damage));
+		*damage = (struct xdpw_frame_damage){ .x = x, .y = y, .width = width, .height = height };
+	}
+}
+
+static void ext_frame_presentation_time(void *data,
+		struct ext_image_copy_capture_frame_v1 *ext_image_copy_capture_frame_v1,
+		uint32_t tv_sec_hi, uint32_t tv_sec_lo, uint32_t tv_nsec) {
+	struct xdpw_screencast_instance *cast = data;
+
+	cast->current_frame.tv_sec = ((((uint64_t)tv_sec_hi) << 32) | tv_sec_lo);
+	cast->current_frame.tv_nsec = tv_nsec;
+	logprint(TRACE, "ext: timestamp %"PRIu64":%"PRIu32, cast->current_frame.tv_sec, cast->current_frame.tv_nsec);
+}
+
+static void ext_frame_ready(void *data,
+		struct ext_image_copy_capture_frame_v1 *ext_image_copy_capture_frame_v1) {
+	struct xdpw_screencast_instance *cast = data;
+	fps_limit_measure_start(&cast->fps_limit, cast->framerate);
+
+	logprint(TRACE, "ext: ready event handler");
+
+	if (cast->ext_session.frame) {
+		ext_image_copy_capture_frame_v1_destroy(cast->ext_session.frame);
+		cast->ext_session.frame = NULL;
+	}
+
+	struct xdpw_buffer *buffer = cast->current_frame.xdpw_buffer;
+	cast->current_frame.completed = true;
+	xdpw_pwr_enqueue_buffer(cast);
+	if (buffer) {
+		// Clear damage for the buffer that was just submitted
+		buffer->damage.size = 0;
+	}
+}
+
+static void ext_frame_failed(void *data,
+		struct ext_image_copy_capture_frame_v1 *ext_image_copy_capture_frame_v1,
+		uint32_t reason) {
+	struct xdpw_screencast_instance *cast = data;
+
+	if (cast->ext_session.frame) {
+		ext_image_copy_capture_frame_v1_destroy(cast->ext_session.frame);
+		cast->ext_session.frame = NULL;
+	}
+
+	switch (reason) {
+	case EXT_IMAGE_COPY_CAPTURE_FRAME_V1_FAILURE_REASON_UNKNOWN:
+		logprint(ERROR, "ext: frame capture failed: unknown reason");
+		xdpw_screencast_instance_destroy(cast);
+		return;
+	case EXT_IMAGE_COPY_CAPTURE_FRAME_V1_FAILURE_REASON_BUFFER_CONSTRAINTS:
+		logprint(ERROR, "ext: frame capture failed: buffer constraint mismatch");
+		xdpw_pwr_enqueue_buffer(cast);
+		return;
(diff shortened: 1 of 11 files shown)


llyyr:
Tested it a bit and there's two issues:

1. `[DEBUG] - pipewire: buffer already exported` is spammed hundreds of times per second while screen capture is active.
2. There's a segfault when stopping screencast in `ext_frame_damage`. `cast->current_frame.xdpw_buffer` is NULL but the logprint line tries to dereference it.

kennylevinsen:
So this turned out to require more refactoring than expected. I've kept the history somewhat progressive for now - I might try to clean it up a bit later.

> Tested it a bit and there's two issues:
> 
> 1. `[DEBUG] - pipewire: buffer already exported` is spammed hundreds of times per second while screen capture is active.
> 2. There's a segfault when stopping screencast in `ext_frame_damage`. `cast->current_frame.xdpw_buffer` is NULL but the logprint line tries to dereference it.

The segfault should be fixed. The "buffer already exported" spam also happens on master with screencopy - probably mishandling of in-flight pipewire buffers or something.

kennylevinsen:
I forgot to mention that this also fixes screen recording for 10-bit outputs when combined with this wlroots MR: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5019

llyyr:
After the latest push, ASAN complains about use-after-free when starting a capture. Though it works when not built with ASAN.

```
==2364089==ERROR: AddressSanitizer: heap-use-after-free on address 0x52100006fca8 at pc 0x5626382e58df bp 0x7ffde67a1790 sp 0x7ffde67a0f48
READ of size 280 at 0x52100006fca8 thread T0
    #0 0x5626382e58de in memcpy /home/abuild/rpmbuild/BUILD/llvm19-19.1.7-build/llvm-19.1.7.src/projects/compiler-rt/lib/asan/../sanitizer_common/sanitizer_common_interceptors_memintrinsics.inc:115:5
    #1 0x7f1a0e37b3a1  (/lib64/libpipewire-0.3.so.0+0x7b3a1) (BuildId: 532de3f5ebfec96c495741fbd1eb20a76cee1d99)
    #2 0x7f1a0e3877f6 in pw_stream_connect (/lib64/libpipewire-0.3.so.0+0x877f6) (BuildId: 532de3f5ebfec96c495741fbd1eb20a76cee1d99)
    #3 0x56263836cb2e in xdpw_pwr_stream_create /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/pipewire_screencast.c:638:2
    #4 0x56263833c210 in start_screencast /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/screencast.c:202:2
    #5 0x56263833a6f6 in method_screencast_start /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/screencast.c:555:9
    #6 0x7f1a0e4381ab  (/lib64/libsystemd.so.0+0x611ab) (BuildId: 1f93369ebde578ed090a17a2affdb2f4b8d2302c)
    #7 0x7f1a0e44617a  (/lib64/libsystemd.so.0+0x6f17a) (BuildId: 1f93369ebde578ed090a17a2affdb2f4b8d2302c)
    #8 0x562638329a17 in main /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/core/main.c:243:11
    #9 0x7f1a0de2b12d in __libc_start_call_main (/lib64/libc.so.6+0x2b12d) (BuildId: 4e306825df357f9b661479a3f9d24a8dbf960c1f)
    #10 0x7f1a0de2b1f8 in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x2b1f8) (BuildId: 4e306825df357f9b661479a3f9d24a8dbf960c1f)
    #11 0x5626382446f4 in _start /home/abuild/rpmbuild/BUILD/glibc-2.41-build/glibc-2.41/csu/../sysdeps/x86_64/start.S:115

0x52100006fca8 is located 1960 bytes inside of 4096-byte region [0x52100006f500,0x521000070500)
freed by thread T0 here:
    #0 0x5626382e6685 in free /home/abuild/rpmbuild/BUILD/llvm19-19.1.7-build/llvm-19.1.7.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cpp:52:3
    #1 0x56263836c0fd in spa_pod_dynamic_builder_clean /usr/include/spa-0.2/spa/pod/dynamic.h:64:3
    #2 0x56263836c940 in xdpw_pwr_stream_create /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/pipewire_screencast.c:633:2
    #3 0x56263833c210 in start_screencast /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/screencast.c:202:2
    #4 0x56263833a6f6 in method_screencast_start /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/screencast.c:555:9
    #5 0x7f1a0e4381ab  (/lib64/libsystemd.so.0+0x611ab) (BuildId: 1f93369ebde578ed090a17a2affdb2f4b8d2302c)

previously allocated by thread T0 here:
    #0 0x5626382e6d59 in realloc /home/abuild/rpmbuild/BUILD/llvm19-19.1.7-build/llvm-19.1.7.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cpp:82:3
    #1 0x56263836e174 in spa_pod_dynamic_builder_overflow /usr/include/spa-0.2/spa/pod/dynamic.h:39:18
    #2 0x562638370d36 in spa_pod_builder_raw /usr/include/spa-0.2/spa/pod/builder.h:140:4
    #3 0x56263836f983 in spa_pod_builder_prop /usr/include/spa-0.2/spa/pod/builder.h:444:9
    #4 0x56263836ef16 in build_format /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/pipewire_screencast.c:103:3
    #5 0x56263836b4f4 in build_formats /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/pipewire_screencast.c:173:21
    #6 0x56263836c934 in xdpw_pwr_stream_create /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/pipewire_screencast.c:632:2
    #7 0x56263833c210 in start_screencast /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/screencast.c:202:2
    #8 0x56263833a6f6 in method_screencast_start /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/screencast.c:555:9
    #9 0x7f1a0e4381ab  (/lib64/libsystemd.so.0+0x611ab) (BuildId: 1f93369ebde578ed090a17a2affdb2f4b8d2302c)

SUMMARY: AddressSanitizer: heap-use-after-free (/lib64/libpipewire-0.3.so.0+0x7b3a1) (BuildId: 532de3f5ebfec96c495741fbd1eb20a76cee1d99)
```

Another issue is that reloading sway config (even with no changes) a few times will cause a segfault if screen share is active. Backtrace for this one:

```
#0  wl_proxy_get_version (proxy=0x0) at ../src/wayland-client.c:2317
#1  0x0000560bc0cb1ccd in ext_image_copy_capture_session_v1_create_frame (ext_image_copy_capture_session_v1=0x0)
    at protocols/ext-image-copy-capture-v1-client-protocol.h:485
#2  0x0000560bc0cb19cc in ext_register_frame_cb (cast=0x560bfbb00700) at ../src/screencast/ext_image_copy.c:534
#3  0x0000560bc0cb19a4 in xdpw_ext_ic_frame_capture (cast=0x560bfbb00700) at ../src/screencast/ext_image_copy.c:572
#4  0x0000560bc0cb2c23 in xdpw_wlr_frame_capture (cast=0x560bfbb00700) at ../src/screencast/wlr_screencast.c:32
#5  0x0000560bc0cac1fe in main (argc=4, argv=0x7ffe48a85938) at ../src/core/main.c:286
```

And reloading sway config (even with no changes) after screensharing at least once and then stopping will cause a difference segfault. Backtrace: [gdb.txt](https://github.com/user-attachments/files/19542994/gdb.txt).
```
>>> p fm_pair
$1 = (struct xdpw_format_modifier_pair *) 0x7f8a67131000
>>> p fm_pair->fourcc
Cannot access memory at address 0x7f8a67131000
>>> p fm_pair->modifier
Cannot access memory at address 0x7f8a67131008
```

kennylevinsen:
Seems stable across output reconfigurations and recording restarts, at least with one output. There's still a few todo's and things needing testing, but we're getting there.

llyyr:
Last push fixed the issues but there's still one crash left. Here's what ASAN says

```
==256291==ERROR: AddressSanitizer: heap-use-after-free on address 0x515000002790 at pc 0x560280bb1208 bp 0x7fff079fa990 sp 0x7fff079fa988
WRITE of size 4 at 0x515000002790 thread T0
    #0 0x560280bb1207 in ext_session_buffer_size /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/ext_image_copy.c:283:34
    #1 0x7fb5c47c0b61  (/lib64/libffi.so.8+0x7b61) (BuildId: b4b459ae6782a758b439061d2b5df3baa953dff7)
    #2 0x7fb5c47bd3ea  (/lib64/libffi.so.8+0x43ea) (BuildId: b4b459ae6782a758b439061d2b5df3baa953dff7)
    #3 0x7fb5c47c016d in ffi_call (/lib64/libffi.so.8+0x716d) (BuildId: b4b459ae6782a758b439061d2b5df3baa953dff7)
    #4 0x7fb5c4b13c72 in wl_closure_invoke /usr/src/debug/wayland-1.23.1/x86_64-suse-linux/../src/connection.c:1228:2
    #5 0x7fb5c4b14428 in dispatch_event /usr/src/debug/wayland-1.23.1/x86_64-suse-linux/../src/wayland-client.c:1674:3
    #6 0x7fb5c4b147aa in dispatch_queue /usr/src/debug/wayland-1.23.1/x86_64-suse-linux/../src/wayland-client.c:1820:3
    #7 0x7fb5c4b147aa in wl_display_dispatch_queue_pending /usr/src/debug/wayland-1.23.1/x86_64-suse-linux/../src/wayland-client.c:2062:8
    #8 0x560280b8ebe4 in main /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/core/main.c:253:10
    #9 0x7fb5c442b12d in __libc_start_call_main (/lib64/libc.so.6+0x2b12d) (BuildId: 4e306825df357f9b661479a3f9d24a8dbf960c1f)
    #10 0x7fb5c442b1f8 in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x2b1f8) (BuildId: 4e306825df357f9b661479a3f9d24a8dbf960c1f)
    #11 0x560280aa96f4 in _start /home/abuild/rpmbuild/BUILD/glibc-2.41-build/glibc-2.41/csu/../sysdeps/x86_64/start.S:115

0x515000002790 is located 400 bytes inside of 480-byte region [0x515000002600,0x5150000027e0)
freed by thread T0 here:
    #0 0x560280b4b685 in free /home/abuild/rpmbuild/BUILD/llvm19-19.1.7-build/llvm-19.1.7.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cpp:52:3
    #1 0x560280b9a328 in xdpw_screencast_instance_destroy /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/screencast.c:104:2
    #2 0x560280bace76 in xdpw_ext_ic_frame_capture /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/ext_image_copy.c:585:3
    #3 0x560280bb6862 in xdpw_wlr_frame_capture /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/wlr_screencast.c:32:3
    #4 0x560280b8f21e in main /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/core/main.c:286:5
    #5 0x7fb5c442b12d in __libc_start_call_main (/lib64/libc.so.6+0x2b12d) (BuildId: 4e306825df357f9b661479a3f9d24a8dbf960c1f)

previously allocated by thread T0 here:
    #0 0x560280b4bb25 in calloc /home/abuild/rpmbuild/BUILD/llvm19-19.1.7-build/llvm-19.1.7.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cpp:75:3
    #1 0x560280b9befb in setup_target /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/screencast.c:185:47
    #2 0x560280b9e825 in method_screencast_select_sources /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/screencast.c:455:36
    #3 0x7fb5c4a571ab  (/lib64/libsystemd.so.0+0x611ab) (BuildId: 1f93369ebde578ed090a17a2affdb2f4b8d2302c)

SUMMARY: AddressSanitizer: heap-use-after-free /home/llyyr/workspace/xdg-desktop-portal-wlr/build/../src/screencast/ext_image_copy.c:283:34 in ext_session_buffer_size
```

I trigger is consistently by starting a screencast, stopping, then starting a new screencast and reloading sway config.

kennylevinsen:
I refactored the shutdown and error behavior a bit to simplify things, avoiding having to track "teardown" states and whatnot. Seems to work better. The history is also cleaned up.

Notes:
1. wl_shm is still untated - how was the existing screencopy impl tested?
2. Tried to test output transforms, but neither screencopy on master or ext-image-copy-capture on this branch yielded any transform handling in my testing. Do we have a client that we know handles transforms?

kennylevinsen:
Once this is in we can consider building `SPA_META_Cursor` support on top.

llyyr:
I've tested and tried very hard to make it do unexpected things, but usage-wise this seems pretty robust to me now. Thanks for the work!

src/screencast/wlr_screencopy.c
```diff
@@ -178,13 +182,10 @@ static void wlr_frame_damage(void *data, struct zwlr_screencopy_frame_v1 *frame,
...
-	logprint(TRACE, "wlroots: damage %"PRIu32": %"PRIu32",%"PRIu32"x%"PRIu32",%"PRIu32, cast->current_frame.damage_count, x, y, width, height);
-	struct xdpw_frame_damage damage = {x, y, width, height};
-	if (cast->current_frame.damage_count < 4) {
-		cast->current_frame.damage[cast->current_frame.damage_count++] = damage;
-	} else {
-		cast->current_frame.damage[3] = merge_damage(&cast->current_frame.damage[3], &damage);
-	}
+	struct xdpw_buffer *buffer = cast->current_frame.xdpw_buffer;
+	logprint(TRACE, "wlroots: damage %"PRIu32": %"PRIu32",%"PRIu32"x%"PRIu32",%"PRIu32, buffer->damage.size, x, y, width, height);
+	struct xdpw_frame_damage *damage = wl_array_add(&buffer->damage, sizeof *damage);
```
emersion, line 168: Style nit: missing parentheses after `sizeof`.

kennylevinsen:
> `xdpw_buffer_constraints` adds a lot of complexity. Do we really need that complexity?

I have simplified it a bit - there were a bunch of checks to avoid calling `pw_update_stream_params` if there weren't any *practical* difference to the constraint, which might have been unnecessary complexity.

I do think it makes sense to have the dedicated struct to accumulate constraints for both the supported protocols, and it allows us to just do a simple comparison for whether a buffer was allocated from the current constriants rather than trying to figure out *compatible* with the constraint. That's not to say that it can't be made simpler, especially if/when we do away with screencopy.

Note that our current behaviour when receiving a new device might not be correct, as we probably need to refcount the old gbm device and avoid destroying it until its BO's are gone. That, or violate protocol and compare the dev_t to avoid opening new gbm devices unnecessarily.

emersion:
My main beef with the complexity is the comparison stuff, the dirty bits and `xdpw_buffer_constraints_move()`. We receive all new constraints whenever they change, so we don't need to keep track of what to update.

> it allows us to just do a simple comparison for whether a buffer was allocated from the current constriants rather than trying to figure out if it is compatible with the constraint

Checking whether it is compatible with the new constraint sounds better to me: if a new format was added by the compositor, we don't need to reallocate.

kennylevinsen:
> My main beef with the complexity is the comparison stuff, the dirty bits and xdpw_buffer_constraints_move()

Note that I removed the comparison stuff earlier, and I just now simplified `_move` further and turned the dirty bitmask into a bool instead.

We can do away with double-buffering, but then we have to add conditionals to various the constraint event handlers to conditionally wipe the existing constraints if it's the first call after "done" - just filling out a temp struct felt cleaner, and the difference is (now) quite negligible.

> Checking whether it is compatible with the new constraint sounds better to me: if a new format was added by the compositor, we don't need to reallocate.

I don't think that changes much with respect to the buffer constraints struct, but I can try to give that a go again. It will definitely be more code than the integer `constraint_id` comparison we have now though.

One caveat with that approach is that ext-image-copy-capture explicitly disallows any equality tests on the sent `dev_t` value, and simultaneously specify that buffers *must* be allocated with the sent device. We would have to write more code to identify the device and check if it's the same to be able to reuse a prior buffer. We won't be able to benefit from any constraint checks without either ignoring this or writing the extra code to solve it.

(It *might* also be better to send the constraints off to pipewire whenever they changed, even if only additive as the sink might benefit from a newly available format, but that's ultimately an optimization.)

emersion:
> We can do away with double-buffering, but then we have to add conditionals to various the constraint event handlers to conditionally wipe the existing constraints if it's the first call after "done" - just filling out a temp struct felt cleaner, and the difference is (now) quite negligible.

Sounds good!

> I don't think that changes much with respect to the buffer constraints struct, but I can try to give that a go again. It will definitely be more code than the integer constraint_id comparison we have now though.

I'm fine with keeping the current logic.

src/screencast/screencast_common.c
```diff
@@ -417,3 +401,110 @@ struct xdpw_frame_damage merge_damage(struct xdpw_frame_damage *damage1, struct
...
+			sizeof(struct xdpw_shm_format),
+			compare_shm_format);
+}
+
+bool xdpw_buffer_constraints_move(struct xdpw_buffer_constraints *dst, struct xdpw_buffer_constraints *src) {
+	if (!src->dirty) {
+		return false;
+	}
+	int dirty = src->dirty;
+	xdpw_buffer_constraints_sort(src);
```
emersion, line 485: Do we need to sort the formats?

kennylevinsen: Not anymore, no - removed!

(There's technically a gap between ext_image_capture_copy_capture being strictly unordered and pipewire *maybe* being ordered by preference? But let's deal with that if we hit it.)

src/screencast/ext_image_copy.c
```diff
@@ -0,0 +1,332 @@
...
+			cast->ctx->ext_output_image_capture_source_manager == NULL) {
+		logprint(INFO, "ext: unsupported");
+		return -1;
+	}
+	ext_register_session_cb(cast);
+
+	// process at least one frame so that we know
+	// some of the metadata required for the pipewire
+	// remote state connected event
+	wl_display_dispatch(cast->ctx->state->wl_display);
```
emersion, line 328: Nit: this dispatch is unnecessary

src/screencast/ext_image_copy.c
```diff
@@ -0,0 +1,332 @@
...
+
+	switch (reason) {
+	case EXT_IMAGE_COPY_CAPTURE_FRAME_V1_FAILURE_REASON_UNKNOWN:
+		logprint(ERROR, "ext: frame capture failed: unknown reason");
+		break;
+	case EXT_IMAGE_COPY_CAPTURE_FRAME_V1_FAILURE_REASON_BUFFER_CONSTRAINTS:
+		logprint(ERROR, "ext: frame capture failed: buffer constraint mismatch");
+		break;
+	case EXT_IMAGE_COPY_CAPTURE_FRAME_V1_FAILURE_REASON_STOPPED:
+		logprint(ERROR, "ext: frame capture failed: capture session stopped");
```
emersion, line 236: Maybe this should be INFO?

src/screencast/ext_image_copy.c
```diff
@@ -0,0 +1,332 @@
...
+	default:
+		abort();
+	}
+
+	if (cast->ext_session.frame) {
+		ext_image_copy_capture_frame_v1_destroy(cast->ext_session.frame);
+		cast->ext_session.frame = NULL;
+	}
+
+	xdpw_pwr_enqueue_buffer(cast);
```
emersion, line 247: Shouldn't we stop the cast unless the reason was `BUFFER_CONSTRAINTS`?

src/screencast/ext_image_copy.c
```diff
@@ -0,0 +1,332 @@
...
+		struct ext_image_copy_capture_frame_v1 *ext_image_copy_capture_frame_v1,
+		int32_t x, int32_t y, int32_t width, int32_t height) {
+	struct xdpw_screencast_instance *cast = data;
+
+	logprint(TRACE, "ext: damage: %"PRId32",%"PRId32"x%"PRId32",%"PRId32, x, y, width, height);
+
+	// Our damage tracking
+	struct xdpw_buffer *buffer;
+	wl_list_for_each(buffer, &cast->buffer_list, link) {
+		struct xdpw_frame_damage *damage = wl_array_add(&buffer->damage, sizeof *damage);
```
emersion, line 187: Nit: missing parentheses

kennylevinsen:
> The last few changes resurface a bug that I'd mentioned and was fixed before.
> 
> * Reloading sway config (or doing "swaymsg output * power off" then on instantly) will break the capture with `pipewire: unable to export buffer`

Ah yeah, I recall that issue: we can't force pipewire to reallocate if it doesn't want to, so checking constraints of buffers it hands us is fairly pointless. We tell pipewire when the constraints change, and the Wayland server will tell us gracefully if a buffer is a no go, so let's just skip the whole constraint ID check altogether.

Less code, more betterer.

(The previous solution was the whole constraint comparison thing, but why complicate things if we don't have to?)

src/screencast/ext_image_copy.c
kennylevinsen: Fair, removed

src/screencast/ext_image_copy.c
```diff
@@ -0,0 +1,332 @@
...
+		logprint(INFO, "ext: unsupported");
+		return -1;
+	}
+	ext_register_session_cb(cast);
+
+	// process at least one frame so that we know
+	// some of the metadata required for the pipewire
+	// remote state connected event
+	wl_display_dispatch(cast->ctx->state->wl_display);
+	wl_display_roundtrip(cast->ctx->state->wl_display);
```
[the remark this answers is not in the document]
kennylevinsen: Done

src/screencast/ext_image_copy.c
kennylevinsen: Yeah, makes sense.

llyyr:
`pipewire: Prepare for supporting multiple formats` breaks capture with wlr-screencopy now.

kennylevinsen:
> `pipewire: Prepare for supporting multiple formats` breaks capture with wlr-screencopy now.

Hmm, the difference up to that commit before and after the recent work is just a few style nits. Reproducible or just a one-off?

PR #329, opened 2025-06-06
Toplevel capture

kennylevinsen:
Quickly hacked together draft.

TODO:
- [ ] Fix up selection, making the exisitng output selection more general
- [ ] Figure out why pipewire seems to ignore geometry changes

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -116,7 +116,7 @@ int main(int argc, char *argv[]) {
 		.bus = bus,
 		.wl_display = wl_display,
 		.pw_loop = pw_loop,
-		.screencast_source_types = MONITOR,
+		.screencast_source_types = MONITOR | WINDOW,
 		.screencast_cursor_modes = HIDDEN | EMBEDDED,
 		.screencast_version = XDP_CAST_PROTO_VER,
 		.screenshot_version = XDP_SHOT_PROTO_VER,
--- a/src/screencast/ext_image_copy.c
+++ b/src/screencast/ext_image_copy.c
@@ -34,6 +34,7 @@ static void ext_session_buffer_size(void *data,
 	cast->pending_constraints.width = width;
 	cast->pending_constraints.height = height;
 	cast->pending_constraints.dirty = true;
+	logprint(TRACE, "ext: buffer_size: %d, %d", width, height);
 }
 
 static void ext_session_shm_format(void *data,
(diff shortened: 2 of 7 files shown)


kennylevinsen:
> Has support for other menus not yet been implemented?

That's the first of the two TODO's, it's hardcoded to just the wofi option right now.

src/screencast/ext_image_copy.c
```diff
@@ -237,6 +238,7 @@ static void ext_frame_failed(void *data,
 		return;
 	case EXT_IMAGE_COPY_CAPTURE_FRAME_V1_FAILURE_REASON_BUFFER_CONSTRAINTS:
 		logprint(ERROR, "ext: frame capture failed: buffer constraint mismatch");
+		pwr_update_stream_param(cast);
```
emersion, line 241: Maybe this should be in a separate commit?

src/screencast/screencast.c
```diff
@@ -164,7 +169,7 @@ bool setup_target(struct xdpw_screencast_context *ctx, struct xdpw_session *sess
 		free(target);
 		return false;
 	}
-	assert(target->output);
+	assert(target->output || target->handle);
```
emersion, line 172: This accesses members of a union before checking the `type` field.

TBH, I think it'd be better to drop the union here to avoid these footguns.

src/screencast/wlr_screencast.c
```diff
@@ -558,6 +698,76 @@ static const struct zwp_linux_dmabuf_feedback_v1_listener linux_dmabuf_listener_
...
+	}
+
+	toplevel->handle = handle;
+	wl_list_insert(&ctx->toplevels, &toplevel->link);
+	ext_foreign_toplevel_handle_v1_add_listener(handle, &foreign_toplevel_handle_listener, toplevel);
+}
+
+static void foreign_toplevel_list_handle_finished(void *data,
+		struct ext_foreign_toplevel_list_v1 *list) {
+	// Clean up object
```
emersion, line 763: Is this a TODO?

src/core/main.c
```diff
@@ -116,7 +116,7 @@ int main(int argc, char *argv[]) {
 		.bus = bus,
 		.wl_display = wl_display,
 		.pw_loop = pw_loop,
-		.screencast_source_types = MONITOR,
+		.screencast_source_types = MONITOR | WINDOW,
```
emersion, line 119: Probably we should set this depending on available protocols.

emersion:
This should help making it easier to call a chooser for toplevels: https://github.com/emersion/xdg-desktop-portal-wlr/pull/333

src/screencast/wlr_screencast.c
```diff
@@ -742,6 +887,7 @@ int xdpw_wlr_screencopy_init(struct xdpw_state *state) {
 
 	// initialize a list of active screencast instances
 	wl_list_init(&ctx->screencast_instances);
+	wl_list_init(&ctx->toplevels);
```
emersion, line 890: This belongs to the previous commit.

emersion:
I've reworked this PR here: https://github.com/emersion/xdg-desktop-portal-wlr/pull/334

PR #330, opened 2025-06-06
Fix wlr_screencopy regression

kennylevinsen:
The new xdpw buffer constraint system did not account for wlr_screencopy sending the same constraints on every single frame.

In wlr_screencopy, check if a buffer matches the current constraints and only complain if it does not. This is the original behavior, albeit checking the allocated buffer details rather than focusing on the configured pipewire details.

(This is unfortunately not portable to ext_image_copy without sorting out dmabuf device comparison.)

Fixes: https://github.com/emersion/xdg-desktop-portal-wlr/pull/326#issuecomment-2948355389

### DIFF
--- a/include/screencast_common.h
+++ b/include/screencast_common.h
@@ -73,6 +73,7 @@ struct xdpw_buffer {
 	uint32_t height;
 	uint32_t format;
 	int plane_count;
+	uint64_t modifier;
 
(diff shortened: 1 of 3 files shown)


emersion:
Sounds like Clang is unhappy in CI.

emersion (approved):
Thanks!

PR #331, opened 2025-06-12
Add wmenu chooser

emersion:
This is used from Sway's default config file, so is more likely to be installed.


xdg-desktop-portal-wlr.5.scd
```diff
@@ -64,7 +64,7 @@ These options need to be placed under the **[screencast]** section.
 
 	The supported types are:
 	- default: xdpw will try to use the first chooser found in the list of hardcoded choosers
-	  (slurp, wofi, bemenu) and will fallback to an arbitrary output if none of those were found.
+	  (slurp, wofi, wmenu, bemenu) and will fallback to an arbitrary output if none of those were found.
```
llyyr, line 67: This should probably match the order in the code

PR #335, opened 2025-07-28
Add rofi chooser

alternateved:
Hello!

It seems that [fork of rofi with Wayland support](https://github.com/lbonn/rofi) is still a popular chooser to the point that official Sway spin on Fedora is using it as default. Would you consider adding it to supported choosers as well?


emersion:
Hm, I'm not sure we should be adding tools which don't have upstream Wayland support…

emersion:
Ah nice!

emersion (approved):

PR #342, opened 2025-08-20
Throttle fps from start of frame capture

kennylevinsen:
The fps limit logic was throttling the time between the completion of one recorded frame and the request for the next frame. This means that the time it takes to request the frame, including if the compositor stalls for a frame to be available, is not measured as part of the current frame rate.

Measure from start of frame till start of frame.

(I didn't run into a particular issue here including in test, just seemed wrong.)


emersion (approved):
Thanks!

PR #343, opened 2025-08-21
Refcount gbm device and only open if it changed

kennylevinsen:
Whenever buffer constraints change, we will be re-informed of the main device even if it hasn't changed. This means that during resize of a toplevel capture, we will constantly be creating and destroying gbm devices for no good reason.

To avoid becoming a stress-test of driver init and teardown paths, wrap the gbm device in a refcounted container and ref it if the drm device hasn't changed.

### DIFF
--- a/include/screencast_common.h
+++ b/include/screencast_common.h
@@ -48,6 +48,12 @@ struct xdpw_chooser {
 	const char *cmd;
 };
 
+struct xdpw_gbm_device {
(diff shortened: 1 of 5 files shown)


kennylevinsen:
That's a good idea, did that in: https://github.com/emersion/xdg-desktop-portal-wlr/pull/345

PR #344, opened 2025-08-22
pipewire: Allow 1 to 2 buffers instead of 2 to 32

kennylevinsen:
Pipewire negotiates the amount of buffers that we will need to allocate with the client. If a client asks for between 1 and 32 with a default of 8, we end up allocating 8 because we allow it and our default is lower.

We don't really benefit from having more than 2 buffers in rotation, so set that as our upper limit. Clients like libwebrtc that default to 8 throw away all buffers but the latest one anyway. The more buffers are negotiated, the more buffers we have to allocate *up front* on every constraint mismatch, and we end up being blamed in memory accounting.

Use a range of 1 to 2 with a default of 1 instead of 2 to 32 buffers.

### DIFF
--- a/contrib/config.sample
+++ b/contrib/config.sample
@@ -3,3 +3,4 @@ output_name=
 max_fps=30
 chooser_cmd=slurp -f %o -or
 chooser_type=simple
+max_buffers=2
--- a/include/config.h
+++ b/include/config.h
@@ -12,6 +12,7 @@ struct config_screencast {
 	char *chooser_cmd;
 	enum xdpw_chooser_types chooser_type;
 	bool force_mod_linear;
+	uint32_t max_buffers;
 };
 
 struct xdpw_config {
--- a/include/pipewire_screencast.h
+++ b/include/pipewire_screencast.h
@@ -3,8 +3,8 @@
 
 #include "screencast_common.h"
 
-#define XDPW_PWR_BUFFERS 2
-#define XDPW_PWR_BUFFERS_MIN 2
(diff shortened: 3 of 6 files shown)


kennylevinsen:
I can't say for sure if we're supposed to have a default of 2 because we work best with two buffers but accept clients with 1 buffer, or if we're supposed to have a default of 1 to not force clients into two buffers that would prefer 1 buffer.

We can always adjust it again later.

emersion:
> We don't really benefit from having more than 2 buffers in rotation

I'm not sure about this. If a client pipes the buffers to a slow encoder, having 8 buffers or so might allow for a smooth capture (with most of the buffers busy in the encoder queue) even if the final stream is delayed. Another case is when a CPU encoder is used and the CPU occasionally spikes doing something else.

kennylevinsen:
Do we have any examples of such use?

I am not sure if our current setup can use more than one buffer at a time, waking up on `process` when the client is done with the single buffer we sent. Maybe if the sink became a driver and manually called `pw_stream_trigger_process` every time it dequeues? We do have plenty of buffers available to dequeue at the time of process instead of the one we use right now, but if we were to keep running on our own we'd need to pace frames.

I considered lazy allocation, but I"m not sure if that's allowed. If there's a supported need for more buffers we could also consider a configurable limit. If we can't ever *use* more than one or two buffers, we shouldn't offer it though.

emersion:
Ref https://github.com/emersion/xdg-desktop-portal-wlr/pull/397: our buffer starvation codepath is really not optimal.

PR #345, opened 2025-08-24
Only reallocate gbm device if it changed

kennylevinsen:
We are notified of the dmabuf_device every time buffer constraints change. Currently, this means that we destroy and reinitialize the gbm device (and possibly the driver backing it) every time constraints changed, which for toplevel capture is quite often during e.g., resize.

Instead, just store the dev_t, and when the constraints are applied, only reinitialize the gbm device if drmDevicesEqual reports that we are dealing with a different device.


emersion (approved):
Nice, thank you!

Issue #337, opened 2025-08-20
Segmentation fault upon window sharing


sewnie:
```
2025/08/20 13:10:33 [INFO] - dbus: app_id: 
2025/08/20 13:10:33 [INFO] - dbus: parent_window: 
2025/08/20 13:10:33 [DEBUG] - dbus: start: found matching session /org/freedesktop/portal/desktop/session/1_80/webrtc_session112397062
Process 15662 stopped
* thread #1, name = 'xdg-desktop-por', stop reason = signal SIGSEGV: address not mapped to object (fault address=0x40)
    frame #0: 0x00007ffff7e6fd14 libwayland-client.so.0`wl_proxy_get_version + 4
libwayland-client.so.0`wl_proxy_get_version:
->  0x7ffff7e6fd14 <+4>: movl   0x40(%rdi), %eax
    0x7ffff7e6fd17 <+7>: popq   %rbp
    0x7ffff7e6fd18 <+8>: retq   
    0x7ffff7e6fd19:      int3   
(lldb) bt
* thread #1, name = 'xdg-desktop-por', stop reason = signal SIGSEGV: address not mapped to object (fault address=0x40)
  * frame #0: 0x00007ffff7e6fd14 libwayland-client.so.0`wl_proxy_get_version + 4
    frame #1: 0x0000555555565a5a xdg-desktop-portal-wlr`ext_register_session_cb(cast=0x00007ffff3862600) at ext_image_copy.c:0:2
    frame #2: 0x00005555555659cb xdg-desktop-portal-wlr`xdpw_ext_ic_session_init(cast=<unavailable>) at ext_image_copy.c:334:2
    frame #3: 0x0000555555566170 xdg-desktop-portal-wlr`xdpw_wlr_session_init(cast=<unavailable>) at wlr_screencast.c:48:4 [artificial]
    frame #4: 0x0000555555563c69 xdg-desktop-portal-wlr`start_screencast(cast=<unavailable>) at screencast.c:227:8
    frame #5: 0x0000555555563892 xdg-desktop-portal-wlr`method_screencast_start(msg=0x00007ffff38e4f00, data=0x00007fffffffe3e0, ret_error=<unavailable>) at screencast.c:585:9
    frame #6: 0x00007ffff7dfaca3 libelogind.so.0`___lldb_unnamed_symbol1189 + 2243
[...]

PR #338, opened 2025-08-20
Check for ext protocol support

sewnie:
With the toplevel protocol being supported, no NULL check was put in place for it if a compositor does not support it. Here, the check was rewritten to check for the screencast type's required protocol.

Fixes #337

PR #346, opened 2025-09-05
screencast: Restrict source types to those supported

kennylevinsen:
A client can specify a type mask with window capture, monitor capture, or both. The prerequisites for window capture may not be met even if the client requests it.

Ensure we only attempt to set up types we support.

Fixes: https://github.com/emersion/xdg-desktop-portal-wlr/issues/337

Alternative to https://github.com/emersion/xdg-desktop-portal-wlr/pull/338


opoku (approved):
Thanks! This should provide a more useful error message for unsupported source types.

PR #347, opened 2025-09-16
announce wlr for use in volare

raboof:
volare is a Sway fork https://codeberg.org/raboof/volare

I understand wlr.portal isn't the 'modern' way to find portal implementations anymore
(https://github.com/emersion/xdg-desktop-portal-wlr/pull/295) but this was an easy way to make screen sharing work. Can you point me in the right direction for the 'proper' solution? :)


emersion:
See this section of the README for the preferred way to configure portals: https://github.com/emersion/xdg-desktop-portal-wlr?tab=readme-ov-file#running

raboof:
thanks for the pointer, can confirm that works just fine, so let's close this.

PR #353, opened 2025-11-17
build: add default for systemd user_unit_dir

fossdd:
Allows building the service files without a systemd dependency by defining a default systemd user unit directory.

Relevant in Alpine, where we can package the service files in a subpackage without having systemd in Alpine.


emersion:
Why not install systemd at build-time to allow xdpw to discover the path properly?

fossdd:
I'm currently enabling a dozen of systemd services in Alpine, and it looks to be a popular way used, yes. If you'd like that, I can do that here aswell.

Issue #278, opened 2023-08-10
Add support for InputCapture portal


theofficialgman:
InputCapture portal spec has been finalized and merged in xdg-desktop-portal https://github.com/flatpak/xdg-desktop-portal/pull/714
It has also been implemented and merged into xdg-desktop-portal-gnome https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/61
This isssue is a request that it be implemented in xdg-desktop-portal-kde

This portal interface, alongside implementing the RemoteDesktop portal https://github.com/emersion/xdg-desktop-portal-wlr/issues/2, is important for Barrier/Input Leap/Synergy usecase

columbarius:
This requires support for libei from compositors and a wayland protocol to relay the setup information.

PedroHLC:
Last week, I finally tried KVMing from input-leap (in GNOME) to waynergy (in Sway), it wasn't as perfect as Synergy was for X11⇾X11, but it was good enough.

I did try to use #263, but it doesn't have the ConnectToEI bits and wasn't enough for input-leap to work as a client (but it does have enough of #2 to work with KDE Connect, as I stated there before).

Waynergy is fine for this scenario, but I'd prefer not to use GNOME on any of these machines.

Sad to see it stalled here, Synergy exists for 22 years now and its UX is battle-proved, while input-leap through libei provides that wonderfully.

sithlord48:
Hello the current libei has support and input-capture is working in the kde and gnome portals. 

Is there any chance of this one being added ?

PR #359, opened 2025-12-05
Implemented InputCapture protocol (issue #278)

aburtasenkov:
Hey all,

I'm submitting the implementation for the org.freedesktop.portal.InputCapture feature, which addresses Issue #278. This is the code that lets applications (like remote desktop tools or VMs) securely grab and inject keyboard/mouse input on wlroots compositors.

How It Works:
The core idea is to use libei (Event Interception) for the data stream and several wlroots protocols to handle the necessary input lockdown:

- An invisible layer-shell surface takes focus to capture all events.

- Pointer constraints lock the cursor inside.

- Keyboard inhibitors prevent system shortcuts (like Mod+Enter) from breaking the session.

Testing Environment: Functional success was verified by using the deskflow client communicating over the network with this build running in a dedicated Sway VM.

I know this feature has been requested for a while. I'm looking forward to getting this merged and seeing it in action. Please let me know if you spot anything that needs attention!

Cheers, 
Anton

### DIFF
--- a/contrib/wlroots-portals.conf
+++ b/contrib/wlroots-portals.conf
@@ -4,4 +4,5 @@ default=gtk
 # ... except for the ScreenCast, Screenshot and Settings (dark/light mode) interface
 org.freedesktop.impl.portal.ScreenCast=wlr
 org.freedesktop.impl.portal.Screenshot=wlr
+org.freedesktop.impl.portal.InputCapture=wlr
 org.freedesktop.impl.portal.Settings=darkman
--- a/include/input_capture.h
+++ b/include/input_capture.h
@@ -1,0 +1,58 @@
+#pragma once
+
+#include "xdpw.h"
+
+#include <systemd/sd-bus.h>
+#include <libei-1.0/libeis.h>
+#include <wayland-client.h>
+#include <wayland-util.h>
+#include <xkbcommon/xkbcommon.h>
+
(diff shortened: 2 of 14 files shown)


grazzolini:
Ok, I modified Arch Linux PKGBUILD for xdg-desktop-portal-wlr to build this and the first thing I found is that we're missing the libei requirement. It is used, but it is marked as `required: false`. This is also why the CI failed. After fixing that on the PKGBUILD (add libei as a dependency), I was able to get a patched build.

Tested with deskflow, I already had a working configuration on plasma. So, If I start deskflow in client mode, it keeps on a loop trying to start the deskflow-core in client mode. In server mode, I can get a client to connect, however, there was no layout I tested that allowed me to move the mouse to the other machine.

waflessnet:
Had a similar experience on Gentoo/Sway: I had to manually adjust the ebuild to make sure libei and libportal were included, as Meson tends to skip them silently if they aren't found.

About the server-mode mouse problem: Testing on Sway shows that the input_capture barrier isn't activating properly. It feels like the capture overlay isn't hitting the right spot in the layer-shell hierarchy and stays hidden behind other layers.

Even after forcing some borders in my Sway layout to trigger it, the portal won't send the event over to libei

KalilDev:
Edit: sorry, commented in the wrong PR

PR #361, opened 2025-12-13
screencast: fix damage reporting via SPA

cptpcrd:
I was recently using an application that carefully inspects and uses the damage regions reported by `xdg-desktop-portal-wlr`, and observed that they were quite broken. I believe this PR fixes all the issues I'm aware of, though I admit to not being an expert in this particular ecosystem.

I believe that commit c14076afd39507e9713394a92f491f4858dfc067 (part of #326) introduced several regressions in `xdpw_pwr_enqueue_buffer`:
1. The first damage region reported to PipeWire via SPA is always left untouched (`d_region++` is run in the loop _before_ assigning to `d_region`). This wastes one of the usually-16 available slots.
2. If `xdg-desktop-portal-wlr` has collected more than 16 damage regions for a particular frame, anything after the 16th damage region or so is not properly included. It appears that the code intends to:

    a. Make a pass through the damage list, incrementing `damage_counter` for each rectangle that it reports as-is to SPA, stopping when it runs out of SPA slots.
    b. On running out of slots, make another pass through the damage list, decrementing `damage_counter`. When it reaches 0 or goes negative, we've reached the remaining damage regions that weren't included, and they are all combined into the last "slot".

   The problem is that `damage_counter` is a `uint32_t`, so it wraps around and becomes positive again, and skips many of the damage regions.
4. I couldn't find an authoritative specification for how the damage information is written, but a comment in the SPA headers says:
    ```
    /**< array of struct spa_meta_region with damage, where an invalid entry or end-of-array marks the end. */
    ```
    Assuming that "invalid entry" means all zeros, `xdg-desktop-portal-wlr` is not properly marking the end.

    Futhermore, libwebrtc, which is probably the most noteworthy consumer of this information, does not actually stop at the first invalid entry; it instead skips over invalid entries. So it's actually safest to mark the end with a series of invalid entries.
5. If I understand the ext_image_copy_capture_v1 protocol correctly, it is indeed relevant to collect damage per-buffer (as done in c14076afd39507e9713394a92f491f4858dfc067) for the _capturing_ side. However, I don't believe this is relevant for the information that `xdg-desktop-portal-wlr` reports to PipeWire - as far as I can tell, those frames always contain full information, and the damage reported is relative to the last frame sent. Thus, when using N buffers, `xdg-desktop-portal-wlr` is collecting and reporting damage from the last ~N frames, when it only actually needs to be reporting from the last 1 frame. This doesn't impact correctness, but it wastefully damages more regions than necessary.

The first commit in this PR addresses items 1-3 by adjusting the logic in `xdpw_pwr_enqueue_buffer` slightly. The second commit addresses item 4 by additionally tracking damage on a per-frame basis, and using that information to populate the information sent to PipeWire.

### DIFF
--- a/src/screencast/ext_image_copy.c
+++ b/src/screencast/ext_image_copy.c
@@ -287,13 +287,10 @@ static void ext_register_frame_cb(struct xdpw_screencast_instance *cast) {
 
 	ext_image_copy_capture_frame_v1_attach_buffer(cast->ext_session.frame,
 			cast->current_frame.xdpw_buffer->buffer);
-	struct xdpw_buffer *buffer;
-	wl_list_for_each(buffer, &cast->buffer_list, link) {
-		struct xdpw_frame_damage *damage;
-		wl_array_for_each(damage, &buffer->damage) {
-			ext_image_copy_capture_frame_v1_damage_buffer(
-					cast->ext_session.frame, damage->x, damage->y, damage->width, damage->height);
-		}
+	struct xdpw_frame_damage *damage;
+	wl_array_for_each(damage, &cast->current_frame.xdpw_buffer->damage) {
+		ext_image_copy_capture_frame_v1_damage_buffer(
+				cast->ext_session.frame, damage->x, damage->y, damage->width, damage->height);
 	}
 	ext_image_copy_capture_frame_v1_capture(cast->ext_session.frame);
 


cptpcrd:
Sorry about the extended delay. I believe I've addressed your initial feedback.

I'll look into opening a libwebrtc bug later.

PR #365, opened 2026-01-15
include fuzzel as chooser

nabnut:
This adds fuzzel (https://codeberg.org/dnkl/fuzzel) to the list of choosers


emersion (approved):
Thanks!

PR #370, opened 2026-02-21
screencast: drive the Pipewire graph by ourselves

### LINKED ISSUE #351: Screencast portal lags around 30fps with new version
When updating xdg-desktop-portal-wlr to 0.8.0, my screen recording in OBS (Pipewire), is extremely unsmooth. All settings are the same as before but the screen capture looks like 30fps while the camera and the video itself remain smooth at 60fps.

Downgrading to 0.7.1 resolves the issue completely. Besides that no config change was able to fix this.

funk443:
Should fix #351, this issue affects any screenshare application that uses Pipewire to capture the screen.

According to my research, this is due to the driver assigned by Pipewire for xdpw is driving the graph at an incorrect rate, `1024/48000` by default, which translates to about 40 fps. As a result, the `pwr_handle_stream_on_process` handler function in `pipewire_screencast.c` is not called frequent enough to capture sufficient amount of frames for a smooth output.

By specifying the max latency while creating the Pipewire node, Pipewire should be able to setup the driver correctly to drive the graph at the correct rate, thus fixing this issue.

### DIFF
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -16,6 +16,7 @@
 #include "wlr_screencast.h"
 #include "xdpw.h"
 #include "logger.h"
+#include "timespec_util.h"
 
 #define DAMAGE_REGION_COUNT 16
 
@@ -328,6 +329,31 @@ done:
 	cast->current_frame.pw_buffer = NULL;
 }
 
+static void pwr_capture_one(void *data) {
+	struct xdpw_screencast_instance *cast = data;
+
+	if (!cast->pwr_stream_state) {
+		logprint(INFO, "pipewire: not streaming");
+		return;
+	}
+
+	if (cast->current_frame.pw_buffer) {
+		logprint(DEBUG, "pipewire: buffer already exported");
+		goto trigger_graph;
+	}
+
+	xdpw_pwr_dequeue_buffer(cast);
+	if (!cast->current_frame.pw_buffer) {
+		logprint(WARN, "pipewire: unable to export buffer");
+		goto trigger_graph;
+	}
+
+	xdpw_wlr_frame_capture(cast);
+
+trigger_graph:
+	pw_stream_trigger_process(cast->stream);
+}
+
 void pwr_update_stream_param(struct xdpw_screencast_instance *cast) {
 	logprint(TRACE, "pipewire: stream update parameters");
 	struct pw_stream *stream = cast->stream;
@@ -359,6 +385,7 @@ static void pwr_handle_stream_state_changed(void *data,
 	switch (state) {
 	case PW_STREAM_STATE_STREAMING:
 		cast->pwr_stream_state = true;
+		pwr_capture_one(cast);
 		break;
 	case PW_STREAM_STATE_PAUSED:
 		if (old == PW_STREAM_STATE_STREAMING) {
@@ -586,27 +613,13 @@ static void pwr_handle_stream_remove_buffer(void *data, struct pw_buffer *buffer
 	buffer->user_data = NULL;
 }
 
-static void pwr_handle_stream_on_process(void *data) {
+static void pwr_handle_stream_trigger_done(void *data) {
 	struct xdpw_screencast_instance *cast = data;
-
-	logprint(TRACE, "pipewire: on process event handle");
-
-	if (!cast->pwr_stream_state) {
-		logprint(INFO, "pipewire: not streaming");
-		return;
+	uint64_t delay_ns = fps_limit_measure_end(&cast->fps_limit, cast->framerate);
+	if (delay_ns <= 0) {
+		delay_ns = (1.0 / cast->framerate) * TIMESPEC_NSEC_PER_SEC;
 	}
-
-	if (cast->current_frame.pw_buffer) {
-		logprint(DEBUG, "pipewire: buffer already exported");
-		return;
-	}
-
-	xdpw_pwr_dequeue_buffer(cast);
-	if (!cast->current_frame.pw_buffer) {
-		logprint(WARN, "pipewire: unable to export buffer");
-		return;
-	}
-	xdpw_wlr_frame_capture(cast);
+	xdpw_add_timer(cast->ctx->state, delay_ns, pwr_capture_one, cast);
 }
 
 static const struct pw_stream_events pwr_stream_events = {
@@ -615,7 +628,7 @@ static const struct pw_stream_events pwr_stream_events = {
 	.param_changed = pwr_handle_stream_param_changed,
 	.add_buffer = pwr_handle_stream_add_buffer,
 	.remove_buffer = pwr_handle_stream_remove_buffer,
-	.process = pwr_handle_stream_on_process,
+	.trigger_done = pwr_handle_stream_trigger_done,
 };
 
 void xdpw_pwr_stream_create(struct xdpw_screencast_instance *cast) {
@@ -652,7 +665,8 @@ void xdpw_pwr_stream_create(struct xdpw_screencast_instance *cast) {
 	pw_stream_connect(cast->stream,
 		PW_DIRECTION_OUTPUT,
 		PW_ID_ANY,
-		PW_STREAM_FLAG_ALLOC_BUFFERS,
+		(PW_STREAM_FLAG_ALLOC_BUFFERS |
+			PW_STREAM_FLAG_DRIVER),
 		params.data, params.size / sizeof(struct spa_pod *));
 
 	spa_pod_dynamic_builder_clean(&builder);
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -16,7 +16,6 @@
 #include "wlr_screencast.h"
 #include "xdpw.h"
 #include "logger.h"
-#include "timespec_util.h"
 
 #define DAMAGE_REGION_COUNT 16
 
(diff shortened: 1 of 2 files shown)


emersion:
Thanks for tracking this down! Hm, I'm not a fan of hardcoding a particular magic number here, since users might have high refresh rate monitors.

IIRC we've switched to PipeWire for driving the source to make it so we don't request frames from the compositor if the sink isn't going to consume them all. Ideally we'd find a way to limit the rate exclusively based on the source and sink queue/dequeue rate.

funk443:
> It sounds like the real fix is going to switch outselves back to being a driver, and use lazy scheduling: https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/653b8703bc0b532a352352b1b16b11940efc79c1/doc/dox/internals/scheduling.dox#L331

This seems to be the correct way to do screencasting in Pipewire. I'll try to work on this in my free time, as I'm kinda busy right now on other stuffs.

ManuLinares:
This doesn't fix https://github.com/emersion/xdg-desktop-portal-wlr/issues/351

ManuLinares:
@emersion Now it works, in my 60hz monitor at 60fps.

I don't like you force-pushed but I can confirm this (commit 8142629da7fd5e400fd00658d5ddaa8ea78cba53) works.

https://github.com/user-attachments/assets/e501e1ec-94fb-40d1-bb0e-3c9206b5f4a5

src/screencast/pipewire_screencast.c
```diff
@@ -586,27 +613,13 @@ static void pwr_handle_stream_remove_buffer(void *data, struct pw_buffer *buffer
...
-
-	if (!cast->pwr_stream_state) {
-		logprint(INFO, "pipewire: not streaming");
-		return;
-	}
-
-	if (cast->current_frame.pw_buffer) {
-		logprint(DEBUG, "pipewire: buffer already exported");
-		return;
+	uint64_t delay_ns = fps_limit_measure_end(&cast->fps_limit, cast->framerate);
```
emersion: Why do we need to look at the FPS limit here? `xdpw_wlr_frame_capture()` already takes care of it, no?

src/screencast/pipewire_screencast.c
```diff
@@ -615,7 +628,7 @@ static const struct pw_stream_events pwr_stream_events = {
 	.param_changed = pwr_handle_stream_param_changed,
 	.add_buffer = pwr_handle_stream_add_buffer,
 	.remove_buffer = pwr_handle_stream_remove_buffer,
-	.process = pwr_handle_stream_on_process,
+	.trigger_done = pwr_handle_stream_trigger_done,
```
emersion: I am not familiar with `trigger_done`, what does it achieve?

Does `pw_stream_trigger_process()` end up calling this?

funk443: > Does `pw_stream_trigger_process()` end up calling this?

Yes.

src/screencast/pipewire_screencast.c
funk443: > `xdpw_wlr_frame_capture()` already takes care of it, no?

It did. Should I just call `pwr_capture_one` at the end of this function, instead of adding another timer? I'm still not very familiar on how xdpw timers work.

funk443: > Why do we need to look at the FPS limit here?

Basically I want a way to queue the function call every `1/fps` second.

src/screencast/pipewire_screencast.c
emersion: Why is this callback necessary? Our older driver logic didn't have this.

src/screencast/pipewire_screencast.c
emersion: Why does the function need to be called at that interval, when `xdpw_wlr_frame_capture()` already limits frame rate?

src/screencast/pipewire_screencast.c
funk443: I used this callback to schedule the next capture, now I realize we don't actually need it, since `pw_stream_trigger_process` can just be called in `ext_frame_ready`.

src/screencast/pipewire_screencast.c
funk443: Because I misunderstood how `xdpw_wlr_frame_capture()` actually works.

emersion:
Sorry for all of the questions, I'm not super familiar with PipeWire. Thanks for continuing to work on this!

funk443:
From my testing, bf6e24d should fix the stuttering problem, OBS works fine with this change. 

But browser screensharing (both Firefox and Chromium) will run out of buffer if the browser window is hidden (e. g., you switch to another workspace,) I don't know if this is the issue of the browsers or xdpw. Maybe we should make xdpw to retry if `xdpw_pwr_dequeue_buffer` fails to get a buffer?

ManuLinares:
It it fine but it skips 1 frame ocassionally. Look at the test video generated by `ffmpeg -f lavfi -i testsrc2=size=1920x1080:rate=60 -pix_fmt yuv420p test.mp4`, played by mpv and recorded by obs:

play it frame by frame and noticed that it duplicated frame 100 and jumps to 102, then it duplicated frame 161 and jumps to 163, etc.


https://github.com/user-attachments/assets/8cf6df4f-80f5-4ea3-b2e0-64a518c312fd

I'm on labwc using `mpv --vo=dmabuf-wayland`

DamianBossPL:
I installed the fork and I can confirm it fixes OBS screen capture:

https://github.com/user-attachments/assets/853adeac-c53e-4ed4-bf50-d261c44fb9c5

DamianBossPL:
I just discovered a problem with this. Some applications may fail and show a freeze frame. From my testing, I saw this in Vesktop.

ManuLinares:
> I just discovered a problem with this. Some applications may fail and show a freeze frame. From my testing, I saw this in Vesktop.

Is this specific to this pull request? If it is, ok.
If not, please open a separate issue.

DamianBossPL:
> > I just discovered a problem with this. Some applications may fail and show a freeze frame. From my testing, I saw this in Vesktop.
> 
> Is this specific to this pull request? If it is, ok. If not, please open a separate issue.

It is to do with this pull request. The regular version functions fine, with the FPS of the share screen being normal in Vesktop, but laggy in OBS. This pull request however fixes OBS, but breaks Vesktop.

funk443:
> I just discovered a problem with this. Some applications may fail and show a freeze frame. From my testing, I saw this in Vesktop.

Because next capture is not scheduled if xdpw encounters "out of buffer" errors. I see some retry logics in xdg-desktop-portal-hyprland, maybe we should add that in.

funk443:
I'm thinking maybe we should just rely on Pipewire timers, since we are not doing any VRR while capturing, only a FPS limit in the config file. We can just initialize a new Pipewire timer with the rate `1/max_fps`, then just make our processing function the callback.

This way, we don't need to worry about how we schdule next capture if we encounter "out of buffer" or "buffer already exported" errors, since the processing function is called no matter what happens in it.

This topic deserves its own PR or issue tho, what do you guys think?

emersion:
Oh, if we can convince PipeWire to rate limit the frame rate for us, that'd be very nice indeed! I agree it should be a separate issue/PR.

Something else broken by driving the source: offscreen toplevel captures are driven by the capture client, so we pump out frames way too fast with this PR (basically in a busy loop).

ManuLinares:
I'll unsibscribe from this one. Please ping me if you need some testing done, thanks!

funk443:
> Something else broken by driving the source: offscreen toplevel captures are driven by the capture client, so we pump out frames way too fast with this PR (basically in a busy loop).

Could you elaborate on this? In all my testings xdpw seems to always be the driver of the stream (tested via `pw_stream_is_driving()`).

emersion:
Sorry, I didn't phrase it very well. What I meant is that when a toplevel is captured while offscreen/invisible, the compositor doesn't have a real output to base its frame scheduling off of. As a result, when a client tries to capture the toplevel, the compositor will immediately perform the capture and send a new frame. The client is responsible for rate limiting its capture based on sink needs. Since xdpw doesn't implement PipeWire lazy scheduling, and not all sinks support lazy scheduling anyways, we need to somehow add an arbitrary limit. Maybe the limit could be the number of buffers in the PipeWire swapchain (if we can't dequeue a buffer from PipeWire, hold off capture until we can), but this would add latency.

funk443:
If I'm understanding you correctly, is the behaviour you are mentioning the cause of "buffer already exported" warnings, as seen in [this part of the code].

If this is the case, shouldn't the return statement catch this, and the performance impact be minimal?

[this part of the code]: <https://github.com/funk443/xdg-desktop-portal-wlr/blob/bf6e24db66544be17b8d07afd62ecc80c99d041a/src/screencast/pipewire_screencast.c#L601-L604>

funk443:
Friendly ping :)

I also have some ideas for the issues you brought up.

> Sorry, I didn't phrase it very well. What I meant is that when a toplevel is captured while offscreen/invisible, the compositor doesn't have a real output to base its frame scheduling off of.

I assume what you're talking about here is that `fps_limit_measure_start`/`fps_limit_measure_end` will not function properly if the captured content is offscreen/invisible, and this will result in frames being captured more frequently than it should.

Maybe we can instead simply use a timer to trigger the graph, and if a buffer is available for dequeue, we dequeue it, then fill it with new frame. If no buffer is available for dequeue, just skip the queue/dequeue part and wait for the next timer call.

herry6548:
Has there been any progress on this fix? To be honest, because of this issue, I've noticed that many users are constantly moving away from compositor that is  based wlroots.

funk443:
Still waiting for the owner's review. I myself am running my fork of xdpw on my desktop, you can test it out if you want to. The `pipewire-timer` branch at https://github.com/funk443/xdg-desktop-portal-wlr.

RLB2310:
>  I've noticed that many users are constantly moving away from compositor that is based wlroots.

Would like to see an update for this. Had to move to xdp-hyprland for now, as it handles screensharing without an issue. I'm not even sure my issues are related to this pull request specifically.

ghost:
> Has there been any progress on this fix? To be honest, because of this issue, I've noticed that many users are constantly moving away from compositor that is based wlroots.

I sadly have to agree with this. I feel really rude and entitled saying this, but I think it needs to be said. Sway is my favorite window manager, but the issues with wlroots and the portal that are just ignored have made me *need* to use a different wm for my specific use cases. I really hope wlroots resolves a lot of the known issues before working on new protocols or whatnot. The amount of time PRs spend in limbo has pushed me away from contributing as well. (Huge respect to the wlroots and Sway maintainers, I understand it is not easy maintaining an open source project).

On a positive note, from my (limited) testing, this PR works flawlessly for me.

emersion:
We're seriously lacking maintainers, and I'm way behind in reviews. Please do feel free to ping me about reviews. I will note that messages such as the above aren't particularly motivating me to look at patches.

emersion (approved):
This new version LGTM. I've tested this with offscreen surfaces and it doesn't seem like it goes up to a crazy rate like I feared.

Thanks!

ghost:
> We're seriously lacking maintainers, and I'm way behind in reviews. Please do feel free to ping me about reviews. I will note that messages such as the above aren't particularly motivating me to look at patches.

I am very sorry I was discouraging, that was not my intent. Lacking maintainters does legitimately suck, and I really do wish the best for you.

PR #375, opened 2026-03-12
pipewire: Set SPA_DATA_FLAG_READABLE flag on pw buffers

mpiatka:
Hello,
I'm a developer of [UltraGrid](https://github.com/CESNET/UltraGrid) and I've noticed that after upgrading to PipeWire 1.6 our screen capture began crashing when used with xdg-desktop-portal-wlr.

I've done some investigation and the issue seems to be that after  PipeWire/pipewire@5ccaf297 buffers are no longer mmaped with PROT_READ unless the SPA_DATA_FLAG_READABLE is set.

Setting the flag fixes the issue for me and seems like an obvious solution, but I'm not absolutely sure if changing the flag could have some other subtle implications so feedback is welcome.

Also, the issue is not reproducible in obs, firefox or chromium as they all seem to negotiate dmaBuf and not MemFd as we do in UltraGrid.

Thanks

### DIFF
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -506,6 +506,7 @@ static void pwr_handle_stream_add_buffer(void *data, struct pw_buffer *buffer) {
 	struct xdpw_screencast_instance *cast = data;
 	struct spa_data *d;
 	enum spa_data_type t;
+	uint32_t flags = SPA_DATA_FLAG_READABLE;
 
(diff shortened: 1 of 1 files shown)


mpiatka:
I wasn't sure about `SPA_DATA_FLAG_MAPPABLE` because it was added relatively recently (PipeWire 1.2) and seems to be added automatically somewhere by PW (I already see it set on the consumer side).

But I guess it doesn't hurt to add it to be sure.

PR #376, opened 2026-03-19
Fix screensharing on pipewire 1.6.x

NeoTheFox:
Skip framerate when framerate == 0 in fixate_format and build_format Prevent division by zero in pwr_handle_stream_param_changed


emersion (approved):
Thanks!

PR #381, opened 2026-05-06
screencast: bump backend to v6 with mapping_id and pipewire-serial

### LINKED ISSUE #377: screencast: support PipeWire object.serial for robust stream targeting
PipeWire node IDs can be reused after node destruction. When a monitor is
hotplugged or reconfigured, the backend tears down the PipeWire stream and
creates a new one. The new stream may receive the same node_id that was just
freed, so consumers holding the old ID can silently connect to the wrong node.

This is particularly relevant for wlr compositors since the backend creates
PipeWire streams directly. There's no compositor-level mediation layer
absorbing the churn (unlike GNOME/KDE where the compositor manages PipeWire
streams through its own API).

PipeWire deprecated `node.target` (node_id based) in 0.3.64 (Jan 2023) in
favor of `target.object` (serial based). The serial (`object.serial`) is a
monotonically increasing uint64 that is never reused. The backend already
calls `pw_stream_get_node_id()` in `pwr_handle_stream_state_changed`;
the serial is available from the same stream's node properties.

I have an upstream PR adding `pipewire-serial` as a ScreenCast v6 stream
property, with positive review from a maintainer and a collaborator:
https://github.com/flatpak/xdg-desktop-portal/pull/1942

That PR also makes the frontend derive its advertised version from the
backend (MIN of backend version and 6), so the frontend will only advertise
v6 to apps when the backend actually provides the serial.

The backend is currently at version 4, which also means `mapping_id` from
version 5 is missing. For monitor streams this would just be the output
name (already available in `cast->target->output->name`). With RemoteDesktop
landing (#325), mapping_id becomes directly relevant since it maps PipeWire
streams to libei input regions.

I plan to submit a PR adding both `mapping_id` and `pipewire-serial` once
the upstream spec change lands (flatpak/xdg-desktop-portal#1942, also
discussed in flatpak/xdg-desktop-portal#979). Filing this issue to track
the work and get any early feedback on the approach.

glamberson:
Closes #377.

Implements xdg-desktop-portal ScreenCast v5 (`mapping_id`) and v6 (`pipewire-serial`) stream properties for the wlr backend.

## Why

PipeWire node IDs (uint32) are reused after stream destruction. When monitor hotplug or similar events tear down and recreate streams, consumers holding the old node_id can silently bind to a different stream. PipeWire fixed this in 0.3.64 with `object.serial` (uint64, monotonically increasing, never reused); the matching portal API change merged as flatpak/xdg-desktop-portal#1942.

`mapping_id` (v5) is the matching feature for multi-monitor input targeting: a stable identifier for the source the stream came from. For output streams that's just the output name.

## What changed

Three things, ~45 lines:

1. `include/screencast_common.h`: `XDP_CAST_PROTO_VER` 4 → 6, plus a new `uint64_t pipewire_serial` field on `xdpw_screencast_instance`.
2. `src/screencast/pipewire_screencast.c`: a small static helper `read_object_serial()` that reads `PW_KEY_OBJECT_SERIAL` from the stream's properties and parses it as `uint64_t`. Called right next to `pw_stream_get_node_id()` in `pwr_handle_stream_state_changed`.
3. `src/screencast/screencast.c`: in `method_screencast_start`, append two new entries to the stream properties dict: `mapping_id` (string, from `cast->target->output->name`) and `pipewire-serial` (uint64, from the cached value).

`meson.build`: bump `libpipewire-0.3` from `>= 0.3.62` to `>= 0.3.64` to align with the project-wide `object.serial` migration cutoff.

## Version negotiation

The xdg-desktop-portal frontend (post flatpak/xdg-desktop-portal#1942) derives its advertised ScreenCast version as `MIN(backend_version, 6)`. So apps see v6 only when this backend is loaded and supplies `pipewire-serial` in the stream dict. No v6 advertisement on backends that haven't bumped.

`pipewire-serial` is only emitted in the dict when extraction actually succeeds (non-zero), so the frontend can rely on its presence whenever v6 is advertised.

## Backward compatibility

Existing consumers that read only `node_id` from the stream tuple are unaffected. The new dict keys are additive. KDE has converged on the same fix in parallel (`plasma-wayland-protocols!129` merged 2026-04-30; `xdg-desktop-portal-kde!539` open).

## Testing

Verified with `gcc -fsyntax-only -Werror -Wall -Wextra`; full meson build via CI.

## Provenance and tooling

This change originated from work on RemoteDesktop integration with the wlr stack, where the need for `mapping_id` and `pipewire-serial` to flow through the wlr portal backend surfaced in real use. I agree to the Developer Certificate of Origin (`Signed-off-by` trailer on the commit) and disclose use of Claude Code (Opus) per the Use of LLMs policy (`Assisted-by` trailer on the commit). All code, architecture, and prose are composed, audited, and signed off by me.

Refs: flatpak/xdg-desktop-portal#1942, flatpak/xdg-desktop-portal#979.

### DIFF
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -10,6 +10,9 @@
 #include <sys/mman.h>
 #include <unistd.h>
 #include <assert.h>
+#include <errno.h>
+#include <inttypes.h>
+#include <stdlib.h>
 #include <libdrm/drm_fourcc.h>
 
 #include "screencast.h"
@@ -347,14 +350,34 @@ void pwr_update_stream_param(struct xdpw_screencast_instance *cast) {
 	wl_array_release(&params);
 }
 
+static uint64_t read_object_serial(struct pw_stream *stream) {
+	const struct pw_properties *props = pw_stream_get_properties(stream);
+	if (!props) {
+		return 0;
+	}
+	const char *serial_str = pw_properties_get(props, PW_KEY_OBJECT_SERIAL);
+	if (!serial_str) {
+		return 0;
+	}
+	char *end = NULL;
+	errno = 0;
+	uint64_t serial = strtoull(serial_str, &end, 10);
+	if (errno != 0 || end == serial_str || end[0] != '\0') {
+		return 0;
+	}
+	return serial;
+}
+
 static void pwr_handle_stream_state_changed(void *data,
 		enum pw_stream_state old, enum pw_stream_state state, const char *error) {
 	struct xdpw_screencast_instance *cast = data;
 	cast->node_id = pw_stream_get_node_id(cast->stream);
+	cast->pipewire_serial = read_object_serial(cast->stream);
 
 	logprint(INFO, "pipewire: stream state changed to \"%s\"",
 		pw_stream_state_as_string(state));
-	logprint(INFO, "pipewire: node id is %d", (int)cast->node_id);
+	logprint(INFO, "pipewire: node id is %d, object.serial is %" PRIu64,
+		(int)cast->node_id, cast->pipewire_serial);
 
 	switch (state) {
 	case PW_STREAM_STATE_STREAMING:
--- a/src/screencast/screencast.c
+++ b/src/screencast/screencast.c
@@ -663,6 +663,20 @@ static int method_screencast_start(sd_bus_message *msg, void *data,
 	if (ret < 0) {
 		return ret;
 	}
+	if (cast->target->output && cast->target->output->name) {
+		ret = sd_bus_message_append(reply, "{sv}",
+			"mapping_id", "s", cast->target->output->name);
+		if (ret < 0) {
+			return ret;
+		}
+	}
+	if (cast->pipewire_serial != 0) {
+		ret = sd_bus_message_append(reply, "{sv}",
+			"pipewire-serial", "t", cast->pipewire_serial);
+		if (ret < 0) {
+			return ret;
+		}
+	}
 	ret = sd_bus_message_close_container(reply);
 	if (ret < 0) {
 		return ret;
--- a/include/screencast_common.h
+++ b/include/screencast_common.h
@@ -9,9 +9,13 @@
 
 #include "fps_limit.h"
 
-// this seems to be right based on
-// https://github.com/flatpak/xdg-desktop-portal/blob/309a1fc0cf2fb32cceb91dbc666d20cf0a3202c2/src/screen-cast.c#L955
-#define XDP_CAST_PROTO_VER 4
+// Backend implements ScreenCast v6:
+//   v4 restore tokens, v5 mapping_id, v6 pipewire-serial.
+// The xdg-desktop-portal frontend exposes MIN(backend_version, 6) to
+// apps (see flatpak/xdg-desktop-portal#1942), so apps see v6 only when
+// this backend is loaded and supplies pipewire-serial in the stream
+// dict.
+#define XDP_CAST_PROTO_VER 6
 #define XDP_CAST_DATA_VER 1
 
 enum cursor_modes {
@@ -196,6 +200,7 @@ struct xdpw_screencast_instance {
 	struct spa_video_info_raw pwr_format;
 	uint32_t seq;
 	uint32_t node_id;
+	uint64_t pipewire_serial;
 	bool pwr_stream_state;
 	uint32_t framerate;
 
--- a/meson.build
+++ b/meson.build
@@ -25,7 +25,7 @@ add_project_arguments([
 inc = include_directories('include')
 
 rt = cc.find_library('rt')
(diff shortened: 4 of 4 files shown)


emersion:
Has this been written with an LLM? See CONTRIBUTING for LLM policy.

glamberson:
I had my LLM check all references and found no policy on LLM use anywhere relevant to your concern: not in this repo's CONTRIBUTING.md, not in the wlroots CONTRIBUTING.md it points to, not in either project's README. LLMs are tools. I use them because I get about 300x the work done with them. I am not going to debate that. I do strictly adhere to all stated policies when I submit anything, and I found none here.

emersion:
Here is my CONTRIBUTING guidelines for all GitHub projects: https://github.com/emersion/.github/blob/main/CONTRIBUTING.md

GitHub shows a link to this file in the form to create a new PR.

glamberson:
Thanks. I missed the org-level .github default. I've read it now, find it a very reasonable point of view and policy, and will make sure to comply. I've had a long day, so I'll consider it with the proper care in the morning.

glamberson:
Per my earlier message, I've updated to comply:

- PR description includes a "Provenance and tooling" section noting LLM use, DCO agreement, and that I compose, audit, and own all output.
- Commit amended (force-pushed to `feat/screencast-v6`) with the two trailers:
  ```
  Signed-off-by: Greg Lamberson <greg@lamco.io>
  Assisted-by: Claude Code (Anthropic, Opus)
  ```

Thanks.

src/screencast/pipewire_screencast.c
```diff
@@ -347,14 +350,34 @@ void pwr_update_stream_param(struct xdpw_screencast_instance *cast) {
...
+		return 0;
+	}
+	const char *serial_str = pw_properties_get(props, PW_KEY_OBJECT_SERIAL);
+	if (!serial_str) {
+		return 0;
+	}
+	char *end = NULL;
+	errno = 0;
+	uint64_t serial = strtoull(serial_str, &end, 10);
+	if (errno != 0 || end == serial_str) {
```
emersion: Could we also add `end[0] != '\0'`, to protect against garbage at the end of the input string?

src/screencast/screencast.c
```diff
@@ -663,6 +663,20 @@ static int method_screencast_start(sd_bus_message *msg, void *data,
 	if (ret < 0) {
 		return ret;
 	}
+	if (cast->target->output && cast->target->output->name) {
+		ret = sd_bus_message_append(reply, "{sv}",
+			"mapping_id", "s", cast->target->output->name);
```
emersion, line 668: Is this a good mapping ID?

Mutter generates a fresh UUID here. Output names can be re-used (when the stream is destroyed and recreated, or when the output is unplugged and replugged).

glamberson: `mapping_id` was introduced in xdg-desktop-portal#1080 and libei !249 as a pairing handle between a PipeWire stream and a libei region. jadahl's PR body: "matching a PipeWire stream with a libei region, so that one e.g. can emit absolute pointer events using libei". The name was moved away from "uuid" to drop the connotation of instance identity.

Mutter (mutter!3228, `meta_remote_desktop_session_acquire_mapping_id`) generates fresh UUIDs and stamps the same UUID on both the libei region (`eis_region_set_mapping_id`) and the PipeWire stream (`org.gnome.mapping-id` tag) inside one process. It only emits `mapping_id` when a RemoteDesktop session is active.

That approach is more efficient than agreeing on a shared key out of band, but it is less functional and arguably wrong against the spec. ScreenCast.xml calls `mapping_id` "an identifier used to map different aspects of the resource this stream corresponds to". A UUID generated per session does not identify a resource; it identifies one stream invocation. Mutter consumers cannot track the same logical resource across stream recreation, and cannot pair a stream outside a RemoteDesktop session with a libei device acquired separately.

xdpw is layered. Any libei region producer for a wlr_output lives in a separate process with no channel to receive a UUID we invented. The pairing key both sides can independently derive from the wlr_output is `output->name`.

The reuse you flagged is the pairing semantic working as intended: a region targeting "DP-1" should pair with whichever stream represents "DP-1" now. Stream instance identity (detecting that the underlying object changed) is the job of `pipewire-serial` in this same PR.

src/screencast/pipewire_screencast.c
glamberson: Fixed. `read_object_serial()` now also rejects trailing garbage via `end[0] != '\0'`. Force-pushed `feat/screencast-v6`.

src/screencast/screencast.c
emersion: I've chatted with Jonas on IRC and he said either should work, it mainly depends on how we want to interact with libei. Let's just use this for now, we can revisit it later.

emersion (approved):

PR #386, opened 2026-05-26
fix(screencast): advertise 8-bit SHM fallback formats for non-10-bit-…

### LINKED ISSUE #385: Screenshare fails with SHM-only PipeWire consumers when compositor uses non-8-bit render format
Screen sharing with Firefox (and likely other SHM only PipeWire consumers) fails when the compositor uses a non-8-bit render format (e.g., XRGB2101010 on 10-bit panels). PipeWire link negotiation fails with "no more output formats" because xdpw only advertises standard 8-bit formats (BGRA/BGRx) with mandatory DMA-BUF modifiers, while the only no-modifier (SHM) format advertised is `xRGB_210LE`, a 10-bit format that Firefox doesn't support.

## Reproduction

- Run sway with default config on a 10-bit capable display (or set `render_bit_depth 10`)
- Open Firefox and attempt to screenshare (WebRTC or via portal)
- Select an output in the chooser

## Expected

Screenshare works  Firefox receives video frames.

## Actual
xdpw crashes with "pipewire: fatal error event from core" and systemd restarts it. PipeWire logs show:
```
pw.link: (100.0.0 -> 80.0.0) negotiating -> error no more output formats (-22) (configure-configure)
xdg-desktop-portal-wlr: [ERROR] - pipewire: fatal error event from core
```

The compositor (wlroots) advertises ARGB8888/XRGB8888 as DMA-BUF fallback formats (with the comment "Not all clients support fancy formats") but only advertises the renderer's preferred read format (XRGB2101010) as SHM. Per [wlroots#2784](https://gitlab.freedesktop.org/wlroots/wlroots/-/work_items/2784), this is intentional, downstream is expected to handle consumer compatibility.

ambasta:
…aware consumers

When the compositor uses a non-8-bit render format (e.g., XRGB2101010 on 10-bit panels), PipeWire link negotiation fails because the only SHM format advertised is xRGB_210LE, which SHM-only consumers like Firefox don't support. The 8-bit formats (BGRA/BGRx) are only offered with mandatory DMA-BUF modifiers, making them unreachable for these consumers.

Advertise ARGB8888/XRGB8888 as additional no-modifier SHM format pods when they appear in the DMA-BUF format list but not in the SHM list. Update buffer creation to compute stride for these fallback formats.

Also fix a type mismatch in build_formats() that iterated the shm_formats array (struct xdpw_shm_format, 8 bytes) with a uint32_t pointer (4-byte stride).

fixes #385

### DIFF
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -179,9 +179,40 @@ static void build_formats(struct spa_pod_builder *builder, struct xdpw_screencas
 		}
 	}
 
-	uint32_t *format;
-	wl_array_for_each(format, &cast->current_constraints.shm_formats) {
-		enum spa_video_format pw_format = xdpw_format_pw_from_drm_fourcc(*format);
+	struct xdpw_shm_format *shm_fmt;
+	wl_array_for_each(shm_fmt, &cast->current_constraints.shm_formats) {
+		enum spa_video_format pw_format = xdpw_format_pw_from_drm_fourcc(shm_fmt->fourcc);
+		if (pw_format != SPA_VIDEO_FORMAT_UNKNOWN) {
+			add_pod(params, build_format(builder, pw_format,
+						cast->current_constraints.width, cast->current_constraints.height,
+						cast->framerate, NULL, 0));
+		}
+	}
+
+	uint32_t fallback_formats[] = { DRM_FORMAT_ARGB8888, DRM_FORMAT_XRGB8888 };
+	for (size_t i = 0; i < sizeof(fallback_formats) / sizeof(fallback_formats[0]); i++) {
+		bool already_advertised = false;
+		wl_array_for_each(shm_fmt, &cast->current_constraints.shm_formats) {
+			if (shm_fmt->fourcc == fallback_formats[i]) {
+				already_advertised = true;
+				break;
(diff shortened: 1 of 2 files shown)


ambasta:
Disclosure: Patch was written using LLM, but do let me know if you need changes

emersion:
We can't announce shm formats which aren't advertised by the compositor.

ambasta:
@emersion What do you think is the correct fix here, advertise additional formats in wlroots? Or wait for HDR in Firefox to land?

emersion:
Yeah, I think the compositor should advertise fallback 8-bit formats if the renderer supports that. We've already done that for DMA-BUF but we're missing it for shm: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5126

One tricky bit is that wlroots renderers can currently only advertise a single readback format via `wlr_texture_preferred_read_format()`, they can't advertise multiple. As a result, wlroots can't know in advance whether the renderer would support 8-bit conversion. The API would need to be extended.

pmkap:
If the fix in wlroots is tricky, maybe we can allocate a second buffer in xdpw and convert to the necessary format?

It's not optimal but at least it would fix screensharing until this is properly handled in wlroots.

PR #389, opened 2026-06-19
screencast: retry on Pipewire buffer errors.

funk443:
So after #370, while screencasting through pipewire, if the consumer hasn't consume the Pipewire buffer when xdpw is trying to capture the next frame, or we're unable to obtain a Pipewire buffer, the stream will simply stop because we return from the processing function without scheduling next capture.

This commit adds logic to retry when these situations happen.

I'm terribly sorry for the inconvenience caused by my oversight and lack of testings in #370.

### DIFF
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -641,6 +641,8 @@ static void pwr_handle_stream_on_process(void *data) {
 	return;
 
 retry:
+	; // Empty statement because declaration cannot follow a goto label.
+
 	uint64_t delay_ns = fps_limit_measure_end(&cast->fps_limit, cast->framerate);
 	if (delay_ns > 0) {
 		xdpw_add_timer(cast->ctx->state, delay_ns, pwr_process_retry, cast);
@@ -648,7 +650,6 @@ retry:
 		pwr_process_retry(cast);
 	}
 }
-
 
 static const struct pw_stream_events pwr_stream_events = {
 	PW_VERSION_STREAM_EVENTS,
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -611,12 +611,7 @@ static void pwr_handle_stream_remove_buffer(void *data, struct pw_buffer *buffer
 	buffer->user_data = NULL;
 }
 
-static void pwr_process_retry(void *data) {
-	struct xdpw_screencast_instance *cast = data;
-	pw_stream_trigger_process(cast->stream);
-}
-
-static void pwr_handle_stream_on_process(void *data) {
+static void pwr_handle_stream_on_trigger_done(void *data) {
 	struct xdpw_screencast_instance *cast = data;
 
 	logprint(TRACE, "pipewire: on process event handle");
@@ -628,27 +623,17 @@ static void pwr_handle_stream_on_process(void *data) {
 
 	if (cast->current_frame.pw_buffer) {
 		logprint(DEBUG, "pipewire: buffer already exported");
-		goto retry;
+		return;
 	}
 
 	xdpw_pwr_dequeue_buffer(cast);
 	if (!cast->current_frame.pw_buffer) {
 		logprint(WARN, "pipewire: unable to export buffer");
-		goto retry;
+		pw_stream_trigger_process(cast->stream);
+		return;
 	}
 
 	xdpw_wlr_frame_capture(cast);
-	return;
-
-retry:
-	; // Empty statement because declaration cannot follow a goto label.
-
-	uint64_t delay_ns = fps_limit_measure_end(&cast->fps_limit, cast->framerate);
-	if (delay_ns > 0) {
-		xdpw_add_timer(cast->ctx->state, delay_ns, pwr_process_retry, cast);
-	} else {
-		pwr_process_retry(cast);
-	}
 }
 
 static const struct pw_stream_events pwr_stream_events = {
@@ -657,7 +642,7 @@ static const struct pw_stream_events pwr_stream_events = {
 	.param_changed = pwr_handle_stream_param_changed,
 	.add_buffer = pwr_handle_stream_add_buffer,
 	.remove_buffer = pwr_handle_stream_remove_buffer,
-	.process = pwr_handle_stream_on_process,
+	.trigger_done = pwr_handle_stream_on_trigger_done,
 };
 
 void xdpw_pwr_stream_create(struct xdpw_screencast_instance *cast) {
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -611,6 +611,11 @@ static void pwr_handle_stream_remove_buffer(void *data, struct pw_buffer *buffer
 	buffer->user_data = NULL;
 }
 
+static void pwr_process_retry(void *data) {
+	struct xdpw_screencast_instance *cast = data;
+	pw_stream_trigger_process(cast->stream);
+}
+
 static void pwr_handle_stream_on_process(void *data) {
 	struct xdpw_screencast_instance *cast = data;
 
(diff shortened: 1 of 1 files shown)


src/screencast/pipewire_screencast.c
```diff
@@ -623,17 +628,28 @@ static void pwr_handle_stream_on_process(void *data) {
...
+	return;
+
+retry:
+	; // Empty statement because declaration cannot follow a goto label.
 
+	uint64_t delay_ns = fps_limit_measure_end(&cast->fps_limit, cast->framerate);
+	if (delay_ns > 0) {
+		xdpw_add_timer(cast->ctx->state, delay_ns, pwr_process_retry, cast);
+	} else {
+		pwr_process_retry(cast);
```
emersion: Doesn't this result in a busy loop? (ie, eating 100% CPU until we get a buffer?)

funk443: It doesn't seem to be a busy loop in my testings using OBS and Mozilla gUM test.

I'm not a master in C, but in my understanding, this is more like, if the buffer is yet consumed by the consumer, or we cannot obtain a new buffer from Pipewire at this moment, we simply call this function (`pwr_handle_stream_on_process`) when the next frame should be captured (determined through fps limit,) and this should be capped by the value of `max_fps`. And if the buffer is still unavailable, we simply repeat this process.

emersion: `max_fps` can be zero (unlimited) for virtual outputs and for toplevel capture.

funk443: Maybe we can change `max_fps` to a sensible default? Like 360 fps, which should be more than enough for most of the people, and if someone needs higher framerate or even unlimited framerate, he can simple set `max_fps` to 0?

I don't see other ways to check if the buffered is clear.

funk443:
I tested the new commit with both top-level capture and `max_fps` set to 0, and it doesn't seem to evolve into a busy loop, whether the capture window is visible or not.

funk443:
Tested with OBS with 3-4 Pipewire screencast source, including serveral top-level captures (a floating window, a scratchpad window, and tiled windows) and an output capture. CPU usage seems quite normal for Pipewire processes and xdpw process, and no "unable to export buffer" or "buffer already exported" observed in xdpw logs.

If there are some edge cases I did not consider, please let me know.

FinGu:
This solves issues i had with slack, discord, firefox and chromium...

emersion:
After reading PipeWire, Mutter and KWin sources, I cannot understand this PR. None of them use `trigger_done`.

I also still don't believe retrying in a busy loop is a good idea.

src/screencast/pipewire_screencast.c
emersion: Could we query `SPA_FORMAT_VIDEO_framerate`/`SPA_FORMAT_VIDEO_maxFramerate` from the final negotiated parameters somehow?

funk443: `SPA_FORMAT_VIDEO_framerate` has no meaning here, since the framerate is variable. And the value of `SPA_FORMAT_VIDEO_maxFramerate` is constructed using `max_fps` if I'm not mistaken.

funk443:
So if I'm understanding the Pipewire documents correctly, `trigger_done` is when the process loop we triggered is done, which implies the consumer has consumed the frame. So in theory when this function is called, we shouldn't need to be worry about whether the buffer we sent to the consumer is consumed or not.

As for the issue of unable to obtain buffer from PW, I have no clue on how to solve this. Obviously we should retry to get one from PW, but how frequent should we try to do that, and how should we obtain the limit on how often we should do that? Since `max_fps` can be 0, we simply cannot rely on it. I guess one way to do it is to calculate average framerate the consumer is consuming, and use that as the limit if `max_fps == 0`?

sithglan:
This fixes my issue I have with firefox/chromium webrtc using screego.

emersion:
In my testing, the freeze only happens when the stream state transitions into "streaming". Are you hitting the "out of buffers" error in other situations?

sithglan:
Hello,

> In my testing, the freeze only happens when the stream state transitions into "streaming". Are you hitting the "out of buffers" error in other situations? 

I made you a video. To show the issue:

Text:

I share my screen using screego in firefox or chromium. It works for a
few seconds and than it freezes. This is on Debian testing which has
packaged the unreleased head of
'https://github.com/emersion/xdg-desktop-portal-wlr/
as

> ii  xdg-desktop-portal-wlr                              0.8.3-1 amd64        xdg-desktop-portal backend for wlroots

If I apply one of these two patches, it is stable for several hours for
me:

FIX: https://github.com/emersion/xdg-desktop-portal-wlr/pull/389
Workaround: https://tg.st/u/0001-screencast-raise-pipewire-buffer-pool-from-2-to-4-b.patch

Here is the video. Notice that I have two screens and I record only one
of two screens. Also notice that I use the same browser window to record
me and show me the recording of myself to demo the issue.

https://tg.st/u/2026-07-09_19-03-20_yuv420p.mp4

Cheers,
        Thomas

funk443:
> Are you hitting the "out of buffers" error in other situations?

It can happen constantly while streaming using Firefox (with gUM test page,) if you make the browser window invisible (e. g. put it to some other workspace, or switch to another window in tab/stack mode).

For OBS, it also happenes, but very rarely, it's still unclear when it can happen.

llyyr:
> In my testing, the freeze only happens when the stream state transitions into "streaming". Are you hitting the "out of buffers" error in other situations?

It seems to happen consistently for me if I make the window being streamed obscured

hgouni:
I'm also running into the `out of buffers` issue; it always occurs when sharing per-window (running xdpw 0.8.3). This patch appears to fix it.

beryll1um:
Please merge it faster because Discord is broken on my Artix Linux.
0.8.2 works fine.
Thanks for your work!

widlarizer:
This fixes window sharing for me with sway 1.12, Intel Arc B580, great work. In case another nixOS user wants to use the current top of this feature branch, here's my override: `xdg-desktop-portal-wlr = prev.xdg-desktop-portal-wlr.overrideAttrs (old: { src = final.fetchFromGitHub { owner = "funk443"; repo = "xdg-desktop-portal-wlr"; rev = "74be7063347880f6bf98689e24dd9a6e98032405"; sha256 = "125rsvls0gyc7a0lspvrqm54ckplfvd18q963yfmq47d6n3im6gr";};});`

src/screencast/pipewire_screencast.c
```diff
@@ -629,19 +629,20 @@ static void pwr_handle_stream_on_process(void *data) {
...
 
-
 static const struct pw_stream_events pwr_stream_events = {
 	PW_VERSION_STREAM_EVENTS,
 	.state_changed = pwr_handle_stream_state_changed,
 	.param_changed = pwr_handle_stream_param_changed,
 	.add_buffer = pwr_handle_stream_add_buffer,
 	.remove_buffer = pwr_handle_stream_remove_buffer,
-	.process = pwr_handle_stream_on_process,
+	.trigger_done = pwr_handle_stream_on_trigger_done,
```
smlx, line 645: By switching to `trigger_done` here, don't you kind of reverse the pacing of the capture to the rate that the consumer processes buffers?

According to [pipewire docs](https://docs.pipewire.org/group__pw__stream.html#gab93fa91d8ea80e562bb5e8f6b5bafe21):

> It is possible for the graph iteration to not finish, so pw_stream_trigger_process() needs to be called again even if process() and trigger_done is not called.

So I think `pw_stream_trigger_process()` needs to be called from the xdpw side regardless of `trigger_done` (or `process` for that matter) being called?

funk443: > So I think `pw_stream_trigger_process()` needs to be called from the xdpw side regardless of `trigger_done` (or `process` for that matter) being called?

Then this brings back the question on how do we limit the fps so we don't blow the user's PC up.

smlx: I think you were on the right track with the original approach. I have sent #397 which basically takes your original approach and modifies it slightly to not call `pw_stream_trigger_process` synchronously inside the process handler. Could you take a look to see if it makes sense?

funk443: Isn't this basically what I did in 66ad3d9e50a613c7e9919c92cb9620de19998dde?

emersion:
Superseded by https://github.com/emersion/xdg-desktop-portal-wlr/pull/397

Thanks a lot for working on fixing this still!

PR #393, opened 2026-07-03
screencast: default SelectSources to monitors

### LINKED ISSUE #379: SelectSources fails with "No supported targets specified" when `types` option is omitted
When a D-Bus caller invokes `SelectSources` without including the `types` key in the options dict, the call fails with:

```
[ERROR] - wlroots: No supported targets specified
```

Per the [portal spec](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.ScreenCast.html#org-freedesktop-portal-screencast-selectsources), the `types` option should default to `MONITOR` (1) when not specified. Many callers reasonably omit it expecting this default.

### Root Cause

In `src/screencast/screencast.c`, `type_mask` is initialized to `0`:

```c
uint32_t type_mask = 0;
```

It only gets updated if the caller explicitly sends a `"types"` key in the options dictionary. If the caller omits `types`, `type_mask` remains `0`, and `setup_target()` fails immediately because `0 & ctx->state->screencast_source_types == 0`.

### Suggested Fix

```diff
- uint32_t type_mask = 0;
+ uint32_t type_mask = MONITOR;
```

This matches the default specified in the portal spec.
Not sure if this has any adverse effects.

### Environment

- Ubuntu 26.04
- xdg-desktop-portal-wlr 0.8.1-1

### How I Found This

I hit this bug with NVIDIA's NVFBC-based screen capture on Ubuntu 26.04. This extra check did not exist before in the version used by Ubuntu 25.10. It calls `SelectSources` with `cursor_mode` and `persist_mode` but without `types`. However, I suspect many other programs rely on the spec-defined default of `MONITOR` and would be similarly broken.

scarab-systems:
## Summary

- Default `SelectSources` to monitor capture when the optional `types` key is omitted.
- Keep explicit `types` handling unchanged.

## Rationale

The ScreenCast portal default for omitted `types` is monitor capture. Starting the mask at zero leaves the request with no supported targets and causes `No supported targets specified` before target selection.

Closes #379.

## Testing

- `git diff --check`
- Fedora 44 container: `meson setup /tmp/xdpw-build -Dman-pages=disabled -Dsystemd=disabled && ninja -C /tmp/xdpw-build`

## Provenance and tooling

This patch was prepared with AI assistance. Scarab Systems reviewed and approved the submitted change. The final patch is intentionally small, manually inspected, and tested by the submitter.

### DIFF
--- a/src/screencast/screencast.c
+++ b/src/screencast/screencast.c
@@ -371,7 +371,7 @@ static int method_screencast_select_sources(sd_bus_message *msg, void *data,
 
 	char *key;
 	int innerRet = 0;
-	uint32_t type_mask = 0;
+	uint32_t type_mask = MONITOR;
 	struct xdpw_screencast_restore_data restore_data = {0};
 	while ((ret = sd_bus_message_enter_container(msg, 'e', "sv")) > 0) {
 		innerRet = sd_bus_message_read(msg, "s", &key);


emersion (approved):
LGTM, thanks!

PR #396, opened 2026-07-09
screencast: raise pipewire buffer pool from 2 to 4 buffers

### LINKED ISSUE #395: Chromium screencast frozen after first frame with 0.8.3 (ext_image_copy_capture): immediate "out of buffers" deadlock, fixed by raising XDPW_PWR_BUFFERS
Since updating from 0.7.1 to 0.8.3, screen sharing from Chromium-based clients (Vesktop/Electron, app_id: org.chromium.Chromium) shows only a single static frame. OBS via the same portal works fine.
Setup: mango (wlroots-based, advertises both zwlr_screencopy v3 and ext_image_copy_capture v1), AMD GPU, Mesa 26.x, PipeWire 
Compiled with libpipewire 1.6.5
Linked with libpipewire 1.6.5, NixOS unstable.
Debug log shows negotiation completing normally (explicit modifier, DMABUF, 2 buffers), then immediately on entering "streaming":
pipewire: stream state changed to "streaming"
pipewire: out of buffers
pipewire: unable to export buffer
after which no frames are ever delivered — looks like the consumer holds both buffers of the 2-buffer pool, pw_stream_dequeue_buffer() returns NULL on the first process event, and nothing re-arms the capture loop.
Observations:

force_mod_linear=1 has no effect — with explicit-modifier negotiation the option is never consulted (only the DRM_FORMAT_MOD_INVALID fallback path reads it).
Patching XDPW_PWR_BUFFERS and XDPW_PWR_BUFFERS_MIN from 2 to 4 fully resolves it.
0.7.1 (zwlr_screencopy path) works unmodified on the same system.

Full debug log attached. Debug log Contain attempted force_mod_linear=1 fix.

[xdpw-0.8.3-debug.log](https://github.com/user-attachments/files/29768142/xdpw-0.8.3-debug.log)

sithglan:
The ext_image_copy_capture_v1 backend negotiates a PipeWire buffer pool via SPA_POD_CHOICE_RANGE_Int(XDPW_PWR_BUFFERS, XDPW_PWR_BUFFERS_MIN, 32), and both constants are hardcoded to 2. With only two buffers, some consumers (observed with Chromium's and Firefox's PipeWire screencast source, both via WebRTC getDisplayMedia) end up holding both buffers at once, so pw_stream_dequeue_buffer() returns NULL and xdpw stops delivering new frames. This does not crash or log an error: the portal service and the PipeWire node both stay in a nominally healthy state, so the capture silently freezes on the last successfully delivered frame while consuming near-zero CPU/bandwidth.

The freeze is easy to miss from the portal side alone since nothing errors, but is clearly visible in the consumer's own encoder stats: in a Chromium chrome://webrtc-internals capture, framesPerSecond collapsed from ~30 to exactly 1 within seconds to a couple of minutes of the screen share starting, and keyFramesEncoded stopped incrementing entirely for the rest of the session (13+ minutes observed), while frame dimensions and encoder implementation stayed correct throughout. OBS, which drives its own consumption loop differently, was unaffected on the same system.

Raising XDPW_PWR_BUFFERS and XDPW_PWR_BUFFERS_MIN from 2 to 4 gives the consumer enough headroom to hold onto buffers briefly without starving the producer, and resolves the freeze. force_mod_linear does not help, since explicit-modifier negotiation bypasses it.

Reproduced on two independent systems: sway 1.12 / wlroots 0.19-0.20 on AMD (amdgpu), and a second system on NixOS; both on xdg-desktop-portal-wlr 0.8.3 after the ext_image_copy_capture_v1 switch. Confirmed on the first system across both Chromium and Firefox, ruling out a browser-specific cause, and with GPU, CPU and network load, loss and latency all ruled out as contributing factors.

Closes: https://github.com/emersion/xdg-desktop-portal-wlr/issues/395

Signed-off-by: Thomas Glanzmann <thomas@glanzmann.de>

Assisted-by: Claude Sonnet 5 (Anthropic)

### DIFF
--- a/include/pipewire_screencast.h
+++ b/include/pipewire_screencast.h
@@ -3,8 +3,8 @@
 
 #include "screencast_common.h"
 
-#define XDPW_PWR_BUFFERS 2
-#define XDPW_PWR_BUFFERS_MIN 2
+#define XDPW_PWR_BUFFERS 4
+#define XDPW_PWR_BUFFERS_MIN 4
 #define XDPW_PWR_ALIGN 16
 
 void xdpw_pwr_enqueue_buffer(struct xdpw_screencast_instance *cast);


emersion:
This is a workaround, the freeze can still happen with an arbitrary number of buffers. The root cause of the freeze should be fixed instead of working around it.

sithglan:
I hit this issue today with two monitors on Debian. During a training session after a Debian testing update. This workaround works for me, but I'll use the final fix once it is available. Thank you. Symptom is. If I share my screen using webrtc (firefox or chrome) it works for a few seconds - minutes and than the stream freezes.

PR #397, opened 2026-07-23
screencast: fix stream freezing on PipeWire buffer starvation

### LINKED ISSUE #390: [bug]The obs recording screen often freezes
env:
sway version 1.12
OBS Studio - 32.1.2

# the obs record eDP-1 monitor:
The screen froze after ten seconds.

https://github.com/user-attachments/assets/d64cde59-7f17-46a2-9775-c0285708a3e1

# gpu-screen-record after obs frrezes:

https://github.com/user-attachments/assets/a7c1d2df-113c-4935-aa5e-c7914bfb5cf4

It can be seen that the obs card died in the final recorded scene.

This issue emerged only after 0.8.3.

# Reproduction steps

```conf
    bindsym $mod+Return exec foot

```

keep hold press mod+return two sconds, let it open ten window at same time.

obs will frrezes.

### LINKED ISSUE #395: Chromium screencast frozen after first frame with 0.8.3 (ext_image_copy_capture): immediate "out of buffers" deadlock, fixed by raising XDPW_PWR_BUFFERS

smlx:
Currently, when the stream runs out of buffers, PipeWire's graph
iteration fails and freezes the stream.

This commit introduces a repeating timer to call
pw_stream_trigger_process at the negotiated frame rate when starved of
buffers to ensure recovery once a buffer becomes available.

Fixes: #390
Fixes: #395

### DIFF
--- a/src/screencast/pipewire_screencast.c
+++ b/src/screencast/pipewire_screencast.c
@@ -369,6 +369,11 @@ static uint64_t read_object_serial(struct pw_stream *stream) {
 	return serial;
 }
 
+static void pwr_disarm_process_retry(struct xdpw_screencast_instance *cast) {
+	xdpw_destroy_timer(cast->process_retry);
+	cast->process_retry = NULL;
+}
+
 static void pwr_handle_stream_state_changed(void *data,
 		enum pw_stream_state old, enum pw_stream_state state, const char *error) {
 	struct xdpw_screencast_instance *cast = data;
@@ -393,6 +398,7 @@ static void pwr_handle_stream_state_changed(void *data,
 		// fall through
 	default:
 		cast->pwr_stream_state = false;
+		pwr_disarm_process_retry(cast);
 		break;
 	}
 }
@@ -612,6 +618,38 @@ static void pwr_handle_stream_remove_buffer(void *data, struct pw_buffer *buffer
 	buffer->user_data = NULL;
 }
 
+static uint64_t pwr_process_retry_delay(struct xdpw_screencast_instance *cast) {
+	// max_framerate is only valid when framerate is (0/1) (variable rate negotiated)
+	struct spa_fraction *rate = &cast->pwr_format.framerate;
+	if (rate->num == 0) {
+		rate = &cast->pwr_format.max_framerate;
+	}
+	if (rate->num > 0 && rate->denom > 0) {
+		return (uint64_t)SPA_NSEC_PER_SEC * rate->denom / rate->num;
+	}
+	return SPA_NSEC_PER_SEC / 60; // ~16ms/60fps fallback
+}
+
+static void pwr_process_retry(void *data);
+
+static void pwr_arm_process_retry(struct xdpw_screencast_instance *cast) {
+	if (cast->process_retry) {
+		return;
+	}
+	cast->process_retry = xdpw_add_timer(cast->ctx->state,
+		pwr_process_retry_delay(cast), pwr_process_retry, cast);
+	if (!cast->process_retry) {
+		logprint(ERROR, "pipewire: failed to arm process retry timer, stream may stall");
+	}
+}
+
+static void pwr_process_retry(void *data) {
+	struct xdpw_screencast_instance *cast = data;
+	cast->process_retry = NULL; // timer destroyed in event loop
+	pwr_arm_process_retry(cast);
+	pw_stream_trigger_process(cast->stream);
+}
+
 static void pwr_handle_stream_on_process(void *data) {
 	struct xdpw_screencast_instance *cast = data;
 
@@ -629,12 +667,14 @@ static void pwr_handle_stream_on_process(void *data) {
 
 	xdpw_pwr_dequeue_buffer(cast);
 	if (!cast->current_frame.pw_buffer) {
-		logprint(WARN, "pipewire: unable to export buffer");
+		logprint(WARN, "pipewire: unable to export buffer, dropping frame");
+		pwr_arm_process_retry(cast);
 		return;
 	}
+
+	pwr_disarm_process_retry(cast);
 	xdpw_wlr_frame_capture(cast);
 }
-
 
 static const struct pw_stream_events pwr_stream_events = {
 	PW_VERSION_STREAM_EVENTS,
--- a/include/screencast_common.h
+++ b/include/screencast_common.h
@@ -203,6 +203,7 @@ struct xdpw_screencast_instance {
 	uint64_t pipewire_serial;
 	bool pwr_stream_state;
 	uint32_t framerate;
+	struct xdpw_timer *process_retry;
 
 	// wlroots
 	union {


src/screencast/pipewire_screencast.c
funk443, line 664: Shouldn't we also retry when the client hasn't consumed the buffer?

src/screencast/pipewire_screencast.c
```diff
@@ -629,13 +636,21 @@ static void pwr_handle_stream_on_process(void *data) {
...
-		logprint(WARN, "pipewire: unable to export buffer");
+		logprint(WARN, "pipewire: unable to export buffer, dropping frame");
+		if (!cast->process_retry) {
+			cast->process_retry = xdpw_add_timer(cast->ctx->state, PWR_PROCESS_RETRY_DELAY_NS, pwr_process_retry, cast);
+		}
 		return;
 	}
+
+	if (cast->process_retry) {
+		xdpw_destroy_timer(cast->process_retry);
```
funk443: It seems that the timer is destroyed automatically after the timer is called, so we probably don't need this part (and `cast->process_retry`).

src/screencast/pipewire_screencast.c
```diff
@@ -612,6 +613,12 @@ static void pwr_handle_stream_remove_buffer(void *data, struct pw_buffer *buffer
 	buffer->user_data = NULL;
 }
 
+static void pwr_process_retry(void *data) {
+	struct xdpw_screencast_instance *cast = data;
+	cast->process_retry = xdpw_add_timer(cast->ctx->state, PWR_PROCESS_RETRY_DELAY_NS, pwr_process_retry, cast);
```
funk443: Why not arrange the new timer in the process function?

src/screencast/pipewire_screencast.c
smlx: If `pw_buffer` is non-NULL, then doesn't it mean that xdpw is waiting on the compositor to fill the buffer? So it is waiting for `xdpw_pwr_enqueue_buffer` to be called?

funk443: I think it means the client didn't dequeue (consume) the buffer we sent to PW.

src/screencast/pipewire_screencast.c
smlx: The timer callback re-arms the timer instead of doing it in the process function because my understanding is there is no guarantee that the process function will be called after a buffer starvation event? The repeating timer means that xdpw will poll PipeWire every 1ms, explicitly requesting `process` to be called. When `process` successfully dequeues a buffer, it disarms the timer and regular streaming resumes.

funk443: So I thought our process function will always be called because we're the driver of the stream.

src/screencast/pipewire_screencast.c
smlx: I reworked how the timer is armed/disarmed which avoids this conditional.

src/screencast/pipewire_screencast.c
smlx: Not according to this doc: https://docs.pipewire.org/group__pw__stream.html#gab93fa91d8ea80e562bb5e8f6b5bafe21

> One iteration of the graph will be scheduled when the stream is driving according to pw_stream_is_driving(). If it successfully finishes, process() will be called and the trigger_done event will be emitted. It is possible for the graph iteration to not finish, so pw_stream_trigger_process() needs to be called again even if process() and trigger_done is not called.

smlx:
I updated this PR to avoid the hard-coded 1ms for the process trigger timer and instead check the negotiated framerate (which I believe is always variable?), then the max framerate, then fall back to 16ms/60fps as a last resort as suggested in [this comment](https://github.com/emersion/xdg-desktop-portal-wlr/pull/389#discussion_r3488039538).

smlx:
I'm seeing the bugs linked in the PR description in Debian testing. They are easy to trigger if you select a window with video content to stream on the [gUM test page](https://mozilla.github.io/webrtc-landing/gum_test.html) in Chrome and then rapidly toggle fullscreen on the chrome window. It seems to be easier to trigger with a higher framerate source (which makes sense). With this patch I have been unable to get the stream to stall.

src/screencast/pipewire_screencast.c
smlx: I'm pretty sure this is the producer-side buffer handle.

1. It is set in `xdpw_pwr_dequeue_buffer` when dequeuing a pipewire buffer that will be filled by the compositor.
2. It is set to NULL immediately after being queued in `xdpw_pwr_enqueue_buffer`, and there's no indication if the client has consumed it at that point.

So non-NULL means "frame capture still in progress" so there's no point in triggering another process.

funk443: Ah, I see. But if you don't retry on this condition, won't the stream just die out if the condition is true, as the loop of triggering process is broken.

smlx: I don't think so because `ext_frame_ready` is called when the compositor has filled the buffer, and that ends with a call to `pw_stream_trigger_process`.

funk443: I vaguely remember xdpw could sometimes still encounter this error after #370. Could be wrong tho, cannot test now.

emersion (approved):
I'm not terribly happy about the situation, and wished PipeWire had a way to signal that a buffer is ready to be unqueued. But oh well… This workaround looks reasonable to me!

Thanks!

PR #399, opened 2026-08-12
Fix stall in the wlr-screencopy backend

ismaell:
The backend only captures a single frame then stalls.

Affects compositors that expose zwlr_screencopy_manager_v1 but not ext-image-copy-capture-v1, i.e. wlroots <= 0.18.


emersion (approved):
Thanks!

